PR #20, opened 2025-10-09
feat: check for update before update mole

kowyo:
- **Check for update before update mole**


tw93:
Thank you very much for your submission. Please update the latest code. There are conflicts that need to be resolved.

PR #45, opened 2025-11-02
feat: add Nix garbage collection to developer tools cleanup

MohammedEsafi:
Adds support for cleaning Nix package manager cache using `nix-collect-garbage -d` https://nix.dev/manual/nix/2.32/command-ref/nix-collect-garbage

## Motivation
Nix accumulates disk space over time from old generations


tw93:
LGTM

PR #46, opened 2025-11-10
fix: mole may broke cache and data of surge mac

carolyn-sun:
This PR fixes an issue where Mole identifies Surge Mac's cache and data directories as orphaned files by adding them to the whitelist.

Otherwise, it leads to a rather awkward situation. Since those folders are deleted, it has to re-download the setup scripts; however, the download itself relies on Surge's proxy under the Chinese network environment.


tw93 (approved):
Thank you very much for your submission.

carolyn-sun:
Hey, it works.

Issue #52, opened 2025-11-15
Analyzer shows wrong disk usage


bunizao:
### Reproduce
- `mole` version latest.
- Launch `mole analyze`.
- Navigate to `~/Library/Group Containers/HUAQ24HBR6.dev.orbstack/data`.
- Observe the reported size for `data.img.raw`.

<img width="1560" height="622" alt="Image" src="https://github.com/user-attachments/assets/41efa656-ddb9-4ad4-9577-364885a029e7" />

<img width="1482" height="336" alt="Image" src="https://github.com/user-attachments/assets/52022600-4365-4379-8bc5-3a8e25604015" />

My two Macs, macOS 26 and macOS 15 both show the same problem.

Besides that, **Google Drive** folders exhibit a similar issue. The cloud-only contents are counted using the remote size even though the files were not downloaded to the local disk.

### Idea

The UI displays the sparse file’s logical capacity (8 TB) and sums it into the directory total, making the chart unusable for spotting real disk usage.

tw93:
Thank you very much for your reproduction. I really didn't consider this scene here. I just upgraded a version to fix this problem. 
`mo update` to 1.9.5 experience. Here I have hinted a lot of experiences analyzed by the way.

bunizao:
`1.9.5` fixed this issue! Thank you for your contribution!

PR #53, opened 2025-11-15
fix: accurate file size calculation

bunizao:
#52 

This PR only addresses the disk space calculation issue for Orbstack containers.
Issues with Google Drive calculations will still be included in /CloudStorage.

CleanShot 2025-11-16 at 00 47 24

### DIFF
--- a/cmd/analyze/main.go
+++ b/cmd/analyze/main.go
@@ -1273,7 +1273,7 @@ func scanPathConcurrent(root string, filesScanned, dirsScanned, bytesScanned *in
 		if err != nil {
 			continue
 		}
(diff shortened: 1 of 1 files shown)


tw93:
Thank you very much for your submission, you can check this commit of mine, both issues should be resolved.

Very unfortunately, a bit late, but I'm looking forward to you providing more great features for mole later.

https://github.com/tw93/Mole/commit/e7aedea3e3af1b7ce46b1b028324aa652b5eabe8#diff-962a969e8d2601a4440297f18884b6f254841d3621ae7929908e34db18b3de3fR2457

PR #54, opened 2025-11-15
fix: auth failed when pwd is `space`

bunizao:
My admin password is made up with `spaces`. The prompt always failed because Bash plain `read -s password` collapses leading whitespace unless `IFS` is cleared.

Update the prompt logic to use `IFS= read -r -s password`, which preserves whitespace exactly as typed. With this change the `sudo` call succeeds even when the password is composed entirely of spaces. :)


tw93:
Version 1.9.6 has been released, go try the effect and see if it's ok.

PR #56, opened 2025-11-17
fix:Adjust string matching regex to support spaces and asterisks at any position in whitelist path, and add Rust documentation whitelist option

ClathW:
Adds a whitelist option to protect Rust documentation from cleanup.


ClathW:
### BUG

image

### fix
Updated whitelist path validation regex to allow spaces in paths and wildcards anywhere in the pattern, enabling proper recognition of paths containing 'Application Support' folders and mid-path wildcards during cleanup

PR #155, opened 2025-12-25
improv: prompt user when Docker daemon not running

kwakubiney:
When running mo clean, if Docker is installed but the daemon isn't running, users are now prompted instead of silently skipping.


tw93:
Thank you very much for your submission, a version will be released today.

PR #156, opened 2025-12-25
feat: pnpm clean

rubnogueira:
Add support for PNPM store.

Clean default path. Since it's possible to change the store path, check which one is set and clean that folder as well.

https://pnpm.io/settings#storedir

Edit:
I changed to call "pnpm store prune" and after it to clean the default path.


tw93:
GOOD JOB 👍 @rubnogueira

tw93:
Thank you very much for your submission, a version will be released today.

PR #211, opened 2025-12-30
feat: enhance raycast quick launcher setup script

thijsvanhal:
This PR improves the Raycast process in [setup-quick-launchers.sh](https://github.com/tw93/Mole/blob/main/scripts/setup-quick-launchers.sh). It adds a missing step, adding the script directory in Raycast and uses deeplinks to navigate directly to the right windows and commands in Raycast.


tw93:
@thijsvanhal Thank you very much for your submission, I have merged it and submitted to main, which can be used directly, and then I also updated some optimization points myself.

PR #228, opened 2026-01-01
feat: add shell completion script

jjmartres:
This commit introduces a new `completion` command that provides shell completion for bash, zsh, and fish.

The completion is implemented as a shell script in `bin/completion.sh` and the main `mole` script has been updated to use it.

This approach was chosen to keep the completion logic in shell script, as the `mole` command is primarily a set of shell scripts.

### DIFF
--- a/mole
+++ b/mole
@@ -16,98 +16,98 @@ VERSION="1.17.0"
 MOLE_TAGLINE="Deep clean and optimize your Mac."
 
 is_touchid_configured() {
-	local pam_sudo_file="/etc/pam.d/sudo"
-	[[ -f "$pam_sudo_file" ]] && grep -q "pam_tid.so" "$pam_sudo_file" 2>/dev/null
+    local pam_sudo_file="/etc/pam.d/sudo"
+    [[ -f "$pam_sudo_file" ]] && grep -q "pam_tid.so" "$pam_sudo_file" 2> /dev/null
 }
 
 get_latest_version() {
-	curl -fsSL --connect-timeout 2 --max-time 3 -H "Cache-Control: no-cache" \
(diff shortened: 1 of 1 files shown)


tw93:
@jjmartres Hi! Thanks for working on the shell completion feature - it's a useful addition.

However, I noticed this PR mixes two unrelated changes:

1. **New feature**: Shell completion script (`bin/completion.sh`) ✅
2. **Formatting**: Converting the entire `mole` file from spaces to tabs (719 lines changed)

The formatting change makes it very difficult to review the actual logic changes, and it will pollute the git history - future `git blame` will show this commit for nearly every line.

Could you please split this into separate PRs?

**PR 1 - Shell Completion** (what we need):
- Add `bin/completion.sh`
- Add only the minimal code in `mole` to handle the `completion` command
- Update help text to mention `mo completion`
- Keep existing formatting unchanged

**PR 2 - Code Formatting** (optional, separate discussion):
- If you think we should standardize indentation, we can discuss this separately
- This should be done when there are no active PRs to avoid conflicts

For now, let's focus on getting the completion feature merged without the formatting changes. Would you mind updating this PR or creating a new one?

Thanks for understanding!

jjmartres:
@tw93  Hi! No problem at all - I completely understand. 

You're absolutely right that mixing the formatting changes made the PR harder to review and would have cluttered the git history.

I've reverted the code formatting changes, so this PR now focuses solely on the shell completion feature:
- Added bin/completion.sh with the completion logic
- Minimal changes to mole to handle the completion command
- Updated help text to mention the new feature
- All existing formatting preserved

The PR should be much cleaner to review now. Thanks for the clear feedback and for taking the time to explain the rationale - it makes total sense!

Let me know if there's anything else that needs adjusting

PR #229, opened 2026-01-01
fix: skip Rails vendor during purge

jalen0x:
## Summary
- avoid purging Rails vendor/ (importmap deps)
- tighten Rails project detection and protected filtering

## Testing
- passed


tw93:
Thank you for your submission. I have also optimized it synchronously and added more.
https://github.com/tw93/Mole/commit/a671c63401144c546626e492cd2ade26de1d2512

PR #236, opened 2026-01-03
fix(ui): preserve sort state after uninstall

jimmystridh:
## Summary
- Fix sort state (mode/order) being lost when returning from uninstall
- Persist via MOLE_MENU_SORT_MODE and MOLE_MENU_SORT_REVERSE env vars

## Test plan
- [x] Change sort mode (S key) and order (O key)
- [x] Uninstall an app, press Enter to return
- [x] Verify previous sort settings are retained


tw93 (approved):
LGTM

Thank you for your submission. It has been merged into main. You can install it like this to see the effect.

```
```

PR #237, opened 2026-01-03
fix(ui): improve arrow key responsiveness in menu

jimmystridh:
## Summary
- Remove drain_pending_input from UP/DOWN handlers
- Buffered arrow keys now process naturally instead of being discarded
- Fixes slow scrolling and "runaway cursor" when holding arrow keys

## Test plan
- [x] Hold down arrow in app list - scrolling is now smooth
- [x] Release key - cursor stops immediately


tw93 (approved):
LGTM

Thank you for your submission. You can install the latest one like this.

```
```

PR #240, opened 2026-01-03
feat: Create utility to find stale app installers

JackPhallen:
I don't know if this is a feature you want in the project. Personally, I found it useful for cleaning up disk space. I have a lot of old dmg files that I don't need.

I do not know if this can go in an existing command. Each file needs review because there are valid reasons to keep .dmg files on your machine.

## Add installers command to find and remove installer files

Adds a new installers command that scans common directories (Downloads, Desktop, Documents, etc.) for installer files and allows users to safely remove them.

Features:
- Detects strong installer types: .dmg, .pkg, .mpkg, .iso, .xip
- Identify .zip files that contain installers (skips any .zip with more than 5 files to avoid performance issues)

Usage: mo installers [--help] [--debug]

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ mo optimize                  # Refresh caches & services
 mo analyze                   # Visual disk explorer
 mo status                    # Live system health dashboard
 mo purge                     # Clean project build artifacts
+mo installers                # Find and remove installer files
 
 mo touchid                   # Configure Touch ID for sudo
 mo completion                # Setup shell tab completion
--- a/bin/installers.sh
+++ b/bin/installers.sh
@@ -1,0 +1,341 @@
+#!/bin/bash
+# Mole - Installers command
+# Highlights and helps remove installer files (.dmg, .pkg, .mpkg, .iso, .xip, .zip)
+
+set -euo pipefail
+
+# shellcheck disable=SC2154
+# External variables set by menu_paginated.sh and environment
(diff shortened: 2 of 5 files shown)


JackPhallen:
I have a version of this that mounts the DMG to check if the application is already installed in /Applications but I did not include that because I was worried about overengineering

tw93 (approved):
Hi @JackPhallen,

Thank you for this excellent PR! The code quality is really impressive - comprehensive tests, clean implementation, and great attention to detail. 🎉

I've merged it into the `dev` branch and made a few small refinements to keep it consistent with the project style:

1. **Renamed command to singular form**: `mo installers` → `mo installer` (to match other commands like `mo status`, `mo purge`)
2. **Removed `--help` flag**: Usage documentation now lives in the README only (consistent with other commands)
3. **Simplified title**: Changed to "Clean Installer Files" for consistency with other command titles

The functionality remains exactly the same - your implementation is solid! All 30 tests are passing perfectly.

Really appreciate you taking the time to build this feature. It's definitely useful for keeping Downloads folders clean. The smart ZIP detection (checking for installer payloads while limiting scan depth) is particularly nice.

Additionally, you can experience the effects by installing the dev version in this way.
```
curl -fsSL https://raw.githubusercontent.com/tw93/mole/main/install.sh | bash -s dev
```

PR #245, opened 2026-01-04
chore: Skip tests that require timeout

JackPhallen:
Skip tests that require timeout/gtimeout if the machine does not have timeout/gtimeout.

I am reusing the pattern from core_timeouts.bats test `run_with_timeout`.


tw93 (approved):
LGTM

PR #246, opened 2026-01-04
feat(purge): Additional python purge targets

JackPhallen:
Add additional python cache and venv targets

.pytest_cache - Created by pytest to store cache data between test runs. 

.mypy_cache - Created by mypy to cache type analysis results for faster subsequent runs.

.tox - Created by tox for isolated test environments. These are full virtual environments and can get quite large. 

.nox - Similar to tox


tw93:
LGTM 🍻

Issue #243, opened 2026-01-04
[BUG] The `renv` whitelist is not taking effect


ykonut:
## Describe the bug

在白名单添加`renv`全局缓存后，`mole`依然会将其清理。我曾经在https://github.com/tw93/Mole/issues/70 汇报过添加`renv`白名单的建议，后来并未试验是否修复，如今发现这个问题并没有解决。   

After adding the renv global cache to the whitelist, Mole still cleans it out. I previously reported the suggestion to add the renv to the whitelist on https://github.com/tw93/Mole/issues/70, but I didn't test whether the issue had been fixed. Now, I find that the problem remains unresolved.

## Steps to reproduce

1. 将renv全局缓存添加到白名单
Add the renv global cache to the whitelist.
2. 运行`mo clean`
Run `mo clean`.
3. 直接观察到`~/Library/Caches/org.R-project.R/R/renv`遭到清理
Notice that `~/Library/Caches/org.R-project.R/R/renv` has been cleaned up.

## Expected behavior

白名单路径不会遭到清理。

The whitelist paths will not be cleared.

## Debug logs

我首先确认了`~/.config/mole/whitelist`的内容
I first confirmed the contents of `~/.config/mole/whitelist`.

```text
# Mole Whitelist - Protected paths won't be deleted
# Default protections: Playwright browsers, HuggingFace models, Maven repo, Ollama models, Surge Mac, R renv, Finder metadata
# Add one pattern per line to keep items safe.

~/.m2/repository/*
~/Library/Caches/ms-playwright*
~/.ollama/models/*
~/Library/Caches/com.nssurge.surge-mac/*
~/Library/Application Support/com.nssurge.surge-mac/*
FINDER_METADATA
~/Library/Caches/org.R-project.R/R/renv/*
```

随后执行了`mo clean --dry-run`得到`clean-list.txt`
Subsequently, the command `mo clean --dry-run` was executed to get `clean-list.txt`

```text
# Mole Cleanup Preview - 2026-01-04 11:58:31
#
# How to protect files:
# 1. Copy any path below to ~/.config/mole/whitelist
# 2. Run: mo clean --whitelist
#
# Example:
#   /Users/*/Library/Caches/com.example.app
#


=== User essentials ===
/Users/ykxie/Library/Caches  # 147.2MB (5 items)
/Users/ykxie/.Trash  # 75.4MB (23 items)

=== Finder metadata ===

=== macOS system caches ===
[...]

PR #248, opened 2026-01-04
fix: Respect whitelisted subdirectories

JackPhallen:
See issue and comments on https://github.com/tw93/Mole/issues/243

Changes:
+ Do not delete a directory if a subdirectory is whitelisted
+ Protect `org.R-project.R/R/renv` rather than just it's subdirectories


tw93 (approved):
LGTM

PR #257, opened 2026-01-05
fix: Use single quotes in trap

JackPhallen:
Use single quotes, otherwise this expands now rather than when signalled.

This ultimately doesn't matter in the current implementation, but convention is to use single quotes to avoid unintended bugs.


tw93:
LGTM

PR #274, opened 2026-01-08
fix: golangci-lint config

alexandear:
This PR fixes wrong golangci-lint's configuration. Before the PR, it contains errors:

```console
$ golangci-lint version
golangci-lint has version 2.8.0 built with go1.25.5 from e2e4002 on 2026-01-07T21:23:22Z

$ golangci-lint config verify
jsonschema: "issues" does not validate with "/properties/issues/additionalProperties": additional properties 'exclude-rules' not allowed
jsonschema: "version" does not validate with "/properties/version/type": got number, want string
jsonschema: "" does not validate with "/additionalProperties": additional properties 'linters-settings' not allowed
The command is terminated due to an error: the configuration contains invalid elements
```

Also, I added `golangci-lint config verify` to `check.sh` in order to prevent such mistakes in the future.

Follows 0cc205209cf9cb8b697c5b48b089a6aafb1f7d4c.

_I created a PR to the `main` branch instead of `dev` because `.golangci-lint.yml` doesn't exist in `dev`_.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,7 +1,7 @@
 # golangci-lint configuration for Mole
 # https://golangci-lint.run/usage/configuration/
 
-version: 2
+version: "2"
 
 run:
   timeout: 5m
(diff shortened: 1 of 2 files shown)


alexandear:
Looks like tests are flaky.

PR #276, opened 2026-01-08
Add MacPorts installation option

eirnym:


tw93:
Thanks for the PR! 

We aim to keep the README extremely simple, so we'll stick to Homebrew and the script for now. Also, the different package name mole-cleaner might be confusing in the main doc.

Appreciate your support!

eirnym:
Thank you, I appreciate your opinion

FYI: Many developers like short or quite common words for their tools. In cases of conflicts with port system convention or an existing port, port maintainers usually come up with subtle renames to keep tool name similar  but expressive. The contents of such ports are still the same in most cases and developer/port system maintainer would prefer to add conflict record, rather than to rename tool. There's few examples of renaming, but they're quite rare.

PS: in that case words `port` and `package` are interchangeable.

PR #292, opened 2026-01-10
feat: Skip Time Machine cleaning if running

JackPhallen:
Add Time Machine activity checks before snapshot cleanup

Adds two helper functions to safely detect Time Machine state:
- tm_is_running() - checks if a backup is actively running via tmutil status
- tm_snapshots_mounted() - checks if local snapshots are mounted (browsing Time Machine)

clean_local_snapshots() now skips cleanup when Time Machine is active or when status cannot be determined, preventing potential conflicts during backup operations.

Also updates clean_time_machine_failed_backups() to use the new tm_is_running helper for consistency.

### DIFF
--- a/lib/clean/system.sh
+++ b/lib/clean/system.sh
@@ -119,7 +119,7 @@ clean_time_machine_failed_backups() {
         echo -e "  ${GREEN}${ICON_SUCCESS}${NC} No incomplete backups found"
         return 0
     fi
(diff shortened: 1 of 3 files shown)


JackPhallen:
Instead of automatically skipping the cleanup if Time Machine is running/unknown, I can change this PR to print a larger warning to the user but allow them to clean the backups anyways. I went with the approach of skipping the clean because deleting the backups while Time Machine is running can have unintended consequences.

Issue #293, opened 2026-01-11
[BUG] Time Machine mount check may be too conservative


JackPhallen:
I added `tm_snapshots_mounted()`, but I noticed today that Time Machine may keep snapshots mounted even when Time Machine is not in use. I am still exploring this issue, but I wanted to raise an issue before it is merged into main.

```
~ » mount
/dev/disk3s1s1 on / (apfs, sealed, local, read-only, journaled)
devfs on /dev (devfs, local, nobrowse)
/dev/disk3s6 on /System/Volumes/VM (apfs, local, noexec, journaled, noatime, nobrowse)
/dev/disk3s2 on /System/Volumes/Preboot (apfs, local, journaled, nobrowse)
/dev/disk3s4 on /System/Volumes/Update (apfs, local, journaled, nobrowse)
/dev/disk1s2 on /System/Volumes/xarts (apfs, local, noexec, journaled, noatime, nobrowse)
/dev/disk1s1 on /System/Volumes/iSCPreboot (apfs, local, journaled, nobrowse)
/dev/disk1s3 on /System/Volumes/Hardware (apfs, local, journaled, nobrowse)
/dev/disk3s5 on /System/Volumes/Data (apfs, local, journaled, nobrowse, protect, root data)
map auto_home on /System/Volumes/Data/home (autofs, automounted, nobrowse)
/dev/disk2s1 on /System/Volumes/Update/SFR/mnt1 (apfs, local, journaled, nobrowse)
com.apple.TimeMachine.2026-01-10-174542.local@/dev/disk3s5 on /Volumes/com.apple.TimeMachine.localsnapshots/Backups.backupdb/Jack’s MacBook Pro/2026-01-10-174542/Data (apfs, local, read-only, journaled, nobrowse, protect)
/dev/disk3s1 on /System/Volumes/Update/mnt1 (apfs, sealed, local, journaled, nobrowse)
```

I will investigated further, but for now I will comment out `tm_snapshots_mounted` until I determine if it is safe to delete when snapshots are mounted.

tw93:
Thanks so much for your work, I‘ve merged it

PR #294, opened 2026-01-11
fix: remove Time Machine mount check

JackPhallen:
I found that Time Machine snapshots may stay mounted even when Time Machine is not active https://github.com/tw93/Mole/issues/293


tw93 (approved):
LGTM

PR #297, opened 2026-01-11
fix: refresh not working in overview mode

frozturk:
Refresh functionality in overview mode does not work if cache exists

- Overview refresh (r) in main.go resets sizes and schedules scanOverviewPathCmd, but measureOverviewSize was short-circuiting on cached values.
- Because measureOverviewSize returned loadStoredOverviewSize, overview refresh never re-scanned directories once a cache entry existed, so sizes stayed stale.

This fix:
- removes the cached-size early return in measureOverviewSize so overview refresh always performs a fresh scan (du/logical/cache fallback) and then re-stores the new size.
- adds a regression test to confirm measureOverviewSize changes when underlying data changes (ensures it’s not returning cached values).

This should be the expected behaviour, given the implementation of prefetchOverviewCache function.

### DIFF
--- a/cmd/analyze/analyze_test.go
+++ b/cmd/analyze/analyze_test.go
@@ -204,7 +204,7 @@ func TestMeasureOverviewSize(t *testing.T) {
 	if err := os.MkdirAll(target, 0o755); err != nil {
 		t.Fatalf("create target: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


tw93 (approved):
LGTM

PR #314, opened 2026-01-14
fix(uninstall): Harden brew uninstall

JackPhallen:
@tw93 I have a few ideas for improvements to the brew uninstall fix you implemented. Opening as a draft for now.

#### 1) Cask detection: from “first match wins” heuristics → deterministic-first + duplicate-safe heuristics
**Before:** detection relied heavily on scanning Caskroom by `.app` basename and taking the first hit (e.g. `find ... | head -1`). This can misidentify the cask when multiple casks ship the same `Foo.app`, leading to uninstalling the wrong cask.

**Now:**
- Stage 1: deterministic detection by resolving the selected app bundle path and checking if it lives under `.../Caskroom/<token>/...`
- Stage 2+: heuristic fallbacks remain, but they are **duplicate-safe**:
  - we collect candidate cask tokens and **only return a result if exactly one unique token matches**
  - if multiple tokens match (ambiguous), we refuse to guess and fall back to manual uninstall

**Why:** prevents wrong-cask uninstalls while still improving detection coverage for casks that don’t link directly into Caskroom.

#### 2) Uninstall behavior: from “brew handles it, skip scanning” → “brew removes the app, Mole removes leftovers”
**Before:** for brew-managed apps, the implementation skipped leftover scanning/previewing under the assumption that brew handles cleanup. 

**Now:** all apps—including brew casks—still go through Mole’s leftover discovery + preview + cleanup:
- we use `brew uninstall --cask` for the app bundle when we have a cask token
- Mole still removes user-level leftovers (prefs/caches/app support/etc.), since Homebrew generally does not remove user data

#### 3) Create `brew autoremove` tip

**Now:** we track successful brew uninstalls and suggest `brew autoremove` to cleanup orphaned dependencies (there are valid reasons to keep orphaned dependencies)


JackPhallen:
@tw93 And I think we need https://github.com/tw93/Mole/pull/314/commits/9f441eea86c747f266cb0639404e2fcb2d5aba2c regardless of my other changes

PR #318, opened 2026-01-15
fix: resolve password input issue with special characters

NanmiCoder:
Remove -icanon mode from stty settings to fix password authentication failures when passwords contain special characters like '.' or '@'.

The non-canonical mode (-icanon min 1 time 0) caused character loss in Terminal.app. Using only -echo keeps canonical mode which provides more reliable character handling across all terminal emulators.


tw93 (approved):
LGTM

PR #340, opened 2026-01-21
feat: add purge, installer, touchid, and completion to main menu

rcongdv:
Quick fix to add the other commands to the main menu


tw93:
Thank you very much for your code submission. I considered that this is something many friends need. When I was doing it, I also thought about whether to release it. More importantly, I thought about keeping it simple and easy to use, so I didn't release it. After thinking about it today, I still think it's better not to release it. Users who need it can directly input the command to use it, as these uses are not too many.

Issue #341, opened 2026-01-21
[BUG] `mo analyze` keybindings misalignment


Glydric:
## Describe the bug

When running mo analyze the bottom bar report some abilities like R Refresh but the keybinding does not work for "R" but only for "r", this happens in this section for other keys

## Steps to reproduce

1. Run command: `mo analyze`
2. press "R" (shift+r)
3. See error (Nothing happens)

## Expected behavior

Both "r" and "R" works

## Environment

Please run `mo update` to ensure you are on the latest version, then paste the output of `mo --version` below:

```text
Mole version 1.22.1
macOS: 15.7.3
Architecture: arm64
Kernel: 24.6.0
SIP: Enabled
Disk Free: 63Gi
Install: Manual
Shell: /bin/zsh
```

## Additional context

iamxorum:
Hello!

I've created a pull request for this "bug" with a possible fix.
Please, @tw93, have a look to it when you have time :)

tw93:
Thank you very much for your submission, I have merged it, and now you can use it according to the latest main code; the release version was released recently.
```
curl -fsSL https://raw.githubusercontent.com/tw93/mole/main/install.sh | bash -s latest
```

PR #342, opened 2026-01-21
bug-fix: mo analyze keybindings misalignment

iamxorum:
- r with R
- q with Q
- k with K
- j with J
- l with L
- b with B
- h with H
- o with O

Issue Link: #341


Glydric:
The same should be done for o, O as well

Glydric:
Perfect

PR #353, opened 2026-01-22
Add Puppeteer cache cleanup

shakeelmohamed:
Remove Puppeteer cache folder `$HOME/.cache/puppeteer`, I saw about 5gb savings and I don’t use Puppeteer often.

This could go under `Developer tools` also, but they are browser files so I went with this implementation.
Also added a manual verification step.

Screenshot 2026-01-22 at 3 48 17 PM
<br><br>
Screenshot 2026-01-22 at 3 50 58 PM

### DIFF
--- a/lib/clean/user.sh
+++ b/lib/clean/user.sh
@@ -12,6 +12,53 @@ clean_user_essentials() {
     fi
 }
 
+clean_puppeteer_cache() {
+    local puppeteer_cache="$HOME/.cache/puppeteer"
+    [[ -d "$puppeteer_cache" ]] || return 0
+
+    local size_kb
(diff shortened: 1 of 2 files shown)


tw93:
Hey @shakeelmohamed!

Thanks for adding Puppeteer cache cleanup support! I've merged your PR and made
a quick optimization to keep it consistent with how we handle other browser
caches.

Changes I made:

 - Simplified the implementation to use safe_clean directly (like Chrome, Edge, Arc caches)
 - Removed the interactive confirmation prompt to avoid interrupting the cleaning flow
 - This keeps the behavior consistent: all browser-related caches clean automatically without extra prompts

Thanks again for the contribution!

Issue #354, opened 2026-01-23
[BUG] `mo update` - visual bug after update


iamxorum:
## Describe the bug

After updating mole, at the end of the script it shows "✓ Updated to latest version (unknown)" with the unknown value.

## Steps to reproduce

1. Run command: `mo update`
2. ...
3. See error

## Expected behavior

It should show the version instead.

## Debug logs

<img width="592" height="511" alt="Image" src="https://github.com/user-attachments/assets/50f56895-3a84-4971-9bc3-1ed9f1e2622c" />

I will check where the possible bug is and then, I can come with a PR for the possible fix.

tw93:
@iamxorum Thanks for reporting this!

This should be fixed in V1.23.2. The version detection now has multiple fallback mechanisms:
1. Extract from install.sh output (version X.X.X)
2. Query mole --version
3. Use the version from GitHub API as final fallback

Could you update and verify if the issue is resolved?

```
curl -fsSL https://raw.githubusercontent.com/tw93/Mole/main/install.sh | bash
mo update
```

If you still encounter any issues, feel free to submit a PR!

iamxorum:
Hi!

I will have a look to it, as it seems that still "unknown" is showed.
I will open a PR once I found something :)

iamxorum:
Hi @tw93! 

Created PR with the Fix, please review it when you have time as it has some "conflicts" as I created the branch from dev, not from main.

<img width="610" height="416" alt="Image" src="https://github.com/user-attachments/assets/18e01fb4-29ff-4bad-a3af-424c510570ed" />

Best,
Andrei

PR #357, opened 2026-01-23
bug-fix: Fix "unknown" version when update is finished

iamxorum:
+ added --force flag to trigger "re-update" of mole to the latest version (this was implemented to see if the "unknown" flag is solved and I think it is cool to have it :) )

image

Issue: #354

### DIFF
--- a/mole
+++ b/mole
@@ -13,7 +13,7 @@ source "$SCRIPT_DIR/lib/core/commands.sh"
 trap cleanup_temp_files EXIT INT TERM
 
 # Version and update helpers
-VERSION="1.21.0"
+VERSION="1.23.2"
 MOLE_TAGLINE="Deep clean and optimize your Mac."
 
 is_touchid_configured() {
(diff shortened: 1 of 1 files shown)


tw93:
Hi @iamxorum, thanks for the PR!

I actually fixed the "unknown" version issue last night in commit c0596f5. The fix parses the
version from the installer output using the (version X.X.X) pattern, which matches the actual
output format from install.sh.

Could you test with the latest main branch and let me know if the issue still occurs for you? If so, could you share:
1. The exact output you're seeing
2. How you triggered the update (Homebrew or direct install)

Regarding the --force flag - it's a nice-to-have feature, but I'd like to confirm the core issue is resolved first before considering additional functionality.

iamxorum:
Hi @tw93!

I have triggered the update via `mo update` (installed via curl......) command and it gave me the same "unknown" value (I didn't took a screenshot of it)

I can retry with the --force flag update with the code you fixed and let you know :)

iamxorum:
Hmm

actually it seems it worked now (I don't know why it didn't, maybe it was still the old binary during the update)

image

I think is everything is fine, sorry the confusion :/

Related to the --force could be cool to have it :), just in case of everything

tw93:
@iamxorum 
Thanks for the PR! The version detection fix has already been merged in main.

The `--force` feature is a nice addition. Could you please:

1. Rebase on latest `main` branch
2. Simplify to only the `--force` logic - no need to modify `process_install_output` or `installer_args`

Minimal changes needed:
- Add `force_update` param to `update_mole()`
- Skip version check when `--force` is passed
- Parse `--force`/`-f` in `main()`

Looking forward to the updated PR! �

iamxorum:
I will close this PR and create a new one as the main topic is fixed :)

PR #361, opened 2026-01-24
fix: handle empty arrays in clean_dev_jetbrains_toolbox

TomP0:
## Summary

- Add early return check when `product_dirs` array is empty to prevent unbound variable error
- Add continue check when `version_dirs` array is empty before iterating

Fixes the error:
```
lib/clean/dev.sh: line 248: product_dirs[@]: unbound variable
```

This occurs when running `mo clean` without JetBrains Toolbox installed, as the empty array triggers an error under `set -u`.

## Note on target branch

I'm targeting `main` instead of `dev` because `clean_dev_jetbrains_toolbox` only exists in `main` (and released versions like V1.23.2), not in `dev`. Happy to retarget if you prefer a different approach.


tw93:
LGTM

Issue #392, opened 2026-01-30
[BUG] Dotfiles not being deleted in Trash


la3rence:
## Describe the bug

When running `mo clean`, files in the Trash (`.Trash`) that start with a dot (`.`) are not being deleted. 

## Steps to reproduce

1. Create a hidden file in your Trash:
   ```bash
   touch ~/.Trash/.test_hidden_file
   ```
2. Also create a regular file for comparison:
   ```bash
   touch ~/.Trash/test_regular_file
   ```
3. Run the clean command:
   ```bash
   mo clean
   ```
4. Check if the hidden file remains:
   ```bash
   ls -la ~/.Trash/
   ```
5. Observe that `.test_hidden_file` is still present while `test_regular_file` has been removed.

## Expected behavior

All files in the Trash should be deleted, including hidden files (those starting with `.`). The `mo clean` command should not differentiate between hidden and non-hidden files when emptying the Trash.

## Debug logs

<details>
<summary>Debug output</summary>

```text
✓ User app cache 94 items (116KB)
[DEBUG] Cleaning: User app logs (0 items)
[DEBUG] Cleaning: Trash (1 items)
[DEBUG] === Trash ===
[DEBUG] Risk Level: MEDIUM (User data files)
[DEBUG] Item count: 1
[DEBUG] Files to be removed: All files listed below
[DEBUG] Removing: /Users/lawrence/.Trash/test_regular_file
 ✓ Trash (0B)
```

</details>

## Environment

```text
Mole version 1.23.2
macOS: 26.2
Architecture: arm64
Kernel: 25.2.0
SIP: Enabled
Disk Free: 104Gi
Install: Homebrew
Shell: /bin/zsh
```

## Additional context

I'm not sure if this is an intentional feature or just a bug.
The issue is in `lib/clean/user.sh` after I let LLM read the source code. This happens maybe because bash's default glob pattern `*` does not match hidden files (files starting with `.`). It told me the solutions:

```bash
# Replace:
safe_clean ~/.Trash/* "Trash"

# With:
find ~/.Trash -mindepth 1 -maxdepth 1 -print0 2>/dev/null | while IFS= read -r -d '' item; do
    safe_remove "$item" true || true
done
```

Or Option 2: Enable `dotglob` temporarily:

```bash
shopt -s dotglob
safe_clean ~/.Trash/* "Trash"
shopt -u dotglob
```

[...]

PR #395, opened 2026-01-31
  fix: handle dotfiles in Trash cleanup

rans0:
## Summary

  Fixes a bug where files in `.Trash` that start with `.` (dotfiles) were not being deleted when running `mo clean`.

  ## Problem #392 

  The fallback cleanup code used a glob pattern `~/.Trash/*` which does NOT match dotfiles in bash by default. This meant that hidden files like `.DS_Store`, `.hidden_file`, etc. were left in the Trash even after cleanup.

  ## Solution

  Replaced the glob pattern with a `find` command that properly finds all files including dotfiles:

  ```bash
  # Before (misses dotfiles)
  safe_clean ~/.Trash/* "Trash"

  # After (finds all files)
  while IFS= read -r -d '' item; do
      safe_remove "$item" true || true
  done < <(command find "$HOME/.Trash" -mindepth 1 -maxdepth 1 -print0 2> /dev/null || true)

  This approach is consistent with the existing pattern used for external volume trash cleanup (lib/clean/user.sh:299-300).

  Changes

  - lib/clean/user.sh: Replace glob with find loop in Trash fallback cleanup
  - tests/clean_user_core.bats: Add test verifying find command finds dotfiles

  Verification

  - Old glob method finds 3 items (missing dotfiles)
  - New find method finds 5 items (including dotfiles)
  - All checks pass (ShellCheck, golangci-lint, formatting)
  - New test passes

### DIFF
--- a/cmd/analyze/heap_test.go
+++ b/cmd/analyze/heap_test.go
@@ -150,4 +150,3 @@ func TestLargeFileHeap(t *testing.T) {
 		}
 	})
 }
-
--- a/lib/clean/user.sh
+++ b/lib/clean/user.sh
@@ -20,7 +20,9 @@ clean_user_essentials() {
                 echo -e "  ${GREEN}${ICON_SUCCESS}${NC} Trash · emptied, $trash_count items"
                 note_activity
             else
-                safe_clean ~/.Trash/* "Trash"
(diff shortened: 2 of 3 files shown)


tw93:
LGTM

Issue #396, opened 2026-02-01
[FEATURE] Homebrew uninstall with `zap`


chungweileong94:
## Feature description

When performing Homebrew uninstallation, it is best if it is uninstalled with the `--zap` option.

## Use case

Mole uninstaller doesn't always find all the leftovers. However, a lot of casks came with a list of predefined directories associated with the application, which uninstalling the application with Zap will automatically remove those directories as well.

## Proposed solution

We just have to append `--zap` flag as part of `brew uninstall` command.

## Alternatives considered

-

## Additional context

https://docs.brew.sh/Manpage#:~:text=installing%20cask%20dependencies.-,%2D%2Dzap,-For%20use%20with

iamxorum:
Hello,

I've opened a PR for enhancing the brew uninstall --cask --zap flags.

Please check it out!

tw93:
Hey @chungweileong94! 
Great suggestion! This has been merged in #397. Thank you for your contribution @iamxorum 
Mole now uses `brew uninstall --cask --zap` to remove all associated files (caches, logs, preferences) during Homebrew cask uninstallation. This provides much more thorough cleanup, which aligns perfectly with Mole's goal.

You can try with:
```
curl -fsSL https://raw.githubusercontent.com/tw93/mole/main/install.sh | bash -s latest
```

PR #397, opened 2026-02-01
feat: Homebrew uninstall with `--zap` flag

iamxorum:
Uninstalling mole with Zap will automatically remove also the directories that brew creates for it. -> More complete uninstallation

Also updated the test about uninstalling

Issued: #396


tw93:
LGTM

Issue #394, opened 2026-01-30
[BUG] mo purge cannot find node_modules folders


Angelk90:
## Describe the bug

Using the `mo purge` command, cannot find `node_modules` folders.

tw93:
Thank you for the feedback, I have optimized it, you can check the effect, just install the latest code package.
```
curl -fsSL https://raw.githubusercontent.com/tw93/mole/main/install.sh | bash -s latest
```

Angelk90:
@tw93: It's not working, the path I tested is like this: `/Users/NameUser/Desktop/nodejs`

ghost:
Wouldn't it be better to make this limit user configurable, because in the fix it states

> - Performance impact: ~15-25% slower scan (acceptable trade-off for correctness)

Angelk90:
@iammike4: works correctly in two previous versions.

ghost:
@Angelk90 

My comment is merely a question about that the fix has an impact on the overall performance. Nothing more nothing less, nothing to do with it not working for you

Angelk90:
@tw93 : In the file `open $HOME/.config/mole/purge_paths` I find this:

```
# Mole Purge Paths - Auto-discovered project directories
# Edit this file to customize, or run: mo purge --paths
# Add one path per line (supports ~ for home directory)

~/Desktop
~/NameUser
```

tw93:
@Angelk90 Thank you very much for your feedback, I have continued to optimize it and then tested it for your situation.The scanner was skipping depth-1 directories. 

Install the latest version:
```
curl -fsSL https://raw.githubusercontent.com/tw93/mole/main/install.sh | bash -s latest
```
Your ~/Desktop/nodejs/node_modules will now be found correctly.

---

@iammike4 That "15-25% slower" is from a different fix (commit 81c58b5) which changed max_depth from 4→6.

This fix changes min_depth from 2→1:
- Performance impact: < 3% (negligible)
- Not a feature, it's a bug fix - depth-1 directories should have always been scanned
- Safety unchanged: depth-0 still protected

No configuration needed. The previous commit was the expensive one, this one is basically free.

Angelk90:
@tw93 : `◎ Mole installed successfully, version 1.23.2-edge`

It's not working.

```
mo purge

Purge Project Artifacts

✓ Great! No old project artifacts to clean
```

tw93:
[...]

PR #399, opened 2026-02-01
Fix use find

Angelk90:
Try to solve the problem: #394

@tw93 : I don't know if it should be put true at the end.

`2>&1 | tee "$output_file.raw" > /dev/null` or `2>&1 | tee "$output_file.raw" > /dev/null || true`

Screenshot 2026-02-01 alle 13 54 51

### DIFF
--- a/lib/clean/project.sh
+++ b/lib/clean/project.sh
@@ -411,21 +411,25 @@ scan_purge_targets() {
 
     if [[ "$use_find" == "true" ]]; then
         # Pruned find avoids descending into heavy directories.
-        local find_expr=()
         local prune_dirs=(".git" "Library" ".Trash" "Applications")
-        for dir in "${prune_dirs[@]}"; do
-            find_expr+=("-name" "$dir" "-prune" "-o")
+        local purge_targets=("${PURGE_TARGETS[@]}")
+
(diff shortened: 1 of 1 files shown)


tw93:
@Angelk90 Is your local test good? You can test with `./mo purge`.

tw93 (approved):
LGTM

tw93:
@Angelk90 Thank you for the PR! I've merged it and applied a follow-up fix.
**The issue:** Your PR was missing `-print -prune` at the end of the find command, which caused `find` to continue recursing into matched directories instead of stopping.

What I fixed:
```
- \( "${target_expr[@]}" \) \
- 2>&1 | tee "$output_file.raw" > /dev/null
+ \( "${target_expr[@]}" \) -print -prune \
+ 2> /dev/null > "$output_file.raw" || true
```

Why this matters:
- -print: Output the matched directory path
- -prune: Stop recursing into that directory (critical for performance)
- || true: Prevent script from failing on permission errors

Your refactoring of the array building logic was good! I just needed to adjust where -print goes (outside the loop, not inside).
Please test again:
```
mo purge
```
Your ~/Desktop/nodejs/node_modules should now be found correctly.

PR #410, opened 2026-02-03
Fix use find

Angelk90:
Try to solve the problem: #394

@tw93 : 
See other possible solutions here: https://github.com/tw93/Mole/issues/394#issuecomment-3835147717

### DIFF
--- a/lib/clean/project.sh
+++ b/lib/clean/project.sh
@@ -429,7 +429,7 @@ scan_purge_targets() {
         command find "$search_path" -mindepth "$min_depth" -maxdepth "$max_depth" -type d \
             \( "${prune_expr[@]}" \) -prune -o \
             \( "${target_expr[@]}" \) -print -prune \
-            2> /dev/null > "$output_file.raw" || true
+            2> /dev/null | cat > "$output_file.raw" || true
 
         process_scan_results "$output_file.raw"
     fi


tw93:
Thanks for the PR, but I don't think this change will fix the issue.
Adding `| cat` doesn't change the behavior - it just adds an unnecessary process. The redirection `2> /dev/null > file` works correctly in Bash.
**The real issue** from #394 was already fixed in two commits:
- `c34d8e7` - Changed min_depth from 2→1  
- `cb779d9` (#399) - Added find fallback when fd fails
If `mo purge` still isn't finding your `~/Desktop/nodejs/node_modules`, the most likely causes are:
1. **Recently modified** - The folder was modified less than 7 days ago (default `MIN_AGE_DAYS=7`). This is a safety feature to avoid deleting active projects.
2. **Path not scanned** - Check if `~/Desktop` is in your scan paths: `cat ~/.config/mole/purge_paths`
3. **fd tool issue** - Try `MO_USE_FIND=1 mo purge` to force using find instead of fd
Could you check these and let me know?

Angelk90:
@tw93 : 

1) The folder is recent, yes.
2) `~/Desktop` it's in the file.
3) Tried it but it doesn't work.

So why is it that if the folder is recent, using `| cat` finds the folders, shouldn't it still not find them because of `MIN_AGE_DAYS=7`?

I tried setting `MIN_AGE_DAYS=0`,
- when I use the solution with `| cat`, it works and the recent text doesn't appear.
- when I use the current solution, it prints: `✓ Great! No old project artifacts to clean`

tw93:
@Angelk90 Thanks for the detailed testing! You found a real bug. The `| cat` isn't actually fixing the issue - the real problem is that when `fd` finds nothing (but exits successfully), the code wasn't falling back to `find`. Plus I suggested using `MO_USE_FIND=1` in my earlier comment but never actually implemented it 😅

I've just pushed a fix (commit `7f787b5`) that implements the `MO_USE_FIND` flag and fixes the fallback logic. Can you test it with 

`curl -fsSL https://raw.githubusercontent.com/tw93/mole/main/install.sh | bash -s latest`

and let me know if `mo purge` works now? I'm closing this PR since the proper fix is in main. Thanks again for helping debug this!

Angelk90:
@tw93 : I tested the latest version, but it doesn't work either.
I also tried `MO_USE_FIND=1 mo purge`, but it doesn't work.
I also tried setting `readonly MIN_AGE_DAYS=0`, but it doesn't work.

It only works if I use `| cat`.

Issue #426, opened 2026-02-07
[FEATURE] Exclude External Disks from Analyze


ghost:
Have an option to exclude external drives under /Volumes to be analyzed while running

` mo analyze`

At the moment I have 3 external drives attached (not even my NAS) and it takes a long time to analyze (which is to be expected) thus I wish to be able to exclude external drives from that option.

TiA

iamxorum:
Hello,

Worked on that 'feature' and proposed it via PR. Let's see what @tw93 will say about this ;)

tw93:
@iamxorum Thank you, your contribution has been merged into the main branch.
@iammike4 The item has been removed, If users want external drives, they can run mo analyze /Volumes (or a specific mounted path).

Try with
```
curl -fsSL https://raw.githubusercontent.com/tw93/mole/main/install.sh | bash -s latest
```

PR #428, opened 2026-02-07
feat: Exclude External Disks from Analyze

iamxorum:
Created optional flag `--exclude-volumes` for `mo analyze` to consent the users to analyze the filesystem without taking in consideration attached Volumes. Normal behaviour (with volumes included) is left untouched.

Issue: #426 

[15:32] iamxorum@xrm-mac-pro ~/Desktop/XRM/Personal/Contributions/Mole git:(feat-426)
➜ ./bin/analyze-go --exclude-volumes
image

[15:32] iamxorum@xrm-mac-pro ~/Desktop/XRM/Personal/Contributions/Mole git:(feat-426)
➜ ./bin/analyze-go
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ mo optimize --dry-run        # Preview optimization actions
 mo optimize --debug          # Run with detailed operation logs
 mo optimize --whitelist      # Manage protected optimization rules
 mo purge --paths             # Configure project scan directories
+mo analyze --exclude-volumes # Skip external drives under /Volumes in overview
 ```
 
 ## Tips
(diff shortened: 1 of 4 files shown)


tw93:
@iamxorum Thanks for the contribution and for raising this pain point.
I merged the PR first, then refined the UX in follow-up:
- We removed the `--exclude-volumes` flag to keep `mo analyze` simple.
- Overview now skips `/Volumes` by default for faster startup.
- If users want external drives, they can run `mo analyze /Volumes` (or a specific mounted path).
So the same goal is preserved, with fewer flags and cleaner CLI behavior.

iamxorum:
Looks good, at the end is your product, you know better what you want to be like ;)

PR #442, opened 2026-02-11
feat(clean): add --trash option to move cleaned files to Trash instead of deleting

### LINKED ISSUE #439: [FEATURE] moving cleaned files to trash can instead of directly removing
I think at least an option not to permanently delete files but moving them to trash bin should be presented.

We must be able to choose between removing and moving to trash.

## Use case

for cleaning function not removing them directly but showing a preview of the files first and an option to moving to trash can.

## Proposed solution

making these as options. previewing before deletion and/or moving to trash.

## Alternatives considered

I can not have. I already tought 2.

## Additional context

None.

spider-yamet:
## Summary
Adds an option to move cleaned files to the system Trash instead of permanently removing them, so users can recover items if needed.

Closes #439

## Changes
- **`lib/core/file_ops.sh`**
  - Add `safe_move_to_trash()` that uses macOS Finder (osascript) to move paths to Trash.
  - When `MOLE_USE_TRASH=1`, `safe_remove()` tries trash first and falls back to `rm` if trash fails (e.g. headless or permission).
  - Dry-run messages show "Would move to Trash" when `--trash` is used.
- **`bin/clean.sh`**
  - Add `--trash` / `-t` flag and set `MOLE_USE_TRASH=1`.
  - Update footer hint to mention `--trash`.
- **`lib/core/help.sh`**
  - Document `--trash, -t` in clean help.

## Usage
- `mo clean --dry-run` — preview (no changes)
- `mo clean --dry-run --trash` — preview move to Trash
- `mo clean --trash` — clean and move items to Trash
- `mo clean` — default: permanent removal (unchanged)

## Notes
- Trash is only used for user-level paths that go through `safe_remove()`. `safe_sudo_remove()` (system paths) is unchanged and still permanently removes.
- On non-macOS or when osascript is unavailable, the code falls back to permanent removal.

## Testing
- `tests/cli.bats`: `mo clean --help` shows `--trash`.
- `tests/clean_core.bats`: `mo clean --dry-run --trash` does not delete files.
- `tests/core_safe_functions.bats`: `safe_remove` with `MOLE_USE_TRASH=1` removes the file (trash or fallback).

### DIFF
--- a/bin/clean.sh
+++ b/bin/clean.sh
@@ -22,6 +22,7 @@ source "$SCRIPT_DIR/../lib/clean/user.sh"
 
 SYSTEM_CLEAN=false
 DRY_RUN=false
+USE_TRASH=false
 PROTECT_FINDER_METADATA=false
 IS_M_SERIES=$([[ "$(uname -m)" == "arm64" ]] && echo "true" || echo "false")
 
@@ -705,7 +706,7 @@ start_cleanup() {
     echo ""
 
     if [[ "$DRY_RUN" != "true" && -t 0 ]]; then
-        echo -e "${GRAY}${ICON_WARNING} Use --dry-run to preview, --whitelist to manage protected paths${NC}"
+        echo -e "${GRAY}${ICON_WARNING} Use --dry-run to preview, --trash to move to Trash, --whitelist to manage protected paths${NC}"
     fi
 
     if [[ "$DRY_RUN" == "true" ]]; then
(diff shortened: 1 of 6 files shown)


tw93:
@spider-yamet Thanks for the suggestion and for taking the time to describe the use case.

After reviewing this, we will not plan this feature for now.

Reasoning:
1. `mo clean` is designed for immediate disk reclaim, while moving files to Trash does not reliably free space right away.
2. We already provide a safe preview path with `--dry-run`, plus protection controls via `--whitelist`.
3. Adding a Trash mode introduces behavior ambiguity across environments (interactive/headless) and increases maintenance complexity.

So for now, we keep the current safety model: preview first (`--dry-run`), then clean with explicit user control (`--whitelist`).

We may revisit this in the future if we can provide a strict and fully predictable Trash-only workflow.

PR #450, opened 2026-02-12
feat: add generic flutter clean support

mariovtor:
## Description

This PR adds support for cleaning Flutter project artifacts (`.dart_tool` and `build` directories) to the `mo clean` command.

## Changes

- **Flutter Support**: Adds parallel scanning for Flutter projects (similar to existing Next.js and Python scans).
- **Targeted Cleaning**: Removes `.dart_tool` and its sibling `build` directory when found.
- **Deep Scan**: configured  with `-maxdepth 5` to correctly locate nested projects (e.g., inside `projects/client/app`).
- **Safety**: Explicitly excludes `.fvm` (Flutter Version Management) directories to prevent accidental deletion of installed SDKs/versions.
- **Compatibility**: implementation uses BSD syntax compatible with macOS.

## Verification

- [x] Verified with `mo clean --dry-run`: correctly identifies Flutter projects.
- [x] Verified safety: confirmed that `.fvm` directories are ignored.
- [x] Ran `./scripts/check.sh`: passed syntax checks and optimizations.

### DIFF
--- a/lib/clean/caches.sh
+++ b/lib/clean/caches.sh
@@ -165,7 +165,7 @@ clean_project_caches() {
             -not -path "*/Library/*" \
             -not -path "*/.Trash/*" \
             -not -path "*/node_modules/*" \
-            -not -path "*/.git/*" \
+            -not -path "*/.*" \
             2> /dev/null || true
     ) > "$nextjs_tmp_file" 2>&1 &
     local next_pid=$!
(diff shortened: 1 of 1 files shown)


tw93:
@mariovtor Merged, thanks! 🎉

The Flutter clean support is now available in `mo clean`. Good job on the parallel scan implementation and `.fvm` protection.

PR #457, opened 2026-02-15
fix: deep clean npm residual caches

### LINKED ISSUE #454: [FEATURE] Fully clean npm cache
## Feature description

Currently `mo clean` only runs `npm cache clean --force` which doesn't clean everything in the npm cache folder

## Use case

On my machine there's 500 MB more I'd like `mo clean` to delete.

Image

## Proposed solution

Delete the following folders after running `npm cache clean --force`:
- `$(npm get cache)/_prebuilds`
- `$(npm get cache)/_npx`
- `$(npm get cache)/_logs`

Another option is to simply delete the whole npm cache folder (result of `npm get cache`).

## Alternatives considered

No

## Additional context

None

mickyyy68:
## Summary
- extend `clean_dev_npm` to clean npm residual directories not removed by `npm cache clean --force`
- clean `_cacache`, `_npx`, `_logs`, and `_prebuilds` under `~/.npm`
- detect `npm config get cache` and clean the same residual directories under custom cache path when it differs from default
- fallback safely to default cache root when npm returns an empty or non-absolute path

## Tests
- `bats tests/clean_dev_caches.bats` ✓
- `./scripts/test.sh` shell suite: `424 tests, 0 failures, 7 skipped` ✓
- `./scripts/check.sh` blocked in this environment by DNS access to `golangci-lint.run`
- `go test ./...` blocked in this environment by DNS access to `proxy.golang.org`

Closes #454

### DIFF
--- a/lib/clean/dev.sh
+++ b/lib/clean/dev.sh
@@ -17,10 +17,35 @@ clean_tool_cache() {
 }
 # npm/pnpm/yarn/bun caches.
 clean_dev_npm() {
+    local npm_default_cache="$HOME/.npm"
+    local npm_cache_path="$npm_default_cache"
+
     if command -v npm > /dev/null 2>&1; then
         clean_tool_cache "npm cache" npm cache clean --force
+
+        start_section_spinner "Checking npm cache path..."
+        npm_cache_path=$(run_with_timeout 2 npm config get cache 2> /dev/null) || npm_cache_path=""
+        stop_section_spinner
+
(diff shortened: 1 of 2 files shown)


tw93:
@mickyyy68  LGTM! Thanks for the refactor. The loop approach is much cleaner and easier to maintain. The added comment also makes it clear why these directories need manual cleanup.

Tested locally:
- Syntax check: passed
- Dry run: shows correct cleanup output
- All 7 dev cache tests passing

PR #461, opened 2026-02-15
feat(purge): add common React Native targets

DimitarNestorov:
Added common build artifacts folders for React Native projects. That `super-app` folder dropped from 6+ GB to 10 MB.

image


tw93:
LGTM

Issue #475, opened 2026-02-20
[BUG] mole purge not show the new configured ~/code path


uluumbch:
## Describe the bug

`mo purge` not read the `~/code` path

## Steps to reproduce

1. Run command: `mo purge`
2. wait until finish

## Expected behavior

my code folder is `~/code` not `~/Code` --notice the capital letter. I try to change it but it has no effect. I try to add new folder and add to config path, it scan corretctly but not my `~/code` folder

## Debug logs

Please run the command with `--debug` flag and paste the output here:

```bash
mo purge--debug
[DEBUG] Using find for scanning
[DEBUG] Using find for scanning
[DEBUG] Using find for scanning
[DEBUG] Using find for scanning
Purge Project Artifacts

✓ Great! No old project artifacts to clean
```

## Environment

Please run `mo update` to ensure you are on the latest version, then paste the output of `mo --version` below:

```text
1.26.0
```

## Additional context

here is my conifg purge_paths
```bash
cat ~/.config/mole/purge_paths
# Mole Purge Paths - Auto-discovered project directories
# Edit this file to customize, or run: mo purge --paths
# Add one path per line (supports ~ for home directory)

~/Documents
~/code
~/flutter
~/kode
```

Angelk90:
@uluumbch: 
What version of Mac OS are you using?
Do you have `Intel` or `Silicon`?
Do you have `fd` installed?

uluumbch:
> [@uluumbch](https://github.com/uluumbch): What version of Mac OS are you using? Do you have `Intel` or `Silicon`? Do you have `fd` installed?

I use apple silicon (M1) with sequoia.

Thanks, installing `fd` fix my issue.
```bash
brew install fd
```

PR #476, opened 2026-02-21
Update README with 'fd' installation instructions

uluumbch:
Added recommendation for installing 'fd' on macOS.
follow up instruction on #475


tw93 (approved):
LGTM

PR #480, opened 2026-02-21
Feat/nanobanana icon

nimiology:


nimiology:
ive added a ui with flutter

Issue #495, opened 2026-02-25
Bug: Application Support scan hangs on large directories


fbenitez0001:
## Bug: Application Support scan hangs on large directories

**Environment:**
- macOS Darwin 25.2.0 (Apple Silicon)
- Mole v1.27.0 (Homebrew)

**Issue:**
Application Support scan takes 8+ minutes and appears to hang.
Root cause: 392K files in ~/Library/Application Support.

**Suggestion:**
- Add progress indicator during scan
- Consider parallel processing or sampling for large directories

tw93:
Hi @fbenitez0001,

This issue has been fixed.

**Root cause**: the Application Support cleanup path used expensive scanning patterns on very large folders (recursive `du` plus repeated `find`) and had limited progress feedback, so it could appear to hang.

The fix has been merged in [#496](https://github.com/tw93/Mole/pull/496), and we also shipped a follow-up optimization to improve long-scan progress visibility and size accounting accuracy.

You can install the latest version to test:

```bash
curl -fsSL https://raw.githubusercontent.com/tw93/mole/main/install.sh | bash -s latest
```
I'll publish a new official release soon.

fbenitez0001:
awesome, thanks @tw93 !

PR #496, opened 2026-02-25
fix: avoid hang on large Application Support directories (#495)

imnotnoahhh:
## Problem

`clean_application_support_logs()` in `lib/clean/user.sh` hangs 8+ minutes on directories with 392K files due to three bottlenecks:

1. **`du` recursive traversal** — `get_path_size_kb` calls `du -skP` on each candidate directory, recursing into all subdirectories
2. **Double `find` calls** — one `find` to check if directory has contents, another to iterate and delete
3. **No progress feedback** — user sees no output during the long scan, making it appear frozen

## Fix

- Replace `du` + double `find` with a single `find -print0` loop that accumulates size via `stat -f%z` during iteration (follows existing pattern from `apps.sh` and `system.sh`)
- Only stat regular files (not subdirectories) for approximate but fast size reporting
- Add `update_progress_if_needed` in the outer `for app_dir` loop for user-visible progress
- Apply the same fix to the Group Containers loop

## Testing

- `./scripts/check.sh` — shfmt, ShellCheck, syntax all pass
- `bats tests/clean_user_core.bats` — 15/15 pass
- `./scripts/test.sh` — 442 tests, 0 failures

### DIFF
--- a/lib/clean/user.sh
+++ b/lib/clean/user.sh
@@ -727,10 +727,17 @@ clean_application_support_logs() {
     local _ng_state
     _ng_state=$(shopt -p nullglob || true)
     shopt -s nullglob
+    local app_count=0
+    local total_apps
+    total_apps=$(find ~/Library/Application\ Support -mindepth 1 -maxdepth 1 -type d 2> /dev/null | wc -l | tr -d ' ')
+    local last_progress_update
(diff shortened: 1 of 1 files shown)


tw93 (approved):
LGTM

PR #502, opened 2026-02-26
ci: add Swift menubar build verification (#501)

### LINKED ISSUE #501: [CI] Add Swift build verification for menubar component
## Problem

The CI workflows (`check.yml`, `test.yml`) verify Go code (build, vet, format, tests) and shell scripts (ShellCheck, syntax), but do not build or verify the Swift menubar component at `cmd/menubar/`.

This means Swift compilation errors introduced by PRs would not be caught until the release workflow runs.

## Proposed solution

Add a `swift build -c release` step to the test workflow that runs on macOS runners. This ensures the menubar binary compiles successfully on every PR.

## Context

The Swift menubar was added in #500. Since it depends on `status-go --json --watch` for data, verifying the Swift build in CI prevents silent breakage.

imnotnoahhh:
Closes #501

## Summary

Add a `swift build -c release` step to the compatibility job in the test
workflow. This ensures the Swift menubar component at `cmd/menubar/` compiles
successfully on both macOS 14 and macOS 15 for every PR.

## Why

The CI currently verifies Go code (build, vet, format, tests) and shell
scripts (ShellCheck, syntax), but does not build the Swift menubar. This
means Swift compilation errors would only be caught during the release
workflow, which is too late.

## Changes

- `.github/workflows/test.yml` — Add "Build Swift menubar" step to the
  compatibility matrix job (macos-14, macos-15)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,6 +46,11 @@ jobs:
           source lib/core/common.sh
           echo "✓ Successfully loaded on ${{ matrix.os }}"
 
+      - name: Build Swift menubar
+        run: |
+          cd cmd/menubar && swift build -c release
+          echo "✓ Swift menubar builds on ${{ matrix.os }}"
+
   security:
     name: Security Checks
     runs-on: macos-latest


tw93:
Hi, thank you for the thoughtful work and detailed proposal.

After review, we decided not to move forward with menubar-related features in Mole core at this stage. We are currently keeping Mole focused on cleanup/optimization workflows and on-demand terminal/TUI commands.

I’m closing this PR as out of current scope for now. We really appreciate your contribution and understanding.

PR #505, opened 2026-02-27
fix(status): correct battery power parsing when discharging

Harrison3-1122:
## Summary
- fix invalid battery power display caused by wrapped `uint64` values from macOS `ioreg`
- normalize `BatteryPower` sign so internal metrics stay consistent (`positive = discharging`)
- keep display-side sanity guard to avoid showing unreasonable power values
- add regression tests for parsing and rendering behavior

## Details
`ioreg` may expose `BatteryPower` as two's-complement wrapped `uint64` values, and its sign is battery-centric (`positive = charging`, `negative = discharging`).

This change:
- restores signed values safely (including wrapped values)
- converts to system-centric internal convention (`positive = discharging`)
- validates sane power ranges before rendering

## Testing
- `./scripts/check.sh`
- `./scripts/test.sh`
- `go test ./cmd/status`

### DIFF
--- a/cmd/status/metrics_battery.go
+++ b/cmd/status/metrics_battery.go
@@ -20,8 +20,8 @@ var (
 )
 
 const (
-	maxSystemPowerMilliwatts = 1000000 // 1000W
-	maxBatteryPowerMilliwatts = 200000 // 200W
+	maxSystemPowerMilliwatts  = 1000000 // 1000W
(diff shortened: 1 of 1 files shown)


tw93:
@Harrison3-1122 Hi, thank you for this fix.

This change is in scope for Mole core and looks useful. We will keep this PR open for normal review and merge flow.

PR #523, opened 2026-03-02
Clean dynamic size formatting

Angelk90:


Angelk90:
@tw93: Could you implement this for purge too?

PR #540, opened 2026-03-05
Add Update as a conditional 6th option in the interactive menu

xde013:
Currently, updating Mole requires knowing and typing mo update each time...
Now when a new version is detected (via the existing update_message cache), a numbered Update entry appears as option 6 in the mo interactive menu. 

Changes:
- show_main_menu: renders option 6 only when MAIN_MENU_SHOW_UPDATE is true
- interactive_main_menu: sets MAIN_MENU_SHOW_UPDATE from the existing update_message cache, keeping it as the single source of truth
- Arrow key DOWN bound is capped at 5 or 6 depending on the flag
- ENTER case 6 and CHAR:6 shortcut both guard on MAIN_MENU_SHOW_UPDATE -> pressing 6 when no update is available is a no-op, consistent with the hidden option
- tests/completion.bats: adds update and update:Update to latest version assertions to the bash and zsh completion tests

### DIFF
--- a/mole
+++ b/mole
@@ -828,6 +828,7 @@ interactive_main_menu() {
                 exec "$SCRIPT_DIR/bin/status.sh"
                 ;;
             "CHAR:6")
+                [[ "${MAIN_MENU_SHOW_UPDATE:-false}" == "true" ]] || continue
                 show_cursor
                 clear
                 update_mole


tw93:
@xde013 Thanks a lot for the contribution and for putting this together.

I didn’t merge #540 directly, but I applied the underlying fix on `main` and linked it as a follow-up to #540.

The main reason is that the PR included more than this specific menu behavior, and the update action could still be triggered through the `U` shortcut even when the entry was hidden. That behavior is now fixed on `main`, and I also added test coverage for both the hidden and visible cases.

So there’s no need to merge #540 for this specific change anymore, but I really appreciate the work here.

Issue #448, opened 2026-02-12
[FEATURE] add Mole to nixpkgs


tolkonepiu:
## Feature description

Please publish Mole as an official package in Nixpkgs so macOS users can install it via nixpkgs / home-manager instead of relying on Homebrew or unofficial overlays.

## Use case

I use Nix on macOS and prefer installing CLI tools from the official Nix package set for reproducibility and clean, declarative setup.
Right now, Mole is easy to install via Homebrew, but not available as an official nixpkgs package, which makes it harder to manage in Nix-based macOS setups.

## Proposed solution

Add Mole to [NixOS/nixpkgs](https://github.com/NixOS/nixpkgs) as a darwin-only package (platforms = darwin).

## Alternatives considered

Maintain a private Nix overlay/flake (works, but not “official” and requires extra maintenance).

tw93:
@tolkonepiu Thanks for the suggestion!                                
                                                                                                                                                                                 
At this time, we don't have plans to maintain an official Nixpkgs package. Mole follows a minimal maintenance philosophy — the current ./install.sh and Homebrew distribution cover most users without adding ongoing packaging overhead.

If there's significant community demand or someone wants to contribute a flake.nix or nixpkgs definition, we're open to reviewing PRs. The project is pure shell/Go, so packaging should be straightforward for anyone familiar with Nix.

Closing for now, but happy to revisit if more users request this.

daniel-got:
i think thats a good option tho, to add support for nix package manager

mikewebbtech:
## Shameless plug:

If this helps. I have created a flake.nix for my own use that peeps are welcome to use and give feedback on.

If it gets used and any issues that have missed get picked up and fixed, I might look at getting it into the official nix packages site.

Have a look and a go and give some feedback

[https://github.com/mikewebbtech/mole-nix](https://github.com/mikewebbtech/mole-nix/)

Thanks again to @tw93 for an awesome app.

TyceHerrman:
https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/mo/mole-cleaner/package.nix

PR #573, opened 2026-03-15
Add support for nix package manager

sughodke:
## Summary

New flake will prefer to use prebuilt binaries from Github. However this requires updating to hash with every release. A script to do so is provided with this commit. The CI/CD workflow has been updated to call this script.

Fixes issue https://github.com/tw93/Mole/issues/448

## Safety Review

No

No

## Tests

All existing automated tests pass

| Test                                   | Result                              |
|----------------------------------------|-------------------------------------|
| nix build (prebuilt)                   | Pass                                |
| nix build .#from-source                | Pass                                |
| mole --version (default)               | 1.30.0                              |
| mole --version (from-source)           | 1.30.0                              |
| mo --version (alias)                   | 1.30.0                              |
| Go binaries present (analyze-go, status-go) | Pass                           |
| Shell libs installed                   | Pass                                |
| update-flake-hashes.sh                 | Pass                                |
| nix develop (devShell)                 | Pass (Go 1.25.7, Bash 5.3, shellcheck, bats) |

## Safety-related changes

- None.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -103,6 +103,42 @@ jobs:
           draft: false
           prerelease: false
 
+  update-flake:
+    name: Update Nix Flake Hashes
+    runs-on: ubuntu-latest
+    needs: release
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v4
+        with:
+          ref: main
+
+      - name: Extract version from tag
+        id: version
+        run: |
+          TAG=${GITHUB_REF#refs/tags/}
+          VERSION=${TAG#V}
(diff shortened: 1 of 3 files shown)


tw93:
@sughodke Thanks again for the contribution.

I took another look at the PR and want to clarify the reason for closing it.

This PR is not about adding Linux runtime support for Mole. It adds Nix packaging support for macOS, including a new `flake.nix` and a release workflow step to keep the flake hashes updated automatically.

I am closing it for now because I do not want to take on this additional distribution and release-maintenance surface in the current release pipeline. The PR also changes the release workflow itself, so the maintenance cost is higher than a normal packaging-only change.

There is also a script issue in the new hash update step that would need to be fixed before this could be considered safely.

Thanks again for the work and the detailed write-up.

PR #578, opened 2026-03-16
protect macOS wallpaper assets from default cleanup

### LINKED ISSUE #577: [BUG] `mo clean` deletes downloaded macOS wallpaper assets under `~/Library/Application Support/   com.apple.wallpaper`
Summary:
  `mo clean` removes previously downloaded macOS wallpaper / aerial video assets from the user Library. These
  are not disposable app caches from a user perspective. Deleting them forces users to re-download large
  wallpaper files again from System Settings > Wallpaper.

  Environment:
  - Mole version: 1.30.0
  - Install method: Homebrew
  - macOS: 26.3.1
  - Architecture: arm64

  What happened:
  After running `mo clean`, multiple downloaded wallpaper video files were removed from:

  `~/Library/Application Support/com.apple.wallpaper/aerials/videos/`

  This caused previously downloaded system wallpapers / aerial wallpapers to disappear and require downloading
  again.

  Reproduction steps:
  1. Open macOS System Settings > Wallpaper.
  2. Download several Apple wallpapers / aerial wallpapers so that files exist under:
     `~/Library/Application Support/com.apple.wallpaper/aerials/videos/`
  3. Run:
     `mo clean`
  4. Check the wallpaper asset directory again, or inspect `~/.config/mole/operations.log`.

  Expected behavior:
  `mo clean` should not delete downloaded macOS wallpaper assets by default.

  Actual behavior:
  `mo clean` deletes them as part of cleanup.

  Evidence:
  Mole currently hardcodes this path in cleanup logic:

  `safe_clean ~/Library/Application\ Support/com.apple.wallpaper/aerials/videos/* "Aerial wallpaper videos"`

  From local installation:
  - `/opt/homebrew/Cellar/mole/1.30.0/libexec/lib/clean/user.sh:430-431`

  My local operation log shows real deletions, for example on 2026-03-16:
  - `REMOVED /Users/sure/Library/Application Support/com.apple.wallpaper/aerials/videos/03EC0F5E-CCA8-4E0A-9FEC-
  5BD1CE151182.mov`
  - `REMOVED /Users/sure/Library/Application Support/com.apple.wallpaper/aerials/videos/2F72BC1E-3D76-456C-81EB-
  842EBA488C27.mov`
  - `REMOVED /Users/sure/Library/Application Support/com.apple.wallpaper/aerials/videos/D7950C0A-27B9-4034-AB5B-
  EDB89D41341A.mov`

  At the same time, Mole skips only the wallpaper cache path:
  - `~/Library/Containers/com.apple.wallpaper.agent/Data/Library/Caches/com.apple.wallpaper.caches`

  Why this is a problem:
  - These files are user-downloaded assets, not obviously disposable cache.
  - Re-downloading them is slow and wasteful.
  - Users can reasonably assume wallpapers they explicitly downloaded will be preserved.
  - Mole already protects many Apple-managed paths, so this behavior is surprising and inconsistent.

  Suggested fix:
  One of these would solve it:
  1. Protect `~/Library/Application Support/com.apple.wallpaper/` by default.
  2. Protect `~/Library/Application Support/com.apple.idleassetsd/` by default as well.
  3. Move wallpaper asset cleanup behind an explicit opt-in flag instead of including it in default `mo clean`.
  4. At minimum, classify it as non-default / higher-risk cleanup with a warning.

  Additional note:
  I understand these files can be regenerated, but from a user perspective they behave like downloaded assets,
  not temporary cache. Default cleanup should preserve them.

Sure-Will:
Fixes #577

## Summary

  Protect downloaded macOS wallpaper assets from default cleanup.

  `mo clean` currently cleans wallpaper-related assets under:

  - `~/Library/Application Support/com.apple.wallpaper/aerials/videos/*`
  - `~/Library/Application Support/com.apple.idleassetsd/*`
  - `/Library/Application Support/com.apple.idleassetsd/Customer/*`

  In practice, these behave more like user-downloaded system assets than disposable cache. When they are
  removed, macOS needs to download them again from System Settings > Wallpaper.

  ## What this PR changes

  - add wallpaper asset paths to the default cleanup whitelist
  - add corresponding entries to `mo clean --whitelist`
  - update nearby comments to reflect the more conservative default behavior

  ## Why

  On my machine, `mo clean` removed multiple files from:

  `~/Library/Application Support/com.apple.wallpaper/aerials/videos/`

  That caused previously downloaded wallpapers to disappear and require large re-downloads.

  Mole already uses safety-first defaults in many other places. This change makes the wallpaper behavior more
  consistent with that approach and avoids surprising data loss from a user perspective.

  ## Notes

  This PR does not remove the underlying cleanup logic. It only protects these paths by default, so advanced
  users can still opt out if they explicitly want to clean them.

### DIFF
--- a/lib/clean/user.sh
+++ b/lib/clean/user.sh
@@ -444,19 +444,22 @@ clean_support_app_data() {
         safe_find_delete "$crash_reporter_dir" "*" "$support_age_days" "f" || true
     fi
 
-    # Keep recent wallpaper assets to avoid large re-downloads.
+    # Wallpaper assets are user-downloaded resources. Protect them by default
+    # via the whitelist to avoid expensive re-downloads, while still allowing
+    # advanced users to opt out of protection manually.
     local idle_assets_dir="$HOME/Library/Application Support/com.apple.idleassetsd"
     if [[ -d "$idle_assets_dir" && ! -L "$idle_assets_dir" ]]; then
         safe_find_delete "$idle_assets_dir" "*" "$support_age_days" "f" || true
     fi
 
-    # Clean system-level idle/aerial screensaver videos (macOS re-downloads as needed).
+    # System-level idle assets are also protected by the default whitelist.
     local sys_idle_assets_dir="/Library/Application Support/com.apple.idleassetsd/Customer"
     if sudo test -d "$sys_idle_assets_dir" 2> /dev/null; then
         safe_sudo_find_delete "$sys_idle_assets_dir" "*" "$support_age_days" "f" || true
     fi
 
-    # Clean old aerial wallpaper videos (can be large, safe to remove).
+    # Downloaded aerial wallpaper videos are protected by default whitelist
+    # entries because users expect them to persist after cleanup.
     safe_clean ~/Library/Application\ Support/com.apple.wallpaper/aerials/videos/* "Aerial wallpaper videos"
 
     # Do not touch Messages attachments, only preview/sticker caches.
--- a/lib/core/base.sh
+++ b/lib/core/base.sh
@@ -105,6 +105,9 @@ declare -a DEFAULT_WHITELIST_PATTERNS=(
     "$HOME/Library/Caches/com.apple.spotlight*"
     "$HOME/Library/Caches/com.apple.Spotlight*"
     "$HOME/Library/Caches/CloudKit*"
+    "$HOME/Library/Application Support/com.apple.wallpaper/*"
+    "$HOME/Library/Application Support/com.apple.idleassetsd/*"
+    "/Library/Application Support/com.apple.idleassetsd/Customer/*"
     "$FINDER_METADATA_SENTINEL"
 )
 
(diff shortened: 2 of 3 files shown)


Sure-Will:
I noticed earlier discussion around cleaning these large wallpaper assets, and I understand why they were added to default cleanup. This PR only proposes a more conservative default because, from a user perspective, they can feel closer to downloaded assets than disposable cache.

tw93:
@Sure-Will Thanks again for the detailed report and the follow-up PR.

I am going to close this PR because the main issue has already been fixed directly on `main` with a smaller change: `mo clean` no longer removes `~/Library/Application Support/com.apple.wallpaper/aerials/videos/*` by default.

I appreciate the time and care you put into this. The broader whitelist-based approach here would need additional follow-up for existing whitelist behavior, so I do not want to merge it in its current form.

Thanks again for helping catch and improve this behavior.

Sure-Will:
Thanks for the clear explanation and for landing the fix on `main`. I understand the decision, and I’m glad this behavior is improved now. Wishing Mole continued success🎉.

PR #584, opened 2026-03-17
Add --all flag to mo purge

MikeSilvis:
## Summary

- Adds `--all` (`-a`) flag to `mo purge` that selects all discovered artifacts (including recent ones) and skips both the interactive picker and confirmation prompt
- Useful for scripting and automation (e.g. running `mo purge --all` in a cleanup script) where interactive selection is not practical
- Can be combined with `--dry-run` to preview what would be cleaned: `mo purge --all --dry-run`

## Changes

- `bin/purge.sh`: Parse `--all`/`-a` flag, export `MOLE_PURGE_ALL=1`, update help text
- `lib/clean/project.sh`: When `MOLE_PURGE_ALL=1`, select all items (including recent) and skip interactive prompt + confirmation
- `tests/purge.bats`: Added 2 tests for the new flag

## Test plan

- [x] All 59 existing tests pass
- [x] New test: `mo purge: accepts --all flag`
- [x] New test: `clean_project_artifacts: --all selects all items including recent`
- [x] Manual test: `mo purge --all --dry-run` shows all items selected without prompting
- [x] Manual test: `mo purge --all` cleans everything without prompting
- [x] Manual test: `mo purge` (without --all) still shows interactive picker as before

### DIFF
--- a/bin/purge.sh
+++ b/bin/purge.sh
@@ -275,6 +275,7 @@ show_help() {
     echo -e "${YELLOW}Usage:${NC} mo purge [options]"
     echo ""
     echo -e "${YELLOW}Options:${NC}"
+    echo "  --all           Select all artifacts without prompting"
     echo "  --paths         Edit custom scan directories"
     echo "  --dry-run       Preview purge actions without making changes"
     echo "  --debug         Enable debug logging"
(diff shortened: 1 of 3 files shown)


tw93:
@MikeSilvis Thanks for the PR. I do not think we should merge this as-is.

`mo purge` is intentionally conservative: recent artifacts are left unselected by default, and interactive deletion requires confirmation. Non-interactive runs already cover the automation case for non-recent artifacts.

So this is not just a convenience flag. It also enables unattended deletion of recent artifacts and skips confirmation, which weakens the current safety model.

I do not see a new low-level safety issue, since deletion still goes through the existing guards, but the behavior is too aggressive for a generic `--all` flag.

PR #593, opened 2026-03-18
Fix: Removed duplicate calls to clean_xcode_tools and clean_code_editors

### LINKED ISSUE #590: [BUG] mo clean --dry-run double-counts items in preview (case-variant project roots and overlapping sections)
Mole 1.30.0 on macOS 26.3.1 (arm64), normal case-insensitive APFS.

`mo clean --dry-run` appears to overcount some items in the preview and in `~/.config/mole/clean-list.txt`.

I found two causes:

1. Case-variant project roots are treated as different roots
- `clean_project_caches` collects roots from defaults/config + home auto-discovery.
- Defaults include `~/Code`.
- My actual on-disk directory is `~/code`.
- On case-insensitive APFS, both exist as path aliases to the same directory, but the code only uses string dedupe (`sort -u` / exact path matching), not canonical/inode dedupe.

Concrete evidence from my run:
- `/Users/cdeil/Code/stocadro/frontend_flutter/apps/mobile/build  # 3.53GB`
- `/Users/cdeil/code/stocadro/frontend_flutter/apps/mobile/build  # 3.53GB`

They are the same filesystem object:
- same inode/dev for both paths

Minimal repro on macOS:
1. Create lowercase `~/code/demo/pubspec.yaml` and `~/code/demo/.dart_tool`
2. Put `~/Code` in `~/.config/mole/purge_paths`
3. Run the project cache scan / `mo clean --dry-run`
4. The same `.dart_tool` is reported twice, once under `/Code/...` and once under `/code/...`

2. Some exact same paths are scanned twice across sections
- `clean_developer_tools` calls `clean_xcode_tools` and `clean_code_editors`
- `clean_user_gui_applications` calls the same functions again

Examples from my dry-run list:
- `~/Library/Application Support/Code/Cache`
- `~/Library/Application Support/Code/CachedData`
- `~/Library/Logs/CoreSimulator`
- `~/Library/Caches/com.apple.dt.Xcode`

Expected:
- dry-run preview should dedupe by canonical path or inode on macOS
- the same cleanup target should not be counted twice across sections

Actual:
- preview list and totals include duplicate entries
- at least part of the reported reclaimable space is overstated

jason-costello:
Fixes #590

### DIFF
--- a/tests/clean_app_caches.bats
+++ b/tests/clean_app_caches.bats
@@ -81,6 +81,8 @@ source "$PROJECT_ROOT/lib/core/common.sh"
 source "$PROJECT_ROOT/lib/clean/app_caches.sh"
 stop_section_spinner() { :; }
 safe_clean() { :; }
+clean_xcode_tools() { echo "xcode"; }
+clean_code_editors() { echo "editors"; }
 clean_communication_apps() { echo "comm"; }
 clean_dingtalk() { echo "dingtalk"; }
 clean_ai_apps() { echo "ai"; }
@@ -88,6 +90,8 @@ clean_user_gui_applications
 EOF
 
     [ "$status" -eq 0 ]
+    [[ "$output" != *"xcode"* ]]
+    [[ "$output" != *"editors"* ]]
     [[ "$output" == *"comm"* ]]
     [[ "$output" == *"dingtalk"* ]]
     [[ "$output" == *"ai"* ]]


tw93:
@jason-costello Thanks for the fix. 

I added a follow-up commit before merging to strengthen the regression test in tests/clean_app_caches.bats, so the PR now explicitly verifies that clean_user_gui_applications() no longer calls clean_xcode_tools or clean_code_editors while still covering the remaining application groups.

PR #594, opened 2026-03-18
fix(ui): use system sleep for inline spinner

### LINKED ISSUE #591: [BUG] Spinner subprocess fails to find `sleep` when PATH contains non-standard directories
## Describe the bug

The inline spinner in `lib/core/ui.sh` spawns a background subshell (`( ... ) &` + `disown`) that calls `sleep 0.05` at line 345. When the user's `sleep` binary is provided by a non-standard PATH entry (e.g. Nix), the detached subshell fails to resolve it:

```
Repairing disk permissions.../opt/homebrew/Cellar/mole/1.30.0/libexec/lib/core/ui.sh: line 345: /Users/<user>/.nix-profile/bin/sleep: No such file or directory
```

The root cause is that the backgrounded subshell inherits the user's `$PATH` which points `sleep` to a Nix store path. During the `disown`ed subprocess execution, this path may become stale or unresolvable depending on the shell environment.

A simple fix would be to use `/bin/sleep` (always available on macOS) instead of the bare `sleep` command in the spinner loop at lines 345 and 369 of `lib/core/ui.sh`.

## Steps to reproduce

1. Install Mole via Homebrew on a system where `sleep` is provided by Nix (`which sleep` → `~/.nix-profile/bin/sleep`)
2. Run `mo repair`
3. See error at the spinner: `line 345: /Users/<user>/.nix-profile/bin/sleep: No such file or directory`

## Expected behavior

The spinner should work regardless of how `sleep` is provided in `$PATH`. Using `/bin/sleep` would avoid any PATH resolution issues in detached subshells.

## Environment

```text
Mole version 1.30.0
macOS: 26.3.1
Architecture: arm64
Kernel: 25.3.0
SIP: Enabled
Install: Homebrew
Shell: /bin/zsh
```

## Additional context

The issue affects any user whose `$PATH` resolves `sleep` to a non-system location (Nix, Homebrew coreutils, etc.). Both occurrences of `sleep` in `ui.sh` (lines 345 and 369) are affected.

tw93:
## Summary
- use `/bin/sleep` for inline spinner delays so detached spinner processes do not depend on user PATH
- add a TTY regression test that injects a fake PATH `sleep` and verifies the spinner ignores it

## Testing
- bats tests/core_common.bats

Fixes #591

### DIFF
--- a/lib/core/ui.sh
+++ b/lib/core/ui.sh
@@ -342,7 +342,7 @@ start_inline_spinner() {
                 # Output to stderr to avoid interfering with stdout
                 printf "\r${MOLE_SPINNER_PREFIX:-}${BLUE}%s${NC} %s" "$c" "$display_message" >&2 || break
                 i=$((i + 1))
-                sleep 0.05
+                /bin/sleep 0.05
             done
 
             # Clean up stop file before exiting
@@ -366,7 +366,7 @@ stop_inline_spinner() {
         # Wait briefly for cooperative exit
         local wait_count=0
         while kill -0 "$INLINE_SPINNER_PID" 2> /dev/null && [[ $wait_count -lt 5 ]]; do
-            sleep 0.05 2> /dev/null || true
+            /bin/sleep 0.05 2> /dev/null || true
             wait_count=$((wait_count + 1))
         done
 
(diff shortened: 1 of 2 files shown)


tw93:
Submitted directly to `main` in 3b46b4a.

PR #599, opened 2026-03-19
fix: pass through sudo's native prompt instead of hardcoding 'Password:'

### LINKED ISSUE #597: [BUG] Wrong output when elevating within mole instead.
## Describe the bug

If you use a different manner of utilizing `sudo` instead of just password, like a yubikey with PIN etc. The "Password:" dialog is just "wrong" output from `mole`. Fixing this just would just remove misconception. Entering the PIN works when it asks for the "Password:".

## Steps to reproduce

1. `mole`
2. Select step 1
3. Asks for sudo password.

## Expected behavior

Send output from `sudo` directly back instead of statically showing `password:`

## Additional context

Here you can see how it is on my machine.

Image

yetval:
Fixes #597

Previously `_request_password` hardcoded `Password:` as the prompt, which is wrong for auth methods like YubiKey that expect a PIN prompt.

Now we call `sudo -v` directly on the TTY and let it handle everything — prompting, echo, and retries. The correct prompt (whatever PAM outputs) shows up naturally.

### DIFF
--- a/lib/core/sudo.sh
+++ b/lib/core/sudo.sh
@@ -57,6 +57,8 @@ _request_password() {
 
     echo -e "${PURPLE}${ICON_ARROW}${NC} Enter your credentials:" > "$tty_path"
 
+    # shellcheck disable=SC2024,SC2094
+    # Intentionally route sudo's native prompt to the same TTY device it reads from.
     if sudo -v < "$tty_path" > /dev/null 2> "$tty_path"; then
         return 0
     fi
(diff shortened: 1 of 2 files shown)


tw93:
@yetval I did one follow-up fix before merging: adjusted the clamshell cleanup path so it clears the correct number of prompt lines when the Touch ID hint is shown, and added two regression tests in `tests/manage_sudo.bats` to cover the clamshell cleanup counts with and without the Touch ID hint. 

I also added a short ShellCheck suppression comment around the intentional TTY read/write routing for native `sudo` prompt passthrough. I re-ran formatting and the targeted sudo tests before merging. Thanks for the improvement here.

weq:
image

fyi works perfectly now :D

PR #625, opened 2026-03-23
feat: colorize cleaned lines green/yellow for above/below 1GB

HaraldNordgren:
This was it's easier to eyeball output and see at a glance which folders were cleaned the most. Currently, it's a bit cumbersome.

### DIFF
--- a/bin/clean.sh
+++ b/bin/clean.sh
@@ -117,7 +117,6 @@ if [[ ${#WHITELIST_PATTERNS[@]} -gt 0 ]]; then
 fi
 
 # Section tracking and summary counters.
-readonly MOLE_ONE_GIB_KB=$((1024 * 1024))
 total_items=0
 TRACK_SECTION=0
 SECTION_ACTIVITY=0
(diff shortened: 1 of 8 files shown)


tw93:
Merged. I kept the feature and followed up with a small fix so the color threshold now matches the displayed decimal `GB` units, which avoids `1.00GB` lines still showing up yellow.

The merged commit is `9757063`. I also reran the project checks locally before landing it.

Thanks for the contribution.

HaraldNordgren:
Thanks for @tw93, and thanks for a great project! ❤️

tw93:
Thanks for the contribution, Harald! The idea of making large cleanups visually stand out is solid.

After shipping it, the feedback we got is that the yellow color feels like a warning rather than a positive signal, which caused some confusion. The uniform green output actually reads more naturally as a success indicator regardless of size.

Reverting this for now, but appreciate the thinking behind it.

HaraldNordgren:
Hi @tw93, I tried it now and it seems quite broken.

- "Chrome Service Worker, 1935MB" not colored at all
- "Xcode derived data 5 items, 6.25GB" colored yellow despite being above 1GB

tw93:
Thanks for the detailed report, @HaraldNordgren, you were right to call this out.

I reproduced the behavior locally and fixed it in follow-up changes:
- Service Worker cleanup output now uses the cleanup success color logic (same path as other cleanup lines).
- I also fixed a scan fallback edge case in purge: when fd output is filtered to empty, we now fall back to find instead of treating it as success.
- Added/updated tests around these paths and reran the full suite.

Validation on latest main:
- `bats tests/purge.bats`: pass (59/59)
- `MOLE_TEST_NO_AUTH=1 ./scripts/test.sh`: pass (545 tests, 0 failures)

Fix is now on `main` in commit `29d6829`.

If you still see inconsistent color output on your side, please share your exact Mole version and a short output snippet, and I will continue to dig in.

HaraldNordgren:
Hi @tw93, this is still not working in 1.34.0 😞

PR #637, opened 2026-03-24
feat(clean): add deep Brave Browser cleanup support

### LINKED ISSUE #632: [FEATURE] support clean brave browser files like edge\chrome

pranahonk:
## Summary
- Adds deep cache cleaning for Brave Browser: app cache, GPU cache, component CRX cache, shader cache, GR shader cache, and Dawn cache — matching existing Chrome and Helium depth
- Adds `clean_brave_old_versions()` to remove stale framework versions from `Brave Browser Framework.framework/Versions/` while keeping `Current` — matching existing Chrome/Edge old version cleanup
- Respects whitelist, dry-run mode, process-running check, and sudo session

Closes #632

## Test plan
- [ ] Verify `./scripts/check.sh` passes (shellcheck, shfmt, syntax)
- [ ] Verify `./scripts/test.sh` has no new failures
- [ ] Test with Brave Browser installed — confirm caches are cleaned
- [ ] Test with Brave Browser running — confirm old versions cleanup is skipped with warning
- [ ] Test with `--dry-run` — confirm sizes are reported without deletion

### DIFF
--- a/lib/clean/user.sh
+++ b/lib/clean/user.sh
@@ -396,6 +396,88 @@ clean_edge_updater_old_versions() {
     fi
 }
 
+# Remove old Brave Browser versions while keeping Current.
+clean_brave_old_versions() {
+    local -a app_paths=(
+        "/Applications/Brave Browser.app"
+        "$HOME/Applications/Brave Browser.app"
+    )
+
+    # Match the exact Brave process name to avoid false positives
+    if pgrep -x "Brave Browser" > /dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


tw93:
@pranahonk LGTM! The implementation closely mirrors the existing Chrome/Edge pattern which is exactly what we want. Good coverage across AppCache, GPUCache, ShaderCache, GrShaderCache, DawnCache, and CRX — actually more thorough than Chrome's current depth. Thanks for the contribution!

One minor thing to note for a follow-up: Chrome itself is missing the shader/dawn cache entries that Brave now has — worth adding parity there separately.

PR #640, opened 2026-03-25
Add safety guards to prevent cleanup of active system state

abcreativ:
## Summary

Mole can inadvertently break active system state during cleanup. This PR adds safety guards across several cleanup modules:

- **Wallpaper/screensaver preservation** — Detects the active wallpaper via `defaults read com.apple.wallpaper` and the per-space `Store/Index.plist`, then excludes those assets from cleanup. Also skips wallpaper agent cache when `WallpaperAgent` is running. Previously, running `mole clean` would reset the user's wallpaper to default.
- **iOS Simulator running check** — Skips Simulator cache, temp, and log cleanup when the Simulator process is active (prevents mid-session crashes).
- **mdfind timeout increase (2s → 5s)** — The 2-second timeout caused false orphan detection on systems with slow Spotlight indexes, leading to deletion of data for apps that are actually installed.
- **Dynamic Claude VM bundle discovery** — Uses `find` instead of a hardcoded path, so it works across Claude Desktop versions.
- **Expanded service worker domain protection** — Adds Google Workspace, GitHub, GitLab, Notion, Figma, and other offline/PWA-capable sites to the protected list.
- **macOS installer version check** — Won't delete installer apps that match the currently running macOS major version (recovery/reinstall safety).
- **Chrome/Edge broken symlink guard** — If the `Current` symlink target doesn't exist, skips version cleanup entirely instead of potentially deleting the active browser version.
- **Mail attachment message improvement** — Shows the age threshold in the cleanup message for transparency.

## Test plan

- [ ] Run `mole clean --dry-run` and verify no errors
- [ ] Verify wallpaper is preserved after `mole clean` (set an aerial wallpaper first)
- [ ] Run with Simulator open — confirm skip message appears
- [ ] Run on a system with slow Spotlight — confirm no false orphan deletions
- [ ] Verify Chrome/Edge version cleanup still works when symlink is valid
- [ ] Verify macOS installer matching current version is kept

### DIFF
--- a/lib/clean/system.sh
+++ b/lib/clean/system.sh
@@ -120,7 +120,7 @@ clean_deep_system() {
     # and not matching the currently installed macOS version (recovery safety).
     local installer_cleaned=0
     local current_macos_version=""
-    current_macos_version=$(sw_vers -productVersion 2>/dev/null | cut -d. -f1 || true)
+    current_macos_version=$(sw_vers -productVersion 2> /dev/null | cut -d. -f1 || true)
     for installer_app in /Applications/Install\ macOS*.app; do
         [[ -d "$installer_app" ]] || continue
         local app_name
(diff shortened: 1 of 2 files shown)


tw93:
@abcreativ Thanks for this PR! The improvements you added are solid and are now merged.

A couple of things were adjusted on your branch before merging. The wallpaper preservation logic used raw `find -delete` which bypasses the project's safety helpers, so that section was reverted to use `safe_find_delete` as before. The WallpaperAgent running guard you added already handles the active-wallpaper protection case cleanly. The macOS installer version check was also updated to read `DTPlatformVersion` instead of `DTSDKBuild`, since the latter is a build number rather than a version string and the comparison wasn't matching correctly.

Everything else landed as-is: the expanded service worker domain list, Simulator running guard, mdfind timeout increase, dynamic Claude VM bundle discovery, and the Chrome/Edge broken symlink guard. These are all great additions. Thanks again!

PR #648, opened 2026-03-27
fix(purge): sort `item_display_paths` by size

### LINKED ISSUE #647: [BUG] `purge` paths are out of sync
**Description**

When running `mo purge`, the interactive directory list is sorted by size (largest first), but the "Full path" footer and the final selection operate on the **original unsorted discovery order**. This means:

- The "Full path" shown at the bottom of the screen does not match the currently highlighted item.
- Pressing Space to select items and then Enter to confirm purges **different directories** than what was highlighted/selected in the UI.

**Steps to reproduce**

1. Run `mo purge` in a project with multiple artifact directories of varying sizes.
2. Navigate the list with ↑/↓ — observe that the "Full path" shown at the bottom does not match the highlighted entry.
3. Select a few items with Space and confirm with Enter — the paths reported for deletion differ from what was selected.

**Root cause**

In `lib/clean/project.sh`, after discovering artifacts, the list is sorted by size descending. During the sort, all parallel arrays are reordered — `menu_options`, `item_paths`, `item_sizes`, `item_size_unknown_flags`,
`item_recent_flags` — **but `item_display_paths` is omitted from the sort loop**. This array is then assigned to `PURGE_CATEGORY_FULL_PATHS_ARRAY` and used to render the "Full path" footer, leaving it in the original unsorted
order while everything else reflects the sorted order.

**Expected behaviour**

The "Full path" footer always matches the highlighted item, and confirming a selection purges exactly the directories that were selected.

**Fix**

Include `item_display_paths` in the sort reorder loop alongside the other arrays so all parallel arrays share a consistent order before being passed to the interactive selector.

## Environment

```text
Mole version 1.31.0
macOS: 15.7.4
Architecture: arm64
Kernel: 24.6.0
SIP: Enabled
Disk Free: 201Gi
Install: Homebrew
Shell: /bin/zsh
```

mshavliuk:
closes #647

## Summary

Fixes a sort-order mismatch in `mo purge` where the "Full path" footer and the confirmation dialog showed paths that didn't correspond to the highlighted/selected items.

## Root cause

`clean_project_artifacts` sorts all discovered artifacts by size (descending) before showing the interactive selector. During this reorder, `item_display_paths` — the array used to populate `PURGE_CATEGORY_FULL_PATHS_ARRAY` (the
"Full path" footer) and the confirmation dialog — was not included in the sort loop. Every other parallel array (`menu_options`, `item_paths`, `item_sizes`, `item_size_unknown_flags`, `item_recent_flags`) was reordered
correctly, leaving `item_display_paths` stuck in the original alphabetical discovery order.

## What broke

- **"Full path" footer** showed the path for a different item than the one currently highlighted.
- **Confirmation dialog** listed display paths that didn't match what the user had selected.

## Fix

Added `item_display_paths` / `sorted_item_display_paths` to the sort loop in `lib/clean/project.sh`, keeping it in sync with all other parallel arrays.

## Tests

Two regression tests added to `tests/purge.bats`:

- **`sort: PURGE_CATEGORY_FULL_PATHS_ARRAY[0] is the largest artifact after size-descending sort`** — creates a small project (`alpha`) and a large project (`beta`), runs `clean_project_artifacts` under a pseudo-terminal,
intercepts `select_purge_categories` to capture `PURGE_CATEGORY_FULL_PATHS_ARRAY`, and asserts index 0 is `beta` (the largest).
- **`sort: PURGE_CATEGORY_FULL_PATHS_ARRAY and PURGE_CATEGORY_SIZES indices are consistent`** — same setup, additionally asserts that sizes are sorted descending and that every path index aligns with the corresponding size
index.

Both tests failed before the fix and pass after.

### DIFF
--- a/lib/clean/project.sh
+++ b/lib/clean/project.sh
@@ -1465,6 +1465,7 @@ clean_project_artifacts() {
         local -a sorted_item_sizes=()
         local -a sorted_item_size_unknown_flags=()
         local -a sorted_item_recent_flags=()
+        local -a sorted_item_display_paths=()
 
         for idx in "${sorted_indices[@]}"; do
             sorted_menu_options+=("${menu_options[idx]}")
@@ -1472,6 +1473,7 @@ clean_project_artifacts() {
             sorted_item_sizes+=("${item_sizes[idx]}")
             sorted_item_size_unknown_flags+=("${item_size_unknown_flags[idx]}")
             sorted_item_recent_flags+=("${item_recent_flags[idx]}")
+            sorted_item_display_paths+=("${item_display_paths[idx]}")
         done
 
         # Replace original arrays with sorted versions
@@ -1480,6 +1482,7 @@ clean_project_artifacts() {
         item_sizes=("${sorted_item_sizes[@]}")
         item_size_unknown_flags=("${sorted_item_size_unknown_flags[@]}")
         item_recent_flags=("${sorted_item_recent_flags[@]}")
+        item_display_paths=("${sorted_item_display_paths[@]}")
     fi
     if [[ -t 1 ]]; then
         stop_inline_spinner
--- a/tests/purge.bats
+++ b/tests/purge.bats
@@ -954,3 +954,143 @@ EOF
 	rm -f "$scan_output"
 	[[ "$result" == "SKIPPED" ]]
 }
+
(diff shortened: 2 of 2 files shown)


tw93:
@mshavliuk LGTM, I checked it myself, and it looks good.

PR #650, opened 2026-03-28
feat(ui): add Ctrl+P/Ctrl+N navigation to menus

bybruno:
## Summary

Add `Ctrl+N` and `Ctrl+P` support to Mole's interactive menus through the shared key reader.

## Motivation

Mole already supports arrow keys and Vim-style bindings for menu navigation. This change extends that keyboard support with `Ctrl+N` and `Ctrl+P`, which are familiar navigation shortcuts in terminal and Readline-style workflows.

The goal is not to replace the current controls, but to make the interface more flexible for users who already rely on those shortcuts as part of their normal terminal muscle memory.

## Changes

- map `Ctrl+N` to next-item navigation
- map `Ctrl+P` to previous-item navigation
- preserve existing arrow-key and Vim-style navigation
- update menu hints to show the new shortcuts
- update README navigation documentation
- add key-mapping coverage for normal and forced-char modes


tw93:
@bybruno Thanks for the contribution and the clear write-up! The implementation is clean and the tests are well structured.

That said, I think this one adds more complexity than it's worth for Mole. The menus already support arrow keys and Vim-style `j/k`, which cover the same navigation need for the vast majority of users. Adding `Ctrl+P/N` means maintaining two extra key bindings and,more visibly, the hint text grows from the clean `↑↓` to `↑↓/Ctrl+P/N`, which makes every menu footer a bit noisier.

Mole aims to stay simple and minimal, so I'd rather keep the hint text clean and the key-handling surface small. Closing for now, but appreciate you taking the time to put this together!

PR #659, opened 2026-03-30
feat(doctor): add mo doctor command for Mac health diagnostics

sebastianbreguel:
## Summary

`mo doctor` — a TUI-based Mac health diagnostic that scores your system across 7 categories (100 points total). Categories appear one-by-one with a spinner during scanning, then transition to an interactive view where you can expand each category for details.

### Storage (15 pts)
| Check | What it measures | Pass | Warn | Fail |
|-------|-----------------|------|------|------|
| Free space | Disk space available | ≥20% free | 10–20% | <10% |
| Recoverable cache | ~/Library/Caches + Logs size | ≤2 GB | 2–5 GB | >5 GB |
| Purgeable space | macOS purgeable storage | Informational | — | — |
| node_modules | Total size across directories | <5 GB | 5–10 GB | >10 GB |
| iOS backups | MobileSync backup size | <5 GB | ≥5 GB | — |

### Performance (20 pts)
| Check | What it measures | Pass | Warn | Fail |
|-------|-----------------|------|------|------|
| RAM usage | Wired + compressed memory | ≤60% | 60–80% | >80% |
| Swap usage | Active swap in use | ≤0.5 GB | 0.5–2 GB | >2 GB |
| Uptime | Days since last restart | ≤7 days | 7–14 days | >14 days |

### Battery (15 pts) — *skipped on desktops*
| Check | What it measures | Pass | Warn | Fail |
|-------|-----------------|------|------|------|
| Cycle count | Battery charge cycles | ≤500 | 500–900 | >900 |
| Health | Capacity vs design capacity | ≥90% | 80–90% | <80% |

### Security (15 pts)
| Check | What it measures | Pass | Warn | Fail |
|-------|-----------------|------|------|------|
| FileVault | Disk encryption | Enabled | — | Disabled |
| Firewall | System firewall | Enabled | — | Disabled |
| SIP | System Integrity Protection | Full | Partial | Disabled |
| macOS version | Current vs latest release | Current | 1 behind | 2+ behind |

### Maintenance (15 pts)
| Check | What it measures | Pass | Warn | Fail |
|-------|-----------------|------|------|------|
| Broken launch agents | LaunchAgents with missing executables | 0 | 1–3 | >3 |
| Unused apps (90+ days) | Apps not opened in 90 days | 0 | 1–5 | >5 |
| Heavy processes | Non-system processes using >5% CPU or >500 MB | 0 | 1–2 | >2 |

### Dev Environment (15 pts)
[...]

PR #666, opened 2026-04-01
feat(check): add dev environment checks to optimize and check

sebastianbreguel:
## Summary

Adds a new **Dev Environment** section to `mo optimize` and `mo check` that surfaces developer-relevant system health information. These checks are purely informational — they show, don't fix.

Discussed in #659 as features that fit naturally inside existing commands.

### Launch Agents
Scans `~/Library/LaunchAgents/*.plist` for broken entries — plists whose `ProgramArguments` binary no longer exists on disk. Common after uninstalling apps.

### Dev Tools
Verifies presence of common developer tools: `git`, `node`, `python3`, `brew`, `docker`, `go`, `xcode-select`. Only reports missing tools.

### Version Mismatches
Detects conflicts between co-installed tool versions (only checked when both sides are installed):

| Pair | How detected |
|------|-------------|
| psql client vs server | `psql --version` vs `postgres --version` |
| node vs nvm default | `node --version` vs `nvm version default` |
| python3 vs pyenv | `python3 --version` vs `pyenv version` |

## Behavior

- Checks only run when relevant tools are installed
- Each check respects the existing whitelist system
- Runs in parallel in `mo check`, sequentially in `mo optimize`
- No new dependencies — uses only built-in macOS commands

## Visual preview

All healthy:
```
➤ Dev Environment
  ✓ Launch Agents All healthy
  ✓ Dev Tools      All present
  ✓ Versions       No conflicts
```

With issues:
```
➤ Dev Environment
  ◎ Launch Agents  2 broken
    com.old.removed.app, com.sketch.service
  ◎ Dev Tools      2 not found (docker, xcode-select)
  ◎ Versions       psql 16.2 vs server 14.1
```

## What changed

| File | Change |
|------|--------|
| `lib/check/dev_environment.sh` | New module: 3 check functions + aggregator |
| `bin/check.sh` | Source new module, run in parallel, display results |
| `bin/optimize.sh` | Source new module, run in `run_system_checks()` |
| `tests/dev_environment.bats` | 8 new tests with mocked system commands |

## Test plan

- [x] `bash -n lib/check/dev_environment.sh` — clean syntax
- [x] `bash -n bin/check.sh && bash -n bin/optimize.sh` — clean syntax
- [x] `bats tests/dev_environment.bats` — all 8 tests pass
- [x] `mo optimize` — Dev Environment section appears after System Configuration
- [x] `bash bin/check.sh` — Dev Environment section appears after Configuration
- [x] Checks only display when tools are installed
- [x] Whitelist keys: `check_launch_agents`, `check_dev_tools`, `check_version_mismatches`

### DIFF
--- a/bin/check.sh
+++ b/bin/check.sh
@@ -14,6 +14,7 @@ source "$SCRIPT_DIR/lib/manage/update.sh"
 source "$SCRIPT_DIR/lib/manage/autofix.sh"
 
 source "$SCRIPT_DIR/lib/check/all.sh"
+source "$SCRIPT_DIR/lib/check/dev_environment.sh"
 
 cleanup_all() {
     stop_inline_spinner 2> /dev/null || true
@@ -42,6 +43,7 @@ main() {
     local health_file=$(mktemp_file)
     local security_file=$(mktemp_file)
     local config_file=$(mktemp_file)
+    local dev_file=$(mktemp_file)
 
     # Run all checks in parallel with spinner
     if [[ -t 1 ]]; then
@@ -58,6 +60,7 @@ main() {
         check_system_health > "$health_file" 2>&1 &
         check_all_security > "$security_file" 2>&1 &
         check_all_config > "$config_file" 2>&1 &
+        check_all_dev_environment > "$dev_file" 2>&1 &
         wait
     }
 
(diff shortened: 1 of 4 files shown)


tw93:
Hey @sebastianbreguel, great feature and solid test coverage! Applied this to main directly (commit c7e71fd) with two small security fixes before merging:

1. **Removed the nvm version check** — the original implementation sourced `$NVM_DIR/nvm.sh` in a subshell, which executes arbitrary shell code and `$NVM_DIR` can be overridden by environment. Dropping it keeps the check simple and safe; `node --version` alone covers the most useful signal.

2. **Replaced `echo -e` with `printf` for plist-derived strings** — label values read from user-owned `.plist` files in `~/Library/LaunchAgents/` were being passed to `echo -e`, which interprets backslash escape sequences. Using `printf "    %s\n"` treats those strings as literals.

Everything else landed as-is, including the header deduplication in `bin/check.sh` which was a nice cleanup. Thanks for the contribution, and to get the new version just run:

```bash
mo update --nightly
```

PR #678, opened 2026-04-03
feat(optimize): add periodic maintenance trigger

sebastianbreguel:
## Summary

Adds a new optimization task that triggers macOS built-in periodic maintenance scripts (`daily`, `weekly`, `monthly`) when they haven't run in 7+ days.

These scripts rotate system logs, rebuild `locate`/`whatis` databases, and clean temp files. They run automatically via `launchd`, but can fall behind if the Mac sleeps through the scheduled window.

## Before

No way to check or trigger periodic maintenance in `mo optimize`.

## After

When scripts are stale (>7 days):
```
  ✓ Periodic maintenance triggered
```

When already current:
```
  ✓ Periodic maintenance already current (2d ago)
```

## What changed

| File | Change |
|------|--------|
| `lib/optimize/tasks.sh` | New `opt_periodic_maintenance()` — checks `/var/log/daily.out` timestamp, runs `sudo periodic daily weekly monthly` if stale |
| `lib/check/health_json.sh` | Register `periodic_maintenance` in optimization item list |
| `tests/optimize.bats` | 3 new tests: fresh log skip, stale log trigger, dispatch |

## Safety

- Checks `/var/log/daily.out` modification time before running — skips if scripts ran within 7 days
- Uses Apple's own `periodic` command — these are the same scripts macOS runs automatically
- Dry-run aware: reports what would happen without executing
- Requires sudo (same as other system-level optimizations in `mo optimize`)

## Test plan

- [x] `bats tests/optimize.bats` — 22/22 pass
- [x] Fresh log (<7 days) → reports "already current"
- [x] Stale log (>7 days) → triggers in dry-run
- [x] `execute_optimization periodic_maintenance` dispatches correctly
- [x] No regressions in existing tests

### DIFF
--- a/lib/check/health_json.sh
+++ b/lib/check/health_json.sh
@@ -143,6 +143,7 @@ EOF
     items+=('disk_permissions_repair|Permission Repair|Fix user directory permission issues|true')
     items+=('bluetooth_reset|Bluetooth Refresh|Restart Bluetooth module to fix connectivity (skips if in use)|true')
     items+=('spotlight_index_optimize|Spotlight Optimization|Rebuild index if search is slow (smart detection)|true')
+    items+=('periodic_maintenance|Periodic Maintenance|Run macOS daily/weekly/monthly maintenance scripts if stale|true')
 
     # Removed high-risk optimizations:
     # - startup_items_cleanup: Risk of deleting legitimate app helpers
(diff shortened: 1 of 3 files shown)


tw93:
Thanks @sebastianbreguel! Merged manually to main in commit 2057c26 with a few fixes:

- Made log path configurable via `MOLE_PERIODIC_LOG` (defaults to `/var/log/daily.out`) so tests exercise the real function rather than redefining it inline
- Rewrote the "fresh log" and "stale log" tests to use a temp file via `MOLE_PERIODIC_LOG`
- Added a test for the missing log file case (correctly treated as stale, triggers maintenance)
- Resolved merge conflict with #675

Issue #676, opened 2026-04-03
[FEATURE] Optimization ideas: disk health, periodic maintenance & old screenshots cleanup


sebastianbreguel:
Hey @tw93! After contributing the quarantine cleanup (#674), Brave/Stremio caches (#673), Python bytecode fix (#672), and the LaunchAgents cleanup (#675), I've been looking at what OnyX and CleanMyMac cover that Mole doesn't yet. Here are a few small additions that would fit naturally into existing commands — happy to PR any of these if you're interested.

## Status

| # | Feature | Command | PR |
|---|---------|---------|-----|
| 1 | Disk SMART Health Check | `mo check` | #677 |
| 2 | Periodic Maintenance | `mo optimize` | #678 |
| 3 | Old Screenshots Cleanup | `mo clean` | — |
| 4 | Orphaned Preferences Cleanup | `mo clean` | — |
| 5 | Xcode Derived Data Cleanup | `mo clean` | — |
| 6 | Homebrew Old Versions Cleanup | `mo clean` | — |

## 1. Disk SMART Health Check → `mo check` — PR #677

Show disk health status alongside existing battery health and memory checks. Parses `diskutil info disk0` for SMART status.

```
  ✓ Disk Health   SMART Verified
  ✗ Disk Health   SMART Failing — back up immediately
```

**Effort**: ~20 lines in `lib/check/all.sh`
**Safety**: Read-only, no system changes

## 2. macOS Periodic Maintenance Scripts → `mo optimize` — PR #678

macOS has built-in daily/weekly/monthly maintenance scripts (`/etc/periodic/`) that rotate logs, rebuild locate/whatis databases, and clean temp files. They run automatically but can fall behind if the Mac sleeps through the scheduled time.

Checks `/var/log/daily.out` timestamp, triggers only if stale (>7 days).

```
  ✓ Periodic maintenance triggered
  ✓ Periodic maintenance already current (2d ago)
```

**Effort**: ~25 lines in `lib/optimize/tasks.sh`
**Safety**: Apple's own scripts, macOS runs them automatically — this just ensures they don't fall behind

## 3. Old Screenshots Cleanup → `mo clean`

Default macOS screenshots pile up on Desktop. Cleanup for screenshots older than 30 days:

```
~/Desktop/Screenshot*.png  (>30 days)
~/Desktop/Screen Recording*.mov  (>30 days)
```

```
  ✓ Old screenshots · 14 files, 89 MB
```

[...]

PR #680, opened 2026-04-03
feat(clean): Homebrew old formula versions cleanup with --prune=30

sebastianbreguel:
## What it does

Adds Homebrew old formula versions cleanup to `mo clean` using `brew cleanup --prune=30`.

`brew cleanup` is officially recommended by Homebrew. The `--prune=30` flag targets formula versions older than 30 days, reclaiming significant disk space from outdated package versions and stale downloads.

### Before

```
  ✓ Homebrew cleanup, 3 items
```

Only runs `brew cleanup` (default prune) + `brew autoremove`. No explicit old version targeting.

### After

```
  ✓ Homebrew cleanup, 3 items
  ✓ Homebrew old formula versions · 12 removed, 1.8 GB
```

Dedicated pass with `--prune=30` specifically targeting old versions, with removed count and freed space reporting.

## Changes

- New `clean_homebrew_old_versions()` in `lib/clean/brew.sh`
- Runs `brew cleanup --prune=30` with 120s timeout via `run_with_timeout`
- Parses output for removed count and freed space
- Handles timeouts gracefully with manual fallback message
- Wired into `clean_developer_tools()` in `lib/clean/dev.sh`
- Supports dry-run mode

**Effort**: ~40 lines in `lib/clean/brew.sh`, 1 line in `lib/clean/dev.sh`
**Safety**: Only removes outdated formula versions (>30 days). Officially recommended by Homebrew.

Ref #676 item 6

## Test plan

- [x] 3 new BATS tests: normal execution with `--prune=30`, brew not installed skip, dry-run
- [x] Existing `clean_dev_caches.bats` tests pass
- [x] shellcheck clean

### DIFF
--- a/tests/clean_dev_caches.bats
+++ b/tests/clean_dev_caches.bats
@@ -457,6 +457,7 @@ clean_dev_editors() { :; }
 clean_code_editors() { :; }
 clean_dev_jetbrains_toolbox() { :; }
 clean_xcode_tools() { :; }
+clean_homebrew_old_versions() { :; }
 safe_clean() { :; }
 debug_log() { :; }
 clean_developer_tools
--- a/tests/dev_extended.bats
+++ b/tests/dev_extended.bats
@@ -323,6 +323,7 @@ clean_dev_ocaml() { :; }
 clean_xcode_tools() { :; }
 clean_code_editors() { :; }
 clean_homebrew() { :; }
+clean_homebrew_old_versions() { :; }
 clean_developer_tools
 EOF
 


tw93:
Thanks @sebastianbreguel! Instead of a separate function, I integrated `--prune=30` directly into the existing `brew cleanup` call in `clean_homebrew()`. This avoids running brew cleanup twice and naturally inherits the existing 7-day cache guard. Merged in 8a019fe.

PR #706, opened 2026-04-10
feat(analyze): add Quick Look file preview with P key

### LINKED ISSUE #704: [FEATURE] File preview
## Feature description

```
Analyze Disk  ~/Downloads/dataset  |  Total: 2.3 GB

 ▶ ○  1. ███████████████████████▒ 100.0%  |  📄 full_dataset.csv                                   2.3 GB

↑↓←→ | Space Select | Enter | R Refresh | O Open | F File | ⌫ Del | T Top 1 | Esc Back | Q/Ctrl+C Quit
```

When I'm on a file, I'd love to open the preview app, just like in finder when pressing space. Right now I can press `O` which loads the entire 2.3 GB file into Excel.

## Use case
I want to just preview files like in the Mac Finder before I delete them. 

## Proposed solution
Adding the option `P Preview` or use `Enter` to preview, as it appears that it's not doing anything when I use it on a file.

## Alternatives considered
Copy past the folder from the top and use `open <path>` to open finder and then use Space for preview

sebastianbreguel:
Closes #704

## Summary
- Press `P` in `mo analyze` to open macOS Quick Look preview on the selected file
- Works in both directory view and large files view
- Skips directories (only previews files)
- Stdout/stderr suppressed to avoid corrupting the BubbleTea TUI
- Follows existing patterns from `O` (Open) and `F` (Finder reveal)
- Uses `open` (default macOS app) instead of `qlmanage` — no crashes, no `[DEBUG]` prefix

## Changes
- `main.go` — `P` key handler + `safePreview()` function using `open`
- `view.go` — `P Preview` added to all help bar variants

## Test plan
- [x] `go build ./cmd/analyze/` — builds clean
- [x] `go test ./cmd/analyze/` — 175 tests pass
- [x] Run `mo analyze`, navigate to a file, press `P` — opens with default app
- [x] Press `P` on a directory — no action (correct)
- [x] Press `P` in large files view — opens file
- [x] Verify TUI redraws correctly after closing preview

### DIFF
--- a/cmd/analyze/main.go
+++ b/cmd/analyze/main.go
@@ -839,6 +839,25 @@ func (m model) updateKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 				m.status = fmt.Sprintf("Showing %s in Finder...", selected.Name)
 			}
 		}
+	case "p", "P":
+		// Quick Look preview (single file only, no multi-select).
+		if m.showLargeFiles {
+			if len(m.largeFiles) > 0 {
+				selected := m.largeFiles[m.largeSelected]
+				go func(path string) {
+					_ = safePreview(path)
+				}(selected.Path)
+				m.status = fmt.Sprintf("Previewing %s...", selected.Name)
+			}
+		} else if len(m.entries) > 0 {
+			selected := m.entries[m.selected]
+			if !selected.IsDir {
+				go func(path string) {
+					_ = safePreview(path)
+				}(selected.Path)
+				m.status = fmt.Sprintf("Previewing %s...", selected.Name)
+			}
+		}
 	case " ":
 		// Toggle multi-select (paths as keys).
 		if m.showLargeFiles {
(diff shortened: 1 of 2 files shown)


PremPrakashCodes:
Hey @sebastianbreguel, It working fine for most of the files, but when I try previewing `.mp4 `video files, the app crashes with the same error:

image

So I’m not sure if this is something specific to my system or a broader issue with video previews. Could you please verify if this works correctly on your side for video files as well?

sebastianbreguel:
Thanks for the report @PremPrakashCodes! This is a real bug — `qlmanage` can crash on certain file types (especially video), and because it shared a process group with Mole, the crash signal propagated and killed the parent process too.

I just pushed a fix in 23b1050 that isolates `qlmanage` into its own process group via `Setpgid`, so even if it crashes, Mole stays alive.

Could you pull the latest and verify it works on your end?

PremPrakashCodes:
@sebastianbreguel i am getting this 

image

sebastianbreguel:
Thanks for the screenshot @PremPrakashCodes 👍

This confirms qlmanage is not reliable for media files. Rather than maintaining a hardcoded extension blocklist that will grow over time, I'd suggest a simpler approach: catch the qlmanage exit code and fall back to `open -R` automatically when it crashes. That way we don't need to predict which file types will fail.

Something like: try qlmanage first, if it exits non-zero, fall back to reveal in Finder. One code path, no extension list to maintain.

PremPrakashCodes:
Thanks, @sebastianbreguel,

`open -R` successfully opens video file in Finder, but it also triggers an error window from `qlmanage`.

image

sebastianbreguel:
@PremPrakashCodes pushed a fix — now if `qlmanage` crashes (like on .mp4 files), it automatically falls back to revealing the file in Finder instead. Could you pull the latest and test if the crash dialog is gone? 🙏

sebastianbreguel:
@PremPrakashCodes pushed another fix — now media files (video/audio) skip `qlmanage` entirely and go straight to Finder reveal. The crash dialog should no longer appear. Could you test again?

PremPrakashCodes:
Congratulations 🥳 @sebastianbreguel, the error is gone!. Is it possible to remove `[DEBUG]` from the file title?

sebastianbreguel:
@PremPrakashCodes Good catch on the `[DEBUG]` title! Dropped `qlmanage` entirely — now using `open` which delegates to the default macOS app. No crashes, no debug prefix. Could you test again?

PR #711, opened 2026-04-12
feat(clean): add confirmation prompt before cleanup

### LINKED ISSUE #645: [FEATURE]  Add single final confirmation prompt for `mole clean` command
## Feature description

When executing the `mole clean` command, add a scan for junk files first, then display a **yes/no confirmation prompt** to let users confirm before performing the actual deletion operation. I mean only a final confirmation for user, not a step-by-step  confirmation.

## Use case

1. Prevent accidental execution of `mole clean` from deleting important cached/garble files irreversibly
2. Let users clearly view the scanned garbage files/size before making a deletion decision
3. Improve the safety and user experience of the cleanup command, avoiding data loss caused by misoperation

## Proposed solution

1. When the user runs `mole clean`, the program first scans all garbage files that need to be cleaned up
2. Display the scan results (e.g., number of files, total size, file paths) in the terminal
3. Pop up a confirmation prompt: `Are you sure you want to delete these files? [y/N]`
4. Only execute the deletion when the user enters `y`/`Y` and confirms; exit directly when entering `n`/`N` or pressing Enter

## Alternatives considered

1. Using a mandatory parameter like `--force` to force delete without confirmation: This increases the command complexity and is not as intuitive as a direct prompt
2. Keeping the current direct deletion logic: It has a high risk of accidental deletion and does not meet user security needs
3. Adding a dry-run mode (`--dry-run`) to preview only: Cannot replace the real-time confirmation during actual cleanup

## Additional context

This is a common security design for CLI cleanup tools (e.g., `rm -i`, `npm clean`), which can effectively avoid accidental operations and make the Mole tool more user-friendly and reliable.

sebastianbreguel:
## Summary
- Adds a final confirmation prompt (`Enter` to proceed, any other key cancels) before `mo clean` executes
- Only shown in interactive, non-dry-run mode (skipped for `--dry-run`, piped input, and external volumes)
- Uses existing `read_key` helper, matching the sudo prompt UX pattern

## Motivation
Closes #645 — users requested a single confirmation gate before cleanup runs, as a safety net against accidental execution.

## Test plan
- [ ] `mo clean` shows confirmation prompt before cleanup
- [ ] Any key except `Enter` at prompt cancels without cleaning
- [ ] `Enter` at prompt proceeds normally
- [x] `mo clean --dry-run` skips confirmation (no prompt shown)
- [x] Non-interactive mode (piped) skips confirmation
- [x] All existing tests pass (`bats tests/clean_core.bats`)

### DIFF
--- a/bin/clean.sh
+++ b/bin/clean.sh
@@ -1217,7 +1217,7 @@ main() {
         echo -ne "${PURPLE}${ICON_ARROW}${NC} Proceed with cleanup? ${GREEN}Enter${NC} continue, ${GRAY}Esc${NC} cancel: "
         local confirm
         confirm=$(read_key)
-        if [[ "$confirm" == "QUIT" ]]; then
+        if [[ "$confirm" != "ENTER" ]]; then
             echo -e " ${GRAY}Canceled${NC}"
             exit 0
         fi
--- a/bin/clean.sh
+++ b/bin/clean.sh
@@ -1211,6 +1211,19 @@ main() {
     done
 
     start_cleanup
+
+    # Final confirmation before cleanup (interactive, non-dry-run only).
+    if [[ "$DRY_RUN" != "true" && -t 0 && -z "$EXTERNAL_VOLUME_TARGET" ]]; then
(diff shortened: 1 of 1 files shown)


sebastianbreguel:
Closing — decided against adding a confirmation prompt to the clean flow.

PR #720, opened 2026-04-13
fix(optimize): remove disk verify that can freeze system on damaged volumes

### LINKED ISSUE #717: [BUG] `mole optimize` freezes system (requires hard reset) when APFS volume has inconsistencies
## Describe the bug

`mole optimize` froze my Mac and required a hard power-button reset. The freeze happened at the start of the optimize run, before any task completed.

After recovery, I checked the kernel log and found that my Data volume (`disk3s5`) has pre-existing APFS sparse-extent inconsistencies. Optimize cannot survive contact with them. The run hits the bad extents (likely via `diskutil verifyVolume /` at `lib/optimize/tasks.sh:1049`, or via earlier cache-clearing writes that touch affected inodes), the kernel-side I/O work stalls indefinitely, and the system becomes unresponsive.

The 30s `run_with_timeout` wrapper around `diskutil verifyVolume /` does not rescue the session, for two reasons:

1. `coreutils` is not installed on this system, so mole falls back to its perl/shell timeout implementation.
2. Even with `gtimeout` available, sending SIGTERM/SIGKILL to `diskutil` only kills the client. The actual verification runs in `diskmanagementd` and the kernel, which ignore signals to the client. The 30s timeout is cosmetic for this command.

## Steps to reproduce

I can't give a clean reproducer without intentionally damaging an APFS volume, but the conditions are:

1. macOS 15.6.1 (Darwin 24.6.0) on Apple Silicon, APFS boot container with a Data volume showing sparse-extent issues.
2. Run `mole optimize`.
3. The run hangs the system early on. Hard reset required.

## Evidence

Operations log (`~/Library/Logs/mole/operations.log`). The optimize session started but logged zero completed operations:

```
# ========== optimize session started at 2026-04-12 14:29:34 ==========
```

(end of file, no further entries)

A prior optimize run earlier the same day (14:14:50) successfully logged 5 operations, so the logging path works when tasks complete. The 14:29:34 run died on its first task.

Kernel log during the freeze window, via `log show --predicate 'process == "kernel"'`. APFS errors began about five seconds after the optimize session started:

```
14:29:39 kernel: (apfs) set_ino_purgeable_state:3119: disk3s5 did not find purgeable record for ino 236292328 (parent id 260956)
14:29:43 kernel: (apfs) apfs_trim_to_prev_fsize:180: disk3s5 *** reset ino 250661 size back to 432632 (from 609792)
14:29:45 kernel: (apfs) apfs_trim_to_prev_fsize:180: disk3s5 *** reset ino 31419849 size back to 3 (from 32768)
14:29:46 kernel: (apfs) apfs_find_gaps_middle_out:20693: disk3s5 Unexpected allocated/unwritten range for 8192 bytes at offset 163840 in inode 236301807 inside the sparse extent of 16384 bytes at offset 159744
14:29:46 kernel: (apfs) fs_map_file_offset_ext:21533: disk3s5 Cannot construct reservation list for sparse extent for 16384 bytes at offset 159744 in inode 236301807: Invalid argument(22)
14:29:46 kernel: (apfs) apfs_find_gaps_middle_out:20693: disk3s5 Unexpected allocated/unwritten range ... inode 236301859 ...
14:29:46 kernel: (apfs) fs_map_file_offset_ext:21533: disk3s5 ... inode 236301859: Invalid argument(22)
```

These are real APFS structural errors on sparse-file extents.

Hard reset confirmed by `/Library/Logs/DiagnosticReports/ResetCounter-2026-04-12-143953.diag`:

```
os_version: macOS 15.6.1 (24G90)
timestamp: 2026-04-12 14:39:53 -0400
Boot faults: btn_rst,finger_reset force_off
```

`btn_rst` and `force_off` indicate a power-button hard reset, about ten minutes after the optimize session started.

## Expected behavior

A maintenance tool should fail gracefully against a damaged disk rather than freeze the system. Optimize should detect a damaged volume and either refuse to run or skip filesystem-touching steps. The 30s timeout around `diskutil verifyVolume /` should actually bound the runtime of the step, which currently it does not, since the work continues in `diskmanagementd` and the kernel after the client is killed.

## Environment

```
Mole version 1.34.0
macOS: 15.6.1 (24G90)
Architecture: arm64
Kernel: Darwin 24.6.0
Boot device: /dev/disk3s1s1 (APFS, sealed snapshot of Macintosh HD)
Affected volume: /dev/disk3s5 (Data, has pre-existing sparse-extent inconsistencies)
Shell: /bin/zsh
coreutils installed: no (gtimeout/timeout both missing; mole uses perl/shell timeout fallback)
Install method: Homebrew
```

---

> [!NOTE]  
> I had claude write this report, so if there's anything else you need for this for this report to be useful, please let me know!

sebastianbreguel:
Fixes #717

## Summary
- Remove `opt_disk_verify()` which runs `diskutil verifyVolume /` during optimize
- On volumes with APFS inconsistencies, this triggers kernel-level I/O that cannot be stopped by killing the client process, freezing the entire system and requiring a hard reset
- The 30s timeout is cosmetic — it kills the `diskutil` client but `diskmanagementd` and kernel work continue indefinitely
- Filesystem verification is outside Mole's scope as a maintenance/cleaning tool

## Test plan
- [x] `bats tests/optimize.bats` — 34/34 pass
- [x] No remaining references to `disk_verify` or `opt_disk_verify` in codebase
- [x] Removed from health_json.sh task list

### DIFF
--- a/lib/check/health_json.sh
+++ b/lib/check/health_json.sh
@@ -145,7 +145,6 @@ EOF
     items+=('spotlight_index_optimize|Spotlight Optimization|Rebuild index if search is slow (smart detection)|true')
     items+=('periodic_maintenance|Periodic Maintenance|Run macOS daily/weekly/monthly maintenance scripts if stale|true')
     items+=('shared_file_list_repair|Shared File Lists|Repair corrupted Finder favorites and recent documents|true')
-    items+=('disk_verify|Disk Health|Verify filesystem integrity|true')
     items+=('login_items_audit|Login Items|Audit login items for broken entries|true')
 
     # System database cleanup (auto-run, low risk)
--- a/lib/optimize/tasks.sh
+++ b/lib/optimize/tasks.sh
@@ -1035,30 +1035,6 @@ opt_notification_cleanup() {
     fi
 }
 
-# Verify filesystem integrity via diskutil.
-opt_disk_verify() {
-    if [[ "${MOLE_DRY_RUN:-0}" == "1" ]]; then
-        opt_msg "Disk verify · skipped in dry-run"
-        return 0
-    fi
-
-    if [[ -t 1 ]]; then
-        MOLE_SPINNER_PREFIX="  " start_inline_spinner "Verifying disk filesystem..."
-    fi
-    local output
-    output=$(run_with_timeout 30 diskutil verifyVolume / 2>&1 || true)
-    if [[ -t 1 ]]; then
-        stop_inline_spinner
-    fi
-
-    if echo "$output" | grep -qi "appears to be OK\|volume appears to be ok"; then
-        opt_msg "Disk filesystem verified OK"
-    elif echo "$output" | grep -qi "error\|corrupt\|invalid"; then
-        echo -e "  ${YELLOW}${ICON_WARNING}${NC} Disk issues detected · run: sudo diskutil repairVolume /"
-    else
-        opt_msg "Disk verify complete"
-    fi
-}
 
 # Clean Knowledge/CoreDuet usage tracking databases.
 opt_coreduet_cleanup() {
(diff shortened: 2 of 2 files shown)


tw93:
@sebastianbreguel thanks for catching this and for the clear explanation. 

I took a slightly more conservative route in `3cd1b46`: keeping `opt_disk_verify` in the codebase but gating it behind `MOLE_ENABLE_DISK_VERIFY=1` so it no longer runs by default. 

Same end result for users, but leaves an opt-in for anyone who explicitly wants filesystem verification. Closing this one in favor of that commit.

PR #721, opened 2026-04-13
feat: alias `mo analyse`

bevanjkay:
## Summary

This PR introduce a simple alias for the `mo analyze` command to `mo analyse` to account for spelling the command the British-English way (as I tend to do). The alternative is to alias this command locally, but I don't think there's any harm in adding this to the package.

## Safety Review

## Tests

## Safety-related changes

- None.

### DIFF
--- a/mole
+++ b/mole
@@ -923,7 +923,7 @@ main() {
         "uninstall")
             exec "$SCRIPT_DIR/bin/uninstall.sh" "${args[@]:1}"
             ;;
-        "analyze")
+        "analyze" | "analyse")
             exec "$SCRIPT_DIR/bin/analyze.sh" "${args[@]:1}"
             ;;
         "status")


tw93:
@bevanjkay thanks for the PR. This one falls into personal-preference territory though, since `analyze` already works and a local shell alias handles the `analyse` spelling without shipping it to every user. Going to pass on adding it to the binary itself, but appreciate you taking the time to open it.

bevanjkay:
@tw93 Thanks for considering it. Unfortunately, it's actually not as simple as an `alias` because it's a subcommand. 

For anyone else who wants to implement something like this for themselves, the below snippet should work for zsh;

```zsh
if (( $+commands[mo] )); then
  mo() {
    if [[ "$1" == "analyse" ]]; then
      command mo analyze "${@:2}"
    else
      command mo "$@"
    fi
  }
fi

if (( $+commands[mole] )); then
  mole() {
    if [[ "$1" == "analyse" ]]; then
      command mole analyze "${@:2}"
    else
      command mole "$@"
    fi
  }
fi
```

tw93:
@bevanjkay really appreciate you pushing on this. 

I sat with it again and you were right, it's worth supporting more users instead of asking everyone to carry a local alias. 

Added a small mention in the README (`bcccefd`) so the spelling is discoverable, kept it light so nothing changes for existing users. 

Thanks again for the patience here.

PR #725, opened 2026-04-13
fix(clean): respect whitelist in clean_service_worker_cache()

### LINKED ISSUE #724: [BUG] Chrome MV3 extension service workers break after mo clean
## Describe the bug

After running `mo clean` while Chrome is open, Chrome extensions using Manifest V3 (MV3) immediately stop functioning mid-session. Affected extensions appear enabled in Chrome's extension manager, but their background logic stops responding. A full Chrome restart restores them.
 
There are **two issues**:
 
1. **Missing whitelist check**: `clean_service_worker_cache()` in `lib/clean/caches.sh` calls `safe_remove()` directly without calling `is_path_whitelisted()`. Extension cache directories use origin hashes (e.g. `a3f8d2...`) rather than domain names, so they never match `PROTECTED_SW_DOMAINS`. This means even when users add Chrome SW paths to `~/.config/mole/whitelist`, CacheStorage subdirectories are still deleted.
 
2. **Double-slash path mismatch in `is_path_whitelisted()`**: The Chrome profile loop in `lib/clean/user.sh` uses a trailing-slash glob (`*/;`), which produces paths like `.../Chrome/Default/`. When `/Service Worker/CacheStorage` is appended, the result contains a double slash: `.../Default//Service Worker/CacheStorage`. However, `is_path_whitelisted()` in `lib/core/app_protection.sh` does not normalize consecutive slashes in the target path, so the comparison with the whitelist entry (which has a single `/`) always fails — even when the whitelist is correctly loaded and active.

## Steps to reproduce

1. Install any Chrome MV3 extension (e.g. uBlock Origin Lite — one with `background.service_worker` in its manifest)
2. Confirm the extension is working normally with Chrome open
3. Add Chrome SW paths to `~/.config/mole/whitelist`:
   ```
   ~/Library/Application Support/Google/Chrome/Default/Service Worker/ScriptCache
   ~/Library/Application Support/Google/Chrome/Default/Service Worker/CacheStorage
   ~/Library/Application Support/Google/Chrome/Default/Service Worker/Database
   ```
4. Run `mo clean` (while Chrome remains running)
5. Observe: whitelist shows "3 custom patterns active", but Chrome Service Worker caches are still deleted
6. Observe: affected extensions immediately stop responding (no blocking, no hotkeys, etc.)
7. Note: a full Chrome restart restores normal behavior

## Expected behavior

- `clean_service_worker_cache()` should respect the whitelist, consistent with `safe_clean()` which already calls `is_path_whitelisted()`
- `is_path_whitelisted()` should normalize consecutive slashes (`//` → `/`) in the target path before comparison, so that glob-expanded paths match whitelist entries correctly

## Debug logs

Please run the command with `--debug` flag and paste the output here:

```bash
mo clean --debug
```

<details>
<summary>Debug output</summary>

```text
From mo clean output (2026-04-13):

Browsers section:
  Chrome Service Worker, 18MB
  Chrome Service Worker ScriptCache 49 items, 3.3MB

Root cause in lib/clean/caches.sh:
clean_service_worker_cache() calls safe_remove() directly without calling
is_path_whitelisted(). Extension SW cache dirs are keyed by origin hash
(not domain names), so they never match PROTECTED_SW_DOMAINS.

Relevant code:
        local is_protected=false
        for protected_domain in "${PROTECTED_SW_DOMAINS[@]}"; do
            ...
        done
        if [[ "$is_protected" == "false" ]]; then
            safe_remove "$cache_dir" true || true   # whitelist not checked
        fi
```

</details>

## Environment

```text
Mole version 1.34.0
macOS: 26.4.1
Architecture: arm64
Kernel: 25.4.0
SIP: Enabled
Install: Homebrew
Shell: /bin/zsh
```

## Additional context

The double-slash issue originates from `lib/clean/user.sh` line 1107-1109:
 
```bash
for _chrome_profile in "$HOME/Library/Application Support/Google/Chrome"/*/; do
    clean_service_worker_cache "Chrome" "$_chrome_profile/Service Worker/CacheStorage"
    safe_clean "$_chrome_profile"/Service\ Worker/ScriptCache/* "Chrome Service Worker ScriptCache"
```
 
`$_chrome_profile` ends with `/` (from glob `*/;`), and appending `/Service Worker/...` produces `//`. The whitelist validation in `bin/clean.sh` rejects patterns containing `//`, so users cannot work around this by adding `//` to their whitelist entries.
 
A fix is proposed in PR #725.

s-ja:
## Summary

`mo clean` running while Chrome is open deletes Chrome Service Worker caches, causing MV3 extensions (e.g. uBlock Origin Lite) to stop working mid-session. Adding paths to `~/.config/mole/whitelist` had no effect due to two bugs:

1. `clean_service_worker_cache()` in `lib/clean/caches.sh` calls `safe_remove()` directly without calling `is_path_whitelisted()`, so the whitelist is never consulted for CacheStorage subdirectories
2. `is_path_whitelisted()` in `lib/core/app_protection.sh` does not normalize consecutive slashes in the target path — the Chrome profile glob (`*/;`) produces trailing-slash paths, and appending `/Service Worker/...` creates `//`, which never matches the single-slash whitelist entry

Fixes #724

## Change

**`lib/clean/caches.sh`** — add whitelist check in `clean_service_worker_cache()` after the `PROTECTED_SW_DOMAINS` loop, consistent with how `safe_clean()` already handles whitelist protection:

bash

```bash
# Respect user whitelist (~/.config/mole/whitelist)
if [[ "$is_protected" == "false" ]] && declare -f is_path_whitelisted > /dev/null 2>&1 && is_path_whitelisted "$cache_dir"; then
    is_protected=true
fi
```

**`lib/core/app_protection.sh`** — normalize consecutive slashes in `is_path_whitelisted()` before path comparison:

bash

```bash
# Collapse consecutive slashes (e.g. /path//to -> /path/to)
# Glob-based paths from shell expansions like "$dir"/*/subpath produce double slashes
while [[ "$normalized_target" == *"//"* ]]; do
    normalized_target="${normalized_target//\/\///}"
done
```

## Test plan

- [ ]  Add Chrome SW paths to `~/.config/mole/whitelist`, run `mo clean` while Chrome is open → Chrome Service Worker lines absent from output (all protected, nothing deleted)
- [ ]  Chrome extensions (uBlock Origin Lite, etc.) remain functional after `mo clean`
- [ ]  Non-whitelisted Chrome caches (e.g. GPU cache) still cleaned normally
- [ ]  `mo clean` without a whitelist entry → Chrome SW caches cleaned as before (no behavior change for users who do not use the whitelist)

### DIFF
--- a/lib/clean/caches.sh
+++ b/lib/clean/caches.sh
@@ -68,6 +68,10 @@ clean_service_worker_cache() {
                 break
             fi
         done
+        # Respect user whitelist (~/.config/mole/whitelist)
+        if [[ "$is_protected" == "false" ]] && declare -f is_path_whitelisted > /dev/null 2>&1 && is_path_whitelisted "$cache_dir"; then
+            is_protected=true
+        fi
         if [[ "$is_protected" == "false" ]]; then
             if [[ "$DRY_RUN" != "true" ]]; then
                 safe_remove "$cache_dir" true || true
--- a/lib/core/app_protection.sh
+++ b/lib/core/app_protection.sh
@@ -866,8 +866,13 @@ is_path_whitelisted() {
     local target_path="$1"
     [[ -z "$target_path" ]] && return 1
 
-    # Normalize path (remove trailing slash)
+    # Normalize path (remove trailing slash and consecutive slashes)
     local normalized_target="${target_path%/}"
+    # Collapse consecutive slashes (e.g. /path//to -> /path/to)
+    # Glob-based paths from shell expansions like "$dir"/*/subpath produce double slashes
+    while [[ "$normalized_target" == *"//"* ]]; do
+        normalized_target="${normalized_target//\/\///}"
+    done
 
     # Empty whitelist means nothing is protected
     [[ ${#WHITELIST_PATTERNS[@]} -eq 0 ]] && return 1


tw93:
Thanks for the issue and the fix — the two-bug diagnosis was spot-on. Closing this in favor of #738 which handles a bash 3.2 edge case: `${var//\/\//\/}` leaves a literal backslash in the replacement on macOS's default shell, so a different approach using indirect variables is needed. That version also increments `protected_count` for correct UI accounting and includes regression tests.

PR #727, opened 2026-04-13
fix: clear queued input before uninstall selection

### LINKED ISSUE #726: [BUG] Residual Enter presses during uninstall scan can auto-select the first app
## Summary

During `mo uninstall`, the app scan phase can briefly stall while still accepting keyboard input from the terminal. If the user presses `Enter` a few times during that stall, those keystrokes are consumed by the next screen and can auto-select the first app in the uninstall list.

This makes accidental deletion much easier than it should be.

## Steps to reproduce

1. Run `mo uninstall`.
2. While the UI is stalled / transitioning through the app scan phase, press `Enter` multiple times.
3. Wait for the app selection screen to appear.

## Actual behavior

The first app can already be selected when the list appears, because the queued `Enter` presses are consumed by the selection UI.

## Expected behavior

Pending keyboard input from the scan phase should not be carried into the app selection screen. When the selection UI appears, it should require fresh user input.

## Root cause

The uninstall flow goes from app scanning / loading straight into `select_apps_for_uninstall` after `load_applications`, but does not clear pending stdin at that transition point.

Mole already has `drain_pending_input()` in `lib/core/ui.sh`, and batch uninstall confirmation already uses it. The missing guard is the boundary between:

- scan / load phase
- app selection phase

## Proposed fix

Call `drain_pending_input` immediately after `load_applications` succeeds and before `select_apps_for_uninstall` starts.

That is the smallest fix that directly addresses the accidental first-item selection.

As a separate hardening layer, the final destructive confirmation could also use a more explicit confirmation step, but the primary fix should be clearing queued input before entering the selector.

## Acceptance

1. Pressing `Enter` repeatedly during app scan must not auto-select the first app when the selection list opens.
2. The selection UI should only react to new input entered after it is shown.
3. Existing uninstall flow and batch confirmation behavior should otherwise remain unchanged.

fanbuz:
## Summary

`mo uninstall` could carry queued `Enter` presses from the scan/loading phase into the app selector. In practice that can auto-select the first app if the UI stalls for a moment and the user keeps hitting `Enter`.

This change clears pending input right after `load_applications` and before `select_apps_for_uninstall`, so the selector only reacts to fresh input.

## What changed

- call `drain_pending_input` before entering the uninstall app selector
- add a regression test around the scan -> selector transition in `main()`

Fixes #726

### DIFF
--- a/bin/uninstall.sh
+++ b/bin/uninstall.sh
@@ -1029,6 +1029,8 @@ main() {
             return 1
         fi
 
+        drain_pending_input
+
         set +e
         select_apps_for_uninstall
         local exit_code=$?
--- a/tests/uninstall.bats
+++ b/tests/uninstall.bats
@@ -307,6 +307,47 @@ EOF
 	[[ "$output" != *"more files"* ]]
 }
 
+@test "main clears pending input before app selection after scan" {
+	run env HOME="$HOME" PROJECT_ROOT="$PROJECT_ROOT" bash --noprofile --norc <<'EOF'
+set -euo pipefail
+
+trace_file="$HOME/uninstall-trace.log"
+app_cache_file="$HOME/apps-cache.txt"
+touch "$app_cache_file"
+
+log_operation_session_start() { :; }
+show_uninstall_help() { :; }
+hide_cursor() { :; }
+show_cursor() { :; }
+clear_screen() { :; }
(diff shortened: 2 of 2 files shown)


tw93:
Thanks for the quick fix! There was a conflict with #734 (which also touched `bin/uninstall.sh`), so I applied the same change directly to main in a082897. Closing here.

PR #750, opened 2026-04-15
fix(clean): accept valid special-char whitelist paths

LocNguyenSGU:
## Summary

Accept valid `mo clean` whitelist paths containing special or non-ASCII characters.

## Reproduction

Whitelist entries were being validated with an ASCII-only allowlist regex in `bin/clean.sh`, which rejected legitimate macOS paths such as:

- `~/Library/Application Support/Foo & Bar`
- `~/Library/Caches/com.example+beta`
- non-ASCII paths such as `~/Library/Caches/บริษัท`

As a result, Mole emitted `Invalid path format` and silently skipped those protection rules.

## Analysis

The parser was blocking many valid filesystem characters up front even though later whitelist matching is based on normalized path comparisons.

For this safety feature, the safer boundary is to reject only actually unsafe cases rather than maintaining a narrow filename character allowlist.

## Patch Summary

- replace the ASCII-only path regex check with a control-character guard
- continue rejecting non-absolute paths, traversal (`..`), protected system paths, and consecutive slashes
- add a regression test covering `&` and non-ASCII whitelist paths

## Risks

- Low risk and localized to whitelist parsing.
- This makes protection rules less unexpectedly restrictive without weakening the existing path-safety boundaries.

## Tests

Automated / local:

- `./scripts/check.sh`
- added regression test in `tests/manage_whitelist.bats`

Manual / focused:

- verified with a temporary HOME that valid entries containing `&` and non-ASCII characters are accepted with no `Invalid path format` warning

## Safety-related changes

Whitelist parsing now rejects control characters rather than rejecting broad classes of valid macOS filename characters. Existing safety checks for absolute paths, traversal, protected system paths, and consecutive slashes remain intact.

### DIFF
--- a/bin/clean.sh
+++ b/bin/clean.sh
@@ -73,7 +73,8 @@ if [[ -f "$HOME/.config/mole/whitelist" ]]; then
         fi
 
         if [[ "$line" != "$FINDER_METADATA_SENTINEL" ]]; then
-            if [[ ! "$line" =~ ^[a-zA-Z0-9/_.@\ *-]+$ ]]; then
+            if [[ "$line" == *$'
+'* || "$line" == *$''* || "$line" == *$'	'* ]]; then
                 WHITELIST_WARNINGS+=("Invalid path format: $line")
                 continue
             fi
--- a/tests/manage_whitelist.bats
+++ b/tests/manage_whitelist.bats
@@ -116,6 +116,35 @@ setup() {
     [ "$status" -eq 1 ]
 }
 
+
+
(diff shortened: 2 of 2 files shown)


tw93:
@LocNguyenSGU thanks for the fix and the regression test, the diagnosis is exactly right. Landed a slightly tightened version on main that uses `[[:cntrl:]]` to cover all control characters in one expression and keeps your test case. Closing this PR in favour of that commit.

Will ship in the next release, you can run `mo update --nightly` to install the latest code version.

PR #770, opened 2026-04-21
fix(optimize): skip periodic maintenance when command is unavailable (closes #767)

### LINKED ISSUE #767: [BUG] Periodic Maintenance fails on macOS 26+ — `periodic` command removed
## Description

The **Periodic Maintenance** step fails on macOS 26+ because Apple has removed `/usr/sbin/periodic` and `/etc/periodic/` entirely.

## Environment

- macOS 26.4.1 (Build 25E253)
- mole version: latest

## Output

```
➤ Periodic Maintenance
  ◎ Failed to run periodic maintenance (exit=1)
[DEBUG] periodic stderr: sudo: periodic: command not found
```

## Root Cause

`/usr/sbin/periodic` does not exist on macOS 26+. Apple appears to have removed the `periodic` command and its configuration directories (`/etc/periodic/daily`, `/etc/periodic/weekly`, `/etc/periodic/monthly`). Confirmed:

```
$ ls /usr/sbin/periodic
ls: /usr/sbin/periodic: No such file or directory

$ ls /etc/periodic/
ls: /etc/periodic/: No such file or directory

$ sw_vers
ProductName:		macOS
ProductVersion:		26.4.1
BuildVersion:		25E253
```

## Suggested Fix

Add a version check or existence check for the `periodic` binary before attempting to run it. On macOS 26+, skip the step gracefully (similar to how other steps handle missing commands) rather than reporting a failure.

```bash
if command -v periodic &>/dev/null; then
    # run periodic maintenance
else
    # skip: periodic not available on this macOS version
fi
```

Thanks for the great tool!

sebastianbreguel:
## Summary

Fixes #767. macOS 26+ removed `/usr/sbin/periodic` and `/etc/periodic/`, so `mo optimize` reports `Failed to run periodic maintenance (exit=1)` with stderr `sudo: periodic: command not found`.

Guard the `sudo periodic` call with a cheap existence check (`command -v periodic` + `-x /usr/sbin/periodic`) so affected systems skip the step cleanly, matching the existing pattern for the `sudo` availability check a few lines below.

## Change

`lib/optimize/tasks.sh`, `opt_periodic_maintenance`:

```bash
if ! command -v periodic &> /dev/null && [[ ! -x /usr/sbin/periodic ]]; then
    opt_msg "Periodic maintenance unavailable (removed in macOS 26+)"
    return 0
fi
```

Placed inside the non-dry-run branch so `MOLE_DRY_RUN=1` paths remain unchanged.

## Test plan

- [x] `shellcheck lib/optimize/tasks.sh` — clean
- [x] `bash -n lib/optimize/tasks.sh` — syntax OK on bash 3.2
- [x] `bats tests/optimize.bats` — 34/34 pass (including the 4 periodic-specific tests)
- [x] `./scripts/check.sh --no-format` — all quality gates green
- [x] Guard logic smoke-tested against a missing binary and an existing one

### DIFF
--- a/lib/optimize/tasks.sh
+++ b/lib/optimize/tasks.sh
@@ -956,6 +956,11 @@ opt_periodic_maintenance() {
     fi
 
     if [[ "${MOLE_DRY_RUN:-0}" != "1" ]]; then
+        # macOS 26+ removed /usr/sbin/periodic and /etc/periodic; skip gracefully.
+        if ! command -v periodic &> /dev/null && [[ ! -x /usr/sbin/periodic ]]; then
+            opt_msg "Periodic maintenance unavailable (removed in macOS 26+)"
+            return 0
+        fi
         if ! sudo -n true 2> /dev/null; then
             opt_msg "Periodic maintenance skipped (requires sudo)"
             return 0


sebastianbreguel:
Heads-up: the failing **Unit & Integration Tests** job is an unrelated flake — test #9 `bundle_has_installed_app finds an app by CFBundleIdentifier (Spotlight miss)` in `tests/bundle_resolver.bats` depends on Spotlight/`mdfind` state on the runner and fails intermittently. Passes cleanly locally (7/7 in that file), and this PR only touches `lib/optimize/periodic.sh`, which does not go through the bundle resolver. A rerun should be green.

tw93:
@sebastianbreguel thanks for the quick response! This was already addressed in commit 05ada61 which shipped in the latest version. You can verify with `mo update --nightly`.

PR #773, opened 2026-04-21
fix(analyze): subtract deleted size instead of rescanning (closes #764)

### LINKED ISSUE #764: [BUG] Analyze Disk rebuilds on folder delete rather than just re-computing the difference
## Describe the bug

When I use the analyze feature in mole to try to figure out what is taking up space on my machine, I'm identifying folders which should be deleted, but then mole goes and re-analyzes the whole folder! This takes time to rebuild the whole index again, when it should just compute based on what it already indexed. Even going into a folder, I find that it does a whole re-compute even when it should already have that information since it just had to compute it for the parent. This makes it practically unusable for me to actually use to find folders which I want to delete and causes a ton of reads on my disk that did not have to be incurred.

## Steps to reproduce

1. Run command: `mo ...`
2. ...
3. See error

## Expected behavior

I would expect for this to use a cache to not have to re-compute and deleting should just be a subtraction of the size that is deleted. It doesn't need to be byte-perfect, this tool is mostly for deleting that 15gb file you forgot about

## Environment

Please run `mo update` to ensure you are on the latest version, then paste the output of `mo --version` below:

```text
Mole version 1.35.0
macOS: 26.4.1
Architecture: arm64
Kernel: 25.4.0
SIP: Enabled
Disk Free: 108Gi
Install: Homebrew
Shell: /bin/zsh
```

## Additional context

Add any other context about the problem here, such as screenshots or related issues.

sebastianbreguel:
## Summary

Fixes #764 — `mo analyze` rescans the current directory every time the user deletes a folder, making the tool "practically unusable" (reporter's word) on large trees. The reporter cited `/System/Library` at ~184 GB; any click on "delete" kicked off a full rewalk before they could even see the next candidate.

## Root cause

`cmd/analyze/main.go`'s `deleteProgressMsg` handler was doing:

```go
invalidateCache(m.path)             // wipes the parent's cache
for i := range m.history { ... Dirty = true }    // all ancestors dirty
for path := range m.cache { ... Dirty = true }   // everything dirty
m.scanning = true
return m, tea.Batch(m.scanCmd(m.path), tickCmd()) // forced full rescan
```

But `removePathFromView` already subtracts the deleted entry's size from `m.totalSize` and removes it from `m.entries`. The UI model was correct the moment trash succeeded — the rescan was pure waste.

## Fix

**No more rescan on delete.** Instead:

- `subtractFromAncestors(path, size)` walks `m.history` and `m.cache`, subtracts the removed size from every entry whose path is an ancestor of the deleted path, and flags `NeedsRefresh = true`. Background refresh picks up drift; no blocking rescan.
- The top-level entry that contains the deleted path is **clamped to 0 B rather than dropped** when the subtracted size exceeds its tracked size. This survives size-accounting drift (APFS clones, symlinks, count-vs-bytes mismatches) without silently vanishing rows — the refresh corrects the number.
- The current view is persisted as the on-disk cache for `m.path` (with `NeedsRefresh=true`) **off the UI thread via a `tea.Cmd`** — `saveCacheToDiskWithOptions` gob-encodes the full `scanResult`, which on a 184 GB scan isn't free and would stall the event loop if done inline.
- `deleteProgressMsg` gains a `paths []string` field so multi-delete reports every removed path to the handler; `deletePathCmd` and `deleteMultiplePathsCmd` both populate it.

## Files

- `cmd/analyze/main.go` — rewrite delete handler + `subtractFromAncestors`, `isAncestorOf`, `saturatingSub`, `pruneEntryContaining` helpers.
- `cmd/analyze/delete.go` — populate `paths` in both command constructors.
- `cmd/analyze/incremental_delete_test.go` — 5 new tests.

## Test plan

- [x] `go test ./cmd/analyze/ -count=1 -short` → 180/180 pass (5 new: single/multi delete don't rescan, ancestor propagation updates history+cache, `isAncestorOf` edge cases including `/` root and `/foo` vs `/foobar`, error path doesn't mutate state).
- [x] `go build ./cmd/analyze/` + `go vet ./cmd/analyze/` clean.
- [x] Manual verification on `$HOME/Downloads` (~12 GB tree): deleting a subfolder updates the view immediately, no spinner, no progress bar, disk I/O from `mo` drops to idle after trash. Back-navigating to parent shows the reduced total with a silent refresh in the background.

## Known follow-up (separate PR)

Deletions from the **large-files view** don't round-trip through `m.entries`, so `subtractFromAncestors` is a no-op for them today. The `Entries` list is still updated (the file is removed from the view) and on-disk cache for the deleted path is invalidated, so the UI is consistent — but ancestor totals in `m.history` won't decrement until a refresh. Happy to tackle that in a follow-up PR once this lands and we've agreed on the shape.

### DIFF
--- a/cmd/analyze/delete.go
+++ b/cmd/analyze/delete.go
@@ -22,12 +22,16 @@ const trashTimeout = 30 * time.Second
 func deletePathCmd(path string, counter *int64) tea.Cmd {
 	return func() tea.Msg {
 		count, err := trashPathWithProgress(path, counter)
-		return deleteProgressMsg{
+		msg := deleteProgressMsg{
 			done:  true,
 			err:   err,
 			count: count,
 			path:  path,
 		}
+		if err == nil {
+			msg.paths = []string{path}
+		}
+		return msg
 	}
 }
 
@@ -43,6 +47,7 @@ func deleteMultiplePathsCmd(paths []string, counter *int64) tea.Cmd {
 			return strings.Count(pathsToDelete[i], string(filepath.Separator)) > strings.Count(pathsToDelete[j], string(filepath.Separator))
 		})
 
+		var removed []string
 		for _, path := range pathsToDelete {
 			count, err := trashPathWithProgress(path, counter)
 			totalCount += count
@@ -51,7 +56,9 @@ func deleteMultiplePathsCmd(paths []string, counter *int64) tea.Cmd {
 					continue
 				}
 				errors = append(errors, err.Error())
+				continue
 			}
+			removed = append(removed, path)
 		}
 
 		var resultErr error
@@ -64,6 +71,7 @@ func deleteMultiplePathsCmd(paths []string, counter *int64) tea.Cmd {
 			err:   resultErr,
 			count: totalCount,
 			path:  "",
+			paths: removed,
 		}
 	}
 }
--- a/cmd/analyze/incremental_delete_test.go
+++ b/cmd/analyze/incremental_delete_test.go
@@ -1,0 +1,240 @@
+//go:build darwin
+
+package main
+
+import (
+	"sync/atomic"
+	"testing"
+
+	tea "github.com/charmbracelet/bubbletea"
+)
+
+func TestDeleteProgressMsgDoesNotTriggerRescan(t *testing.T) {
+	childPath := "/tmp/mole-incremental-delete/child"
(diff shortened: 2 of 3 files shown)


tw93:
@sebastianbreguel thanks for the detailed fix and tests! This was already addressed in commit 4b40bd0 which shipped in the latest version. You can verify with `mo update --nightly`.

PR #787, opened 2026-04-24
fix: align APFS disk stats with df

### LINKED ISSUE #784: [BUG]
## Describe the bug
After cleaning with Mole, there’s a discrepancy between the free disk space reported by Finder and the output of df in the terminal.

df is assumed to be accurate. What could explain this difference?
I’ve noticed that after updating macOS, the discrepancy was resolved, but it reappears after running Mole again.

Image

Image

## Expected behavior
Consistency between df and Finder in reported free disk space.

## Environment

Mole version 1.35.0
macOS: 26.4.1
Architecture: arm64
Kernel: 25.4.0
SIP: Enabled
Disk Free: 68Gi
Install: Homebrew
Shell: /bin/zsh

yetval:
Use raw filesystem usage for APFS stats so Mole matches df, and explain Finder differences when local snapshots exist.

Fixes #784.

Tests: ./scripts/check.sh, go test ./..., bats tests/clean_system_maintenance.bats, bats tests/cli.bats

Note: ./scripts/test.sh still has unrelated failures in tests/clean_browser_versions.bats.

### DIFF
--- a/bin/clean.sh
+++ b/bin/clean.sh
@@ -1167,6 +1167,12 @@ perform_cleanup() {
         summary_details+=("Free space now: $(get_free_space)")
     fi
 
+    if [[ "$DRY_RUN" != "true" && -z "$EXTERNAL_VOLUME_TARGET" ]]; then
+        local snapshot_space_note
+        snapshot_space_note=$(get_apfs_snapshot_space_note || true)
+        [[ -n "$snapshot_space_note" ]] && summary_details+=("$snapshot_space_note")
+    fi
+
     if [[ $had_errexit -eq 1 ]]; then
         set -e
     fi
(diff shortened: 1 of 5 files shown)


tw93:
@yetval thanks for the careful work. Mole intentionally uses Finder's accounting in `mo status` so the displayed free space matches what users see in Finder's "About This Mac". Switching to raw `df`/statfs would change that user-facing contract.

The snapshot note is a nice idea but only fires for snapshots, not for purgeable cache (the more common cause of the gap), so it doesn't fully replace the Finder integration. I'd rather keep the current behavior and explain the gap on #784.

Closing this PR, appreciate the patch and the test coverage.

PR #800, opened 2026-04-26
fix: portable sed -i for GNU/Linux compatibility

### LINKED ISSUE #796: [BUG] Failing installation via script during `sed` substitution
## Describe the bug

Error during installation. 
Offending line at https://github.com/tw93/Mole/blob/main/install.sh#L682

## Steps to reproduce

1. Install with: `curl -fsSL https://raw.githubusercontent.com/tw93/mole/main/install.sh | bash`
2. See error
`sed: can't read s|SCRIPT_DIR=.*|SCRIPT_DIR="/Users/<my-user>/.config/mole"|: No such file or directory`

## Environment

```text
mo version
/usr/local/bin/mole: line 10: /usr/local/bin/lib/core/common.sh: No such file or directory
```

xronocode:
## Summary

Fixes #796

`sed -i ''` is BSD/macOS syntax. GNU sed (common on Linux) interprets `''` as the backup extension argument, then treats the substitution pattern as a filename, producing:

```
sed: can't read s|SCRIPT_DIR=.*|SCRIPT_DIR="/Users/<user>/.config/mole"|: No such file or directory
```

This causes `curl | bash` installation to fail on Linux.

## Changes

- Detect GNU vs BSD sed at runtime via `sed --version`
- Use `sed -i` on GNU sed, `sed -i ''` on BSD sed
- Only one call site affected: `install.sh:682` (SCRIPT_DIR rewrite)

## Test plan

- [x] `bash -n install.sh` passes (syntax)
- [ ] Manual: run install script on macOS — SCRIPT_DIR rewritten correctly
- [ ] Manual: run install script on Linux (GNU sed) — no error

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -679,7 +679,11 @@ install_files() {
     fi
 
     if [[ "$source_dir_abs" != "$install_dir_abs" ]]; then
-        maybe_sudo sed -i '' "s|SCRIPT_DIR=.*|SCRIPT_DIR=\"$CONFIG_DIR\"|" "$INSTALL_DIR/mole"
+        local sed_inplace=(-i '')
+        if sed --version 2>/dev/null | grep -q GNU; then
+            sed_inplace=(-i)
+        fi
+        maybe_sudo sed "${sed_inplace[@]}" "s|SCRIPT_DIR=.*|SCRIPT_DIR=\"$CONFIG_DIR\"|" "$INSTALL_DIR/mole"
     fi
 
     if ! download_binary "analyze"; then


tw93:
@xronocode follow-up: my original close reason was wrong. Looking at #796 again, the reporter is on macOS, not Linux -- their PATH had GNU sed (from Homebrew gnu-sed) shadowing BSD sed, which produces the exact same error. So this is a real bug for some macOS users, not a Linux-only concern. Thanks for surfacing it.

I went a different route to fix it: rather than detect GNU vs BSD at runtime, the two call sites now invoke `/usr/bin/sed` directly. On macOS that path is always BSD sed and cannot be PATH-shadowed, which keeps the patch tiny and preserves Mole's macOS-only assumption. Sorry for the noise on the close reason -- appreciate you taking the time to send the patch.

PR #809, opened 2026-04-28
feat(ui): add vim G and gg keybindings for jump-to-top/bottom

FMVPJet:
## Summary

- Add `G` keybinding to jump to the last item in paginated menus
- Add `gg` keybinding to jump to the first item in paginated menus
- Works in both normal and filter modes (falls back to filter input when search is active)

Complements the existing `j`/`k` navigation, making long lists (like `mo uninstall` with many apps) much faster to navigate for vim users.

## Changes

- `lib/core/ui.sh`: `read_key()` now emits `BOTTOM` for `G` and `TOP` for `gg` (with 0.5s timeout for the second `g`)
- `lib/ui/menu_paginated.sh`: Handle `TOP`/`BOTTOM` events and `CHAR:G`/`CHAR:g` in filter mode

## Test plan

- [x] `shellcheck` passes on both changed files
- [x] `shfmt` formatting applied
- [x] Existing bats tests pass (uninstall, whitelist, cli)
- [ ] Manual: `mo uninstall` → press `G` → cursor jumps to last app
- [ ] Manual: `mo uninstall` → press `gg` → cursor jumps to first app
- [ ] Manual: In search mode (`/`), typing `g` or `G` enters filter text instead of jumping

### DIFF
--- a/lib/core/ui.sh
+++ b/lib/core/ui.sh
@@ -168,47 +168,47 @@ read_key() {
             return 0
         }
         case "$key" in
-            $'\n' | $'\r') echo "ENTER" ;;
-            $'\x7f' | $'\x08') echo "DELETE" ;;
-            $'\x15') echo "CLEAR_LINE" ;; # Ctrl+U (often mapped from Cmd+Delete in terminals)
-            $'\x1b')
-                if IFS= read -r -s -n 1 -t 1 rest 2> /dev/null; then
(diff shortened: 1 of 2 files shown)


tw93:
@FMVPJet thanks for this, vim-style G/gg is a nice idea. I've re-implemented it with a smaller diff that avoids the reformatting changes. The feature will ship in the next release.

PR #811, opened 2026-04-28
fix(whitelist): detect menu cancel, surface "no changes saved" feedback (closes #807)

### LINKED ISSUE #807: [BUG] mo clean --whitelist not retaining choices
## Describe the bug

running `mo --whitelist` I see the following

Image

I remove several (see picture)

Image

But then run `mo --whitelist` again and the ones I deselected are back

Image

## Environment

Mole version 1.36.2
Channel: Nightly (24b94f7)
macOS: 15.7.5
Architecture: arm64
Kernel: 24.6.0
SIP: Enabled
Disk Free: 402Gi
Install: Manual
Shell: /bin/zsh

sebastianbreguel:
## Problem

`mo clean --whitelist` opens a multi-select menu. Pressing **Q** to quit silently aborted with no feedback. Reopening showed defaults re-preselected, leading users to think their deselects had not saved (#807).

## Root cause

Two layered issues:

1. The function ran under `set -euo pipefail`, which aborted at `paginated_multi_select` returning rc=1 **before** the next line could capture `$?`. The cancel-detection branch was effectively dead code.
2. Footer label `Q Exit` reads as "quit & save" — there was no signal that Q discarded the in-progress selection.

## Fix

- Capture the menu's exit code with the `|| exit_code=$?` idiom so `set -e` does not trip.
- Treat cancel as success: print `Cancelled, no changes saved` and `return 0`. Real failures (file write, etc.) still propagate.
- Rename `Q Exit` → `Q Cancel` in both menu variants (`menu_simple.sh`, `menu_paginated.sh`).
- Drop incidental shadowing of the `exit` builtin (`local exit=` → `local cancel_label=`).

Mirrors the existing cancel-message pattern in `lib/clean/project.sh:1615` (`echo -e "\${GRAY}Purge cancelled\${NC}"`). No new flags, no config knobs.

## Test

New BATS regression in `tests/manage_whitelist.bats`: pipes `q` to `mo clean --whitelist` and asserts (a) the whitelist file's sha is unchanged from before the cancel and (b) the `Cancelled` substring is present in stdout.

- 14/14 BATS whitelist suite pass (test #7 is the new regression test).
- 688/688 full BATS suite pass.

## Manual repro

| Input | rc | File state | Output |
|---|---|---|---|
| `printf '\n'` (Enter) | 0 | written | `Whitelist Updated` |
| `printf 'q'` (Cancel) | 0 | unchanged | `Cancelled, no changes saved` |

Closes #807.

### DIFF
--- a/lib/manage/whitelist.sh
+++ b/lib/manage/whitelist.sh
@@ -372,13 +372,13 @@ ${GRAY}Edit: ${display_config}${NC}"
     fi
 
     MOLE_SELECTION_RESULT=""
-    paginated_multi_select "$menu_title" "${menu_options[@]}"
+    local exit_code=0
+    paginated_multi_select "$menu_title" "${menu_options[@]}" || exit_code=$?
     unset MOLE_PRESELECTED_INDICES
-    local exit_code=$?
 
-    # Normal exit or cancel
     if [[ $exit_code -ne 0 ]]; then
-        return 1
+        echo -e "${GRAY}Cancelled, no changes saved${NC}"
+        return 0
     fi
 
     # Convert selected indices to patterns
--- a/lib/ui/menu_paginated.sh
+++ b/lib/ui/menu_paginated.sh
@@ -444,7 +444,7 @@ paginated_multi_select() {
             for ((i = 0; i < items_per_page; i++)); do
                 printf "${clear_line}\n" >&2
             done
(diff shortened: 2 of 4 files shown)


tw93:
@sebastianbreguel thanks for the fix, merged! Will ship in the next release.

PR #829, opened 2026-05-04
Expand rebuildable cache cleanup coverage

yuzeguitarist:
## Summary

- Add additional rebuildable Apple user cache cleanup targets.
- Add conservative system cache cleanup for the iconservices cache.
- Add narrowly scoped GPU cache cleanup under accessible `/private/var/folders/*/*/C` cache shards.
- Cover the new cleanup paths with targeted Bats tests.

## Safety

- Uses existing `safe_clean` and `safe_sudo_remove` helpers.
- Does not use raw `rm -rf` or `find -delete`.
- Keeps system-level scope limited to rebuildable cache directories.
- Limits GPU cache cleanup to cache shards whose contained files are older than the configured retention window.
- Uses non-sudo discovery for GPU cache shards so the scan only covers paths visible to the current process; deletion still flows through `safe_sudo_remove`.
- Leaves app bundles and protected system areas untouched.

## Verification

- `MOLE_TEST_NO_AUTH=1 bats tests/core_safe_functions.bats tests/clean_system_maintenance.bats tests/clean_user_core.bats`
- `MOLE_TEST_NO_AUTH=1 ./scripts/check.sh --no-format`
- `git diff --check`

### DIFF
--- a/lib/clean/system.sh
+++ b/lib/clean/system.sh
@@ -1,6 +1,45 @@
 #!/bin/bash
 # System-Level Cleanup Module (requires sudo).
 set -euo pipefail
+
+is_rebuildable_gpu_cache_dir() {
+    local cache_dir="$1"
+
+    # Only match current-user-accessible Darwin cache shards under C/.  Do not
+    # match T/ temp folders, generic /private/var/folders entries, or arbitrary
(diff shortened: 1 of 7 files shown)


yuzeguitarist:
@codex Review

PR #840, opened 2026-05-04
perf: optimize status, analyze, and cleanup scans

yuzeguitarist:
## Summary
- optimize mo status collection paths with cached command lookups, bounded GPU/trash scans, safer counter deltas, and heap-based top process selection
- optimize mo analyze scanning/rendering by reducing duplicate work and tightening concurrent directory size accounting
- speed up cleanup scans by avoiding redundant purge fallbacks, deduplicating scan output in one pass, and reducing per-directory process spawning for __pycache__ checks
- harden Homebrew-related update/version detection with timeouts and test fakes to avoid slow or network-backed checks during source-tree runs

## Validation
- MOLE_TEST_NO_AUTH=1 ./scripts/check.sh --no-format
- MOLE_TEST_NO_AUTH=1 ./scripts/test.sh
- MOLE_TEST_NO_AUTH=1 go test ./...
- make build
- git diff --check


yuzeguitarist:
I’ve pushed several rounds trying to stabilize the CI around the Homebrew/nightly update test, but it looks like this area may need maintainer guidance because the current behavior, mocks, and expected output are not fully aligned.

The main performance/status/analyze changes are still here, and the remaining failure appears isolated to `tests/update.bats` around rejecting `mo update --nightly` for Homebrew installs.

I’m going to stop iterating blindly here and defer to maintainers on the preferred direction:

* keep the new Homebrew detection timeout behavior and adjust the test expectation
* restore the previous user-facing output shape
* or split/revert the Homebrew update-detection part from this performance PR

Happy for the maintainers to take over or cherry-pick whatever parts are useful.

PR #854, opened 2026-05-08
feat(uninstall): add per-file deselect at confirm prompt (#852)

### LINKED ISSUE #852: [FEATURE] Select/deselect items in "mo uninstall"
## Feature description

Select/unselect file in `mo uninstall`

## Use case

I have to delete part of an app fairly regularly. For example, when it's not working as expected, I'll delete the app's preferences without touching the .app file itself..

## Proposed solution

For example, pressing the right arrow key on the keyboard while one of the items is highlighted might deselect it.

sebastianbreguel:
## Summary

Closes #852. Lets the user opt OUT of individual paths during `mo uninstall` instead of removing every preference and cache as one batch.

At the existing confirm prompt, press `R` to open a review picker. Every row is preselected; SPACE deselects. Deselect a preference plist or a cache file → that path is skipped. Deselect the `.app` bundle row → the app stays on disk and only the leftover files are cleaned ("kept .app" appears in the summary).

The picker reuses the existing `paginated_multi_select` UI used for app selection. No new flag, no new subcommand, no new config option.

```
→ Remove 2 apps, 412 MB  Enter confirm, R review, ESC cancel:
```

## Safety Review

Yes — touches the uninstall path. Walked the safety boundaries carefully:

- **Picker selection is the source of truth for what gets deleted.** `stop_launch_services` is now unload-only — it never calls `safe_remove`/`safe_sudo_remove` on plists. Plist deletion is owned by `remove_file_list`, which iterates the post-review surviving paths. If the user deselects a LaunchDaemon plist in the picker, the plist stays on disk.
- **`launchctl unload` always runs before deletion**, regardless of `keep_app`. When the user keeps the `.app` but removes a LaunchDaemon plist, the daemon is unloaded first so we never strand a registered-but-fileless daemon (zombie until reboot).
- **Sudo / brew / size aggregates rebuilt from scratch after each review pass** via `_recompute_uninstall_aggregates` — kept apps drop out of `sudo_apps` and `brew_cask_apps`, system files left after deselection still trigger sudo, and `size_display` reflects the trimmed selection.
- **Kept apps don't leak into removal side effects.** `success_items` only collects fully-removed apps; `kept_items` is separate. Dock cleanup (`remove_apps_from_dock`), Local Network warnings, and System Extension warnings only fire for non-kept apps. Brew uninstall, login-item removal, `unregister_app_bundle`, `mole_delete` of the `.app`, and `defaults delete` are all gated behind `keep_app != true`.
- Bundle ID validation in `stop_launch_services` (reverse-DNS regex) is unchanged.
- Dry-run path is unchanged.

## Tests

Automated:

- `bats tests/uninstall_review_files.bats` — 14 new tests covering `interactive_review_files` flatten/picker/rebuild/cancel, `keep_app` polarity, empty-bucket edge case, sudo + brew_cask + size recompute, `kept_items` vs `success_items` routing, Dock-gating, and the two safety invariants on `stop_launch_services` (no `safe_remove`, called outside the `keep_app != true` gate).
- `bats tests/uninstall.bats tests/uninstall_remove_file_list.bats tests/uninstall_naming_variants.bats tests/brew_uninstall.bats` — full uninstall suite, no regressions.
- `bats tests/scripts.bats tests/cli.bats tests/completion.bats` — broader sanity, no regressions.
- 139/139 green across the suites above.
- `shellcheck -S warning lib/uninstall/batch.sh` clean. `bash -n` clean.

Manual:

- Walked the destructive paths in source: confirmed `remove_file_list` only touches the post-review surviving paths, confirmed `_recompute_uninstall_aggregates` runs before the sudo gate so the prompt asks for sudo only when a kept app's surviving system files actually need it.
- I have not yet exercised the picker against a real cask uninstall on a live machine — would appreciate a maintainer's eye on the brew-cask + kept-`.app` interaction in particular, since `brew_cask_apps` correctly drops kept entries but the manual cleanup path still runs for non-kept casks as before.

## Safety-related changes

Yes:

- `stop_launch_services` is now unload-only. Previously it deleted plists itself via `safe_remove`/`safe_sudo_remove`, which would override an explicit deselect from the picker. Plist removal is now exclusively owned by `remove_file_list`, which respects the picker's surviving paths.
- `stop_launch_services` is now called regardless of `keep_app`, so daemons are unloaded before any plist deletion even when the `.app` is being preserved.
- `app_details` rows gain a 13th column, `keep_app=true|false`, default `false`. Existing call sites that read 12 columns were updated; legacy rows without the 13th column would parse `keep_app` as empty, which is treated as `false` everywhere it's read.

### DIFF
--- a/lib/uninstall/batch.sh
+++ b/lib/uninstall/batch.sh
@@ -92,7 +92,11 @@ decode_file_list() {
 }
 # Note: find_app_files() and calculate_total_size() are in lib/core/common.sh.
 
-# Stop Launch Agents/Daemons for an app.
+# Unload Launch Agents/Daemons for an app. **Unload only** -- never remove the
+# plist file. Plist deletion is owned by `remove_file_list` so the review
+# picker's selection is the single source of truth for what survives on disk.
+# Always called before file removal so we don't strand a running daemon when
+# its plist is about to be deleted.
 # Security: bundle_id is validated to be reverse-DNS format before use in find patterns
 stop_launch_services() {
     local bundle_id="$1"
@@ -116,7 +120,6 @@ stop_launch_services() {
     if [[ -d ~/Library/LaunchAgents ]]; then
         while IFS= read -r -d '' plist; do
             launchctl unload "$plist" 2> /dev/null || true
-            safe_remove "$plist" 2> /dev/null || true
         done < <(find ~/Library/LaunchAgents -maxdepth 1 -name "${bundle_id}*.plist" -print0 2> /dev/null)
     fi
 
@@ -124,13 +127,11 @@ stop_launch_services() {
         if [[ -d /Library/LaunchAgents ]]; then
             while IFS= read -r -d '' plist; do
                 sudo launchctl unload "$plist" 2> /dev/null || true
-                safe_sudo_remove "$plist" 2> /dev/null || true
             done < <(find /Library/LaunchAgents -maxdepth 1 -name "${bundle_id}*.plist" -print0 2> /dev/null)
         fi
         if [[ -d /Library/LaunchDaemons ]]; then
             while IFS= read -r -d '' plist; do
                 sudo launchctl unload "$plist" 2> /dev/null || true
-                safe_sudo_remove "$plist" 2> /dev/null || true
             done < <(find /Library/LaunchDaemons -maxdepth 1 -name "${bundle_id}*.plist" -print0 2> /dev/null)
         fi
     fi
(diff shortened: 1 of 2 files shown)


tw93:
@sebastianbreguel thank you, merged this PR. This is on main now and will be in the next nightly; update with `mo update --nightly` if you use the script install.

julienducos:
ok, great!
is there a page that lists the nightly versions, like for example https://github.com/tw93/Mole/releases/tag/nightly ?

tw93:
@sebastianbreguel thank you again for the PR, and sorry for the churn after merge.

I reverted the per-file review picker before release because I do not want `mo uninstall` to become a partial file-management flow. I did keep the safer LaunchAgent direction: launch services are unload-only, and plist/file deletion stays routed through the validated removal path.

No action needed from you; I just wanted to be transparent about what changed after merge.

PR #855, opened 2026-05-08
fix: route sudo-required uninstall paths into user Trash to prevent System Data growth

IMZihad21:
## Summary
- Fix: sudo‑required files now move into the user’s `~/.Trash` instead of root’s hidden `/var/root/.Trash`.
- Stops “System Data” bloat; users can now see, restore, and manage those files via Finder.
- Uses a new function `_mole_move_sudo_path_to_user_trash` with `sudo mv`, ownership repair, safe home resolution, and fallback to permanent delete on failure.

## Safety Review
- **Affected behaviour:** Uninstall, remove, analyze delete (only files needing `sudo`). User‑owned file handling stays the same.
- **New boundary:** Admin‑assisted move from protected dirs into the user’s Trash, with:
  - Strict home dir validation (rejects `/`, `/var/root`, etc.).
  - User‑owned `~/.Trash` at `700` permissions.
  - Filename sanitisation and conflict resolution (timestamp/PID/suffix).
  - No cross‑volume risk; always falls back to permanent delete if the trash move fails.

## Tests
- **Automated:** `bash -n lib/core/file_ops.sh` ✔; dedicated Bats test with faked `sudo`, `trash`, `osascript` verifies only `sudo mv` is used; full suite: 754 tests, 0 failures, 15 skipped.
- **Manual:** Verified Finder visibility, “Put Back” restore, correct user ownership, safe renaming of duplicates, and graceful fallback on unwritable Trash.

## Safety‑related changes
None.

### DIFF
--- a/lib/core/file_ops.sh
+++ b/lib/core/file_ops.sh
@@ -538,15 +538,14 @@ _mole_move_to_trash() {
         return 1
     fi
 
-    # Prefer the `trash` CLI (Homebrew formula) when available, it's faster
-    # and does not need Finder running. Fall back to AppleScript, which
-    # ships with macOS but prompts for auth on root-owned targets.
+    if [[ "$needs_sudo" == "true" ]]; then
+        _mole_move_sudo_path_to_user_trash "$path"
+        return $?
+    fi
+
+    # Prefer the `trash` CLI (Homebrew formula) for normal user-owned paths.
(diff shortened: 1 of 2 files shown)


tw93:
@IMZihad21 thank you, merged this PR. This is on main now and will be in the next nightly; update with `mo update --nightly` if you use the script install.

PR #870, opened 2026-05-11
feat(uninstall): interactive file selection, orphan cleanup, and Group Container discovery

guobosheng:
## Summary
- Add interactive file selection before uninstall confirmation — users can now choose which leftover files to remove
- Add orphan leftover cleanup for manually deleted apps (bundle gone but data remains)
- Improve leftover file discovery for Group Containers (including TeamID-prefixed) and diagnostic reports
- Show detecting message and improve not-found wording during uninstall scan
- Fix indentation, redirect spacing, and pipefail safety across multiple files

### Interactive file selection
- New `lib/uninstall/file_selector.sh` — multi-select TUI for choosing files before removal
- Team-prefix Group Containers deselected by default to avoid removing shared developer data
- Integrated into `batch_uninstall_applications()` flow before final confirmation

### Orphan leftover cleanup
- New `lib/uninstall/orphan_resolver.sh` — resolves bundle IDs for apps whose .app bundle was already deleted
- Detects leftovers via Container/Preferences/Group Container paths even when the app is gone
- Fallback to bundle domain prefix matching when team ID extraction is unavailable

### Group Container & diagnostic report discovery
- Team ID extraction from code signature for Group Container matching
- Bundle domain prefix fallback when app_path is unavailable (orphan case)
- Deduplication of directory entries across multiple discovery sources
- Diagnostic report paths matched by executable name with prefix boundary checking

### Style fixes
- Fix 8→12 space indentation in nested blocks
- Normalize redirect spacing to `2> /dev/null`
- Replace `&& ||` chains with `if/else` for pipefail safety
- Expand single-line compound blocks to multi-line form

## Test plan
- [x] `mo uninstall` interactive file selection flow
- [x] Orphan cleanup detection for manually deleted apps
- [x] Group Container matching with TeamID and domain prefix
- [x] Diagnostic report prefix boundary not over-matching
- [x] All existing uninstall tests pass
- [x] Code style checks pass (shfmt, shellcheck, bash syntax)


tw93 (changes requested):
@guobosheng thank you for the large uninstall pass. I cannot merge this safely as one PR because it combines interactive file selection, orphan cleanup, and Group Container discovery in the destructive uninstall path.

The main blockers are that the orphan resolver can return bundle IDs for apps that are still installed, and deselecting files in the selector does not fully constrain later `defaults delete` / ByHost cleanup. Please split this into separate PRs for the selector, Group Container discovery, and orphan cleanup, with regression tests for each behavior before we revisit it.

tw93:
@guobosheng thank you for the thoughtful uninstall work here. I am going to close this PR because the current shape combines several destructive-path changes into one review surface, so it is not something I can safely repair and merge as a maintainer patch.

The useful directions are still welcome as smaller PRs: one for interactive file selection, one for Group Container discovery, and one for orphan leftover cleanup. That will let us review and test each safety boundary properly.

PR #874, opened 2026-05-12
feat(uninstall): improve Group Container and diagnostic report discovery

guobosheng:
## Summary
Improve leftover file discovery for Group Containers and diagnostic reports during uninstall.

### Group Container discovery (3 tiers)
Previously only searched Group Containers by bundle-ID substring, missing TeamID-prefixed containers (e.g. `FN2V63AD2J.com.tencent` where the group-id is just the domain prefix, not the full bundle ID).

Now adds two more discovery tiers:
1. **Bundle ID substring** (existing) — `*com.tencent.xinWeChat*`
2. **Team ID via code signature** (new) — Extracts TeamIdentifier from app's code signature and searches `*<TeamID>*`. Catches containers where the group-id uses a shorter domain prefix.
3. **Domain prefix fallback** (new) — When app_path is unavailable, strips the last bundle ID component (e.g. `com.tencent.qq` → `com.tencent`) and searches `*<prefix>*`.

All tiers include deduplication to avoid listing the same container twice.

### Diagnostic report improvements
- Lower minimum prefix length from 3 to 2 chars (supports short-named apps like QQ)
- Case-insensitive matching via `-iname` for the orphan path where prefix comes from user input
- Support empty `app_path` parameter (orphan case) with fallback to nospace name matching

## Test plan
- [x] TeamID-prefixed Group Containers discovered via codesign
- [x] Domain prefix fallback when app_path is unavailable
- [x] Deduplication prevents duplicate entries
- [x] Diagnostic report matching with empty app_path
- [x] Case-insensitive prefix boundary not over-matching
- [x] Pre-commit checks pass (bash syntax, shfmt, shellcheck)


guobosheng:
@tw93 I sent a private security/safety report by email. Please temporarily revert this PR pending investigation.

PR #875, opened 2026-05-12
feat(uninstall): add interactive file selection before removal

guobosheng:
## Summary
Add interactive file selection before uninstall confirmation — users can now review and deselect individual files before removal. Also fix a bug where defaults delete and ByHost cleanup would run even when the user deselected those plists.

### New module: `lib/uninstall/file_selector.sh`
- Lightweight multi-select TUI for choosing which files to remove
- Team-prefix Group Containers (e.g. `FN2V63AD2J.com.foo`) are deselected by default — shared across apps from the same developer
- Integrated into `batch_uninstall_applications()` flow before final confirmation

### Bug fix: defaults delete / ByHost cleanup respects file selector
Previously, `defaults delete "$bundle_id"` and ByHost `find`+`mole_delete` ran unconditionally after successful app removal, ignoring the user's file selection. Now both operations check whether the corresponding plists were actually selected before executing.

### Integration
- `batch_uninstall_applications()`: pre-scan phase collects files → file selector filters them → execution uses filtered list
- When no files are selected for an app, it is skipped entirely
- When the app bundle is deselected but leftovers are kept, only the leftovers are removed

## Test plan
- [x] Deselecting preferences plist prevents defaults delete
- [x] Deselecting ByHost plists prevents ByHost cleanup
- [x] Selecting all files runs both defaults delete and ByHost cleanup
- [x] Pre-commit checks pass (bash syntax, shfmt, shellcheck)

### DIFF
--- a/lib/uninstall/batch.sh
+++ b/lib/uninstall/batch.sh
@@ -472,6 +472,9 @@ batch_uninstall_applications() {
     local _fs_i _fs_detail _fs_an _fs_ap _fs_bi _fs_tk _fs_ef _fs_esf _fs_hs _fs_ns _fs_ib _fs_cn _fs_eds _fs_hln
     local _fs_rel _fs_sys _fs_diag _fs_rel_kb _fs_sys_kb _fs_app_kb _fs_new_tk _fs_new_ef _fs_new_esf
 
+    sudo_apps=()
+    brew_cask_apps=()
+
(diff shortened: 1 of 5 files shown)


guobosheng:
@tw93 I sent a private security/safety report by email. Please temporarily revert this PR pending investigation.

PR #900, opened 2026-05-15
feat(i18n): add Simplified Chinese (zh-CN) localization support

dunova:
## Summary

Add a lightweight, zero-dependency internationalization (i18n) framework that enables Mole to display all user-facing text in **Simplified Chinese (简体中文)** when the system locale is `zh_CN` or when `MO_LANG=zh` is set.

## Motivation

Mole is widely used in the Chinese developer community (作者本身也是中文开发者！). Adding native Chinese language support lowers the barrier for non-English-speaking users and makes the tool more accessible.

### Shell (Bash) i18n
- **`lib/core/i18n.sh`**: Locale detection (`MO_LANG` → `LC_ALL` → `LANG`) and `msg()`/`msgf()` lookup functions with English fallback
- **`lib/i18n/en.sh`**: English string definitions (~200 keys covering commands, menus, help text, status messages)
- **`lib/i18n/zh_CN.sh`**: Professionally translated Simplified Chinese strings
- Updated `commands.sh`, `help.sh`, `common.sh` to use `msg()` lookups

### Go TUI i18n
- **`internal/locale/`**: Lightweight locale detection + message map system
- **`internal/locale/en.go`**: English strings for `mo status` and `mo analyze` TUI
- **`internal/locale/zh.go`**: Chinese strings for both TUI components
- Updated `cmd/status/` and `cmd/analyze/` view and main files

## Design Decisions

- **Zero external dependencies** — no i18n libraries needed
- **Fully backwards compatible** — English remains the default; behavior unchanged for existing users
- **Language auto-detected** from `$LANG`/`$LC_ALL`; can be overridden with `MO_LANG=zh`
- **Technical terms preserved** — macOS, Homebrew, Touch ID, sudo, Finder etc. stay in English
- **CJK width handling** — leverages existing `get_display_width()` in `ui.sh`
- **Extensible** — adding more languages only requires a new locale file

## Usage

```bash
# Auto-detect from system locale (zh_CN.UTF-8 → Chinese)
mo

# Explicit override
MO_LANG=zh mo          # 简体中文
MO_LANG=en mo          # English (default)
```

## Testing

- [x] `go build ./cmd/status/` — ✅
- [x] `go build ./cmd/analyze/` — ✅
- [x] `go vet ./internal/locale/ ./cmd/status/ ./cmd/analyze/` — ✅
- [x] Verified English output unchanged when `LANG=en_US.UTF-8`
- [x] Verified Chinese output renders correctly with `MO_LANG=zh`

## Files Changed (14 files, +1277 / -168)

| File | Change |
|------|--------|
| `internal/locale/locale.go` | NEW — Go locale framework |
| `internal/locale/en.go` | NEW — English messages (Go) |
| `internal/locale/zh.go` | NEW — Chinese messages (Go) |
| `lib/core/i18n.sh` | NEW — Shell i18n framework |
| `lib/i18n/en.sh` | NEW — English messages (Shell) |
| `lib/i18n/zh_CN.sh` | NEW — Chinese messages (Shell) |
| `lib/core/commands.sh` | MODIFIED — Use msg() for descriptions |
| `lib/core/help.sh` | MODIFIED — Use msg() for help text |
| `lib/core/common.sh` | MODIFIED — Load i18n + translate brew msgs |
| `cmd/status/view.go` | MODIFIED — Use locale.T() for all labels |
| `cmd/status/main.go` | MODIFIED — Init locale + translate loading |
| `cmd/status/metrics_health.go` | MODIFIED — Translate health labels |
| `cmd/analyze/view.go` | MODIFIED — Use locale.T() for UI text |
| `cmd/analyze/main.go` | MODIFIED — Init locale + translate status |


dunova:
Update: rewrote the i18n framework to use plain shell variables + indirect expansion (`${!varname}`) instead of associative arrays, so it now works on macOS stock `/bin/bash 3.2` (which Mole already targets across the rest of the codebase, see `tests/uninstall_scan_bash32.bats` and the existing 3.2-compat comments).

### What was wrong in the previous force-push
The original commit used `declare -gA MOLE_MSG=(...)`, which requires bash 4+. On a clean macOS that ships only `/bin/bash 3.2`, every `mo` invocation crashed with:

```
lib/i18n/en.sh: line 12: declare: -g: invalid option
lib/i18n/en.sh: line 12: TAGLINE: unbound variable
```

### Fixes in this revision
1. `lib/core/i18n.sh`: dropped `declare -gA`, look up keys via `${!MOLE_MSG_<KEY>}`. Also implemented true fallback layering — English is sourced first, the target locale overlays it, so a missing zh_CN key shows English instead of the raw key.
2. `lib/i18n/{en,zh_CN}.sh`: each entry is now a flat `MOLE_MSG_<KEY>="..."` variable.
3. `lib/core/common.sh`: moved the `source i18n.sh` line above `source help.sh`, since `help.sh` calls `msg()` at parse time. The previous order would not even have loaded.
4. `mole`: localized the brand banner tagline, `show_version`, the second half of `show_help`, and the three remaining "Unknown ... option" error paths that the original commit missed.
5. `lib/core/help.sh`: switched from `echo "$(...)"` to direct `msg` / `printf`, fixing 12 SC2005 shellcheck warnings, and added a missing `HELP_UNINSTALL_EXAMPLES` key so the "Examples:" header is also translated.

### Verification
- `/bin/bash 3.2` (default macOS): `mo --help`, `mo --version`, `mo clean --help`, `mo touchid --help`, `mo nonexistent`, `mo update --bogus`, `mo remove --bogus` — all render correctly in both `en` and `MO_LANG=zh`.
- `shellcheck lib/core/i18n.sh lib/core/help.sh lib/core/commands.sh lib/i18n/*.sh lib/core/common.sh mole` — clean.
- `go build ./cmd/status ./cmd/analyze` — clean.
- `go vet ./...` — clean.
- `bats tests/` (full suite) — **793/793 passing**, including `tests/uninstall_scan_bash32.bats`.

tw93:
Thanks for the substantial effort, and for caring about Chinese-speaking users. We're keeping the CLI English-only for now: a full i18n framework means every new user-facing string carries a bilingual maintenance cost, and the CLI changes often on nightly, which would keep this in near-constant merge conflict. Localization investment goes into the native macOS app. Closing as not planned, thank you for understanding.

PR #907, opened 2026-05-16
fix(clean): skip dotdir orphan when an enabled Claude plugin owns it (#889)

### LINKED ISSUE #889: `mo clean` flags enabled agent-plugin state as orphan dotfiles
`mo clean` reports `~/.cc-safety-net` as a potential orphan:

```text
• Potential orphan dotfile: ~/.cc-safety-net (102KB)
↳ No matching binary in PATH
```

But this directory is owned by an enabled Claude Code plugin:

- `~/.claude/plugins/installed_plugins.json` contains `safety-net@cc-marketplace`
- `~/.claude/settings.json` has `safety-net@cc-marketplace` enabled
- `~/.cc-safety-net/logs/*.jsonl` stores safety-net command records

The orphan-dotdir hint in `lib/clean/hints.sh` checks top-level `~/.<dir>` names against a static safe list, PATH binaries, and LaunchAgent refs. Agent plugins can own state directories without installing a matching binary, so active plugin state becomes a false positive.

This is similar to #872, but for agent plugins instead of GUI apps/casks. #879 added `dotdir_has_owning_gui_app` to skip dotdirs whose name token matches an installed `.app` or Homebrew cask. That handles GUI app ownership, but not Claude/plugin-owned state.

## Requested Behavior

Add a parallel ownership gate for enabled agent plugins before reporting a dotdir as orphaned.

Possible output:

```text
• Plugin-owned dotfile: ~/.cc-safety-net
↳ Owned by enabled Claude plugin safety-net@cc-marketplace
```

## Implementation Note

For Claude Code, check `~/.claude/settings.json` `enabledPlugins` plus `~/.claude/plugins/installed_plugins.json`.

Prefer explicit dotdir references in plugin manifests/scripts/settings. Plugin-name token matching could reuse the #879 style, but should stay weaker or informational so stale real orphans are not hidden just because an enabled plugin has a similar generic name.

therahul-yo:
## Summary

Fixes #889. `show_orphan_dotdir_hint_notice` was flagging active Claude Code plugin state directories (e.g. `~/.cc-safety-net` owned by `safety-net@cc-marketplace`) as "Potential orphan dotfile" because the existing ownership gates only know about installed `.app` bundles and Homebrew casks (PR #879's scope) — not agent plugins. The reporter's own example walks through this exact false-positive.

## Approach

Mirrors PR #879's GUI-app gate one-for-one:

| Existing (#879, GUI apps + casks) | New (this PR, agent plugins) |
|---|---|
| `_dotdir_owner_collect_tokens` | `_dotdir_owner_collect_agent_plugin_tokens` |
| `dotdir_has_owning_gui_app` | `dotdir_has_owning_agent_plugin` |
| Cached 5 min at `~/.cache/mole/installed_app_tokens_cache` | Cached 5 min at `~/.cache/mole/installed_agent_plugin_tokens_cache` |
| ≥4-char token threshold (avoids `.ai-old` vs `AI.app`) | Same threshold (avoids `ai@market` vs `.ai-old`) |
| Wired in via `dotdir_has_owning_gui_app "\$name"` | Wired in right after, via `dotdir_has_owning_agent_plugin "\$name"` |

## Why bash-native JSON parsing

`bin/optimize.sh` already has the comment `Bash-native JSON parsing helpers (no jq dependency)`, so adding a jq dep for one new feature would regress Mole's portability posture. The new `_dotdir_owner_collect_agent_plugin_tokens` uses a small awk script to walk `enabledPlugins` and emit the keys of `true`-valued entries.

## Only ENABLED plugins count

The reporter's hint warned: *"stale real orphans [should] not be hidden just because an enabled plugin has a similar generic name"*. So I only honor entries whose value is `true` in `enabledPlugins`. A plugin that's installed but disabled keeps its state directory eligible for the orphan flag — which is what the user probably wants when they disabled the plugin to clean it up.

## What I deliberately did NOT do

- **No installed_plugins.json parsing.** The reporter's implementation note mentioned it as a secondary signal, but it lists every install regardless of enabled status — which would over-suppress orphan detection. `enabledPlugins` is the right truth source for "is this plugin actively in use right now".
- **No plugin manifest scraping** for explicit dotdir references. The reporter mentioned this as an ideal but acknowledged token matching is the pragmatic v1 (matching PR #879). Manifest scraping can land in a follow-up.
- **No support for other agent platforms yet.** The function is named generically (`agent_plugin`, not `claude_plugin`) so a future PR for Codex/Cursor/etc. can extend the collector without an interface change.

## Test plan

5 new bats cases in `tests/clean_hints.bats`:

| Test | Asserts |
|---|---|
| skips dotdir owned by enabled Claude plugin | The exact `~/.cc-safety-net` + `safety-net@cc-marketplace` repro from #889 is no longer flagged. |
| still flags dotdir for DISABLED plugin | Same dotdir, plugin entry now `false` → still surfaces as orphan candidate (no over-suppression). |
| `dotdir_has_owning_agent_plugin` returns 0 for enabled plugin match | Helper returns success on the happy path. |
| `dotdir_has_owning_agent_plugin` returns 1 when settings.json missing | Graceful no-op when Claude isn't installed. |
| `dotdir_has_owning_agent_plugin` rejects short names | 4-char threshold honored so `.ai-old` isn't masked by an enabled `ai@market` plugin. |

```
$ bats tests/clean_hints.bats
... (snip)
ok 18 show_orphan_dotdir_hint_notice limits output to max 5 candidates
ok 19 show_orphan_dotdir_hint_notice skips dotdir owned by enabled Claude plugin (#889)
ok 20 show_orphan_dotdir_hint_notice still flags dotdir for DISABLED plugin (#889)
ok 21 dotdir_has_owning_agent_plugin returns 0 for enabled plugin match (#889)
ok 22 dotdir_has_owning_agent_plugin returns 1 when settings.json missing (#889)
ok 23 dotdir_has_owning_agent_plugin rejects short names (#889)

23 tests, 0 failures
```

## Files

- `lib/clean/hints.sh` — `+89 / -0`
- `tests/clean_hints.bats` — `+124 / -0`

### DIFF
--- a/lib/clean/hints.sh
+++ b/lib/clean/hints.sh
@@ -672,6 +672,91 @@ _dotdir_owner_collect_tokens() {
     fi
 }
 
+# Emit every alnum token found in *enabled* agent-plugin names, lowercased,
+# one per line.  Currently covers Claude Code's plugin manifest at
+# ``~/.claude/settings.json`` — only entries whose value is ``true`` count,
+# so a stale dotdir for a disabled or uninstalled plugin is still surfaced
+# as an orphan (which is what the user usually wants).
+#
+# Parsing is bash/awk native to match Mole's existing "no jq dependency"
+# stance (see ``bin/optimize.sh``).
+# shellcheck disable=SC2329
+_dotdir_owner_collect_agent_plugin_tokens() {
+    local settings="$HOME/.claude/settings.json"
+    [[ -r "$settings" ]] || return 0
+
+    # The shape inside settings.json is:
+    #   "enabledPlugins": { "<name>@<marketplace>": true|false, ... }
+    # We only emit names whose value is ``true`` and strip the
+    # @marketplace suffix so the token list mirrors the GUI-app collector.
+    awk '
+        /"enabledPlugins"[[:space:]]*:[[:space:]]*\{/ { in_block=1; depth=1; next }
+        in_block {
+            n=gsub(/\{/, "&"); depth+=n
+            n=gsub(/\}/, "&"); depth-=n
+            if (depth <= 0) { exit }
+            if (match($0, /"[^"]+"[[:space:]]*:[[:space:]]*true/)) {
+                line = substr($0, RSTART)
+                if (match(line, /"[^"]+"/)) {
+                    key = substr(line, RSTART + 1, RLENGTH - 2)
+                    sub(/@.*$/, "", key)
+                    print key
+                }
+            }
+        }
+    ' "$settings" 2> /dev/null |
+        LC_ALL=C tr '[:upper:]' '[:lower:]' |
+        LC_ALL=C tr -cs 'a-z0-9' '\n'
+}
+
+# Return 0 if any ≥4-char token from `name` matches a token harvested from
+# enabled Claude Code agent plugins.  Cached for 5 minutes alongside the
+# GUI-app token cache.  Mirrors ``dotdir_has_owning_gui_app`` so an active
+# plugin's state directory (e.g. ``~/.cc-safety-net`` for the
+# ``safety-net@cc-marketplace`` plugin) is not flagged as an orphan
+# dotfile.  See issue #889.
+# shellcheck disable=SC2329
+dotdir_has_owning_agent_plugin() {
+    local name="$1"
+    [[ -z "$name" ]] && return 1
+    [[ ${#name} -lt 4 ]] && return 1
+
+    local cache_dir="$HOME/.cache/mole"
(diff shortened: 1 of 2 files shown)


tw93:
Thanks for the thorough PR. #889 is already fixed on `main`: `hint_collect_claude_plugin_tokens` collects plugin tokens from both `enabledPlugins` in `settings.json` **and** `~/.claude/plugins/installed_plugins.json`, so a dotdir owned by any installed Claude plugin is spared. We went with the broader source deliberately, flagging the state dir of a plugin a user still has installed felt like the worse failure mode. Closing as already addressed; appreciate the detailed write-up.

PR #916, opened 2026-05-18
feat(optimize): add Spotlight deep reset

thedavidweng:
## Summary

- add explicit `mo optimize --spotlight-deep-reset` flow with typed confirmation before sudo
- limit deletion to current-user Spotlight/CoreSpotlight cache targets and keep default clean behavior unchanged
- expose the option in optimize help, shell completions, optimize whitelist, and security notes

## Test Plan

- [x] `bash -n mole bin/optimize.sh lib/optimize/tasks.sh lib/core/help.sh bin/completion.sh tests/optimize.bats tests/completion.bats`
- [x] `git diff --check`
- [x] `bats tests/optimize.bats tests/completion.bats`
- [x] `env GOCACHE=/private/tmp/mole-go-build GOPATH=/private/tmp/mole-go go test ./...`
- [x] `env GOCACHE=/private/tmp/mole-go-build GOPATH=/private/tmp/mole-go scripts/test.sh tests/optimize.bats tests/completion.bats`
- [x] manual dry-run, confirmation, symlink refusal, and simulated mdutil flow checks

Note: a full local `scripts/test.sh` run on a clean temp worktree still has the existing baseline failures in `tests/cli.bats` for `mo status --json`; the same four failures reproduce without this patch.

### DIFF
--- a/SECURITY_AUDIT.md
+++ b/SECURITY_AUDIT.md
@@ -301,6 +301,8 @@ Key coverage areas include:
 - bundle ID boundary matching and malformed-ID rejection (`tests/uninstall_safety.bats`)
 - bash 3.2 empty-array nounset compatibility (`tests/uninstall_scan_bash32.bats`)
 
+Spotlight user search caches are protected from default `mo clean` removal. The explicit `mo optimize --spotlight-deep-reset` path is separate from normal cleanup and optimization, requires typed confirmation, and is limited to current-user Spotlight/CoreSpotlight cache targets before restarting indexing for `/`.
+
 ## Known Limitations and Future Work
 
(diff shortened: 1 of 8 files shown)


tw93:
@thedavidweng Thanks for the thoughtful implementation and tests. I am going to close this one because the feature direction is outside Mole’s current safety boundary.

`mdutil -E /`, toggling Spotlight indexing, and restarting `corespotlightd` are system-level repair actions. Mole is trying to stay focused on helping users safely find and remove cleanable space after confirmation, not becoming a broader macOS repair suite. Without a concrete issue showing this flow is needed and safe for real users, I do not want to ship a deep Spotlight reset command.

Appreciate the work here; the implementation discipline is good, but the product boundary is the blocker.

PR #923, opened 2026-05-18
Fix uninstall dry-run leftover warnings

M-Hassan-Raza:
mo uninstall --dry-run was leaving files in place like it should, then checking those same files and printing Could not remove.

that made a preview look like a failed uninstall.

this skips that leftover check only in dry-run. real uninstall still reports leftovers.

tested with bats tests/uninstall.bats, ./scripts/check.sh --no-format, ./scripts/test.sh, make build


tw93:
@M-Hassan-Raza Thanks for the focused dry-run fix and regression test. Merged into main.

PR #926, opened 2026-05-18
fix(purge): restore cursor when purge exits early

### LINKED ISSUE #915: [BUG]  Mole - Purge command can't show_cursor
Mole - Purge command can't show_cursor

The cursor did not recover after running mole

YuanHanzhong:
Summary
- Restore the cursor from a cleanup trap after `mo purge` hides it.
- Preserve the purge command exit status while still running cleanup.
- Add a focused regression test for early purge exit after cursor hide.

Testing
- bash -n bin/purge.sh tests/purge.bats
- shellcheck --rcfile .shellcheckrc bin/purge.sh tests/purge.bats
- BATS_FORMATTER=tap bats tests/purge.bats -f "mo purge restores cursor when purge exits after hiding it"
- git diff --check

Notes
- The full tests/purge.bats suite still has two unrelated tilde-expansion failures on upstream main.

Fixes #915

### DIFF
--- a/bin/purge.sh
+++ b/bin/purge.sh
@@ -13,13 +13,29 @@ export LANG=C
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 source "$SCRIPT_DIR/../lib/core/common.sh"
 
-# Set up cleanup trap for temporary files
-trap cleanup_temp_files EXIT INT TERM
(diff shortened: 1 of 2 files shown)


tw93:
@YuanHanzhong Thanks for jumping on this cursor restore bug. I merged #928, which covers the same #915 issue and also keeps the existing temp-file cleanup trap intact across exit paths. Closing this one as a duplicate so we do not carry two competing implementations for the same fix.

PR #927, opened 2026-05-18
feat(clean): add dirs_cleaner cleanup command

richardgetz:
## Summary
- audit `/private/var/dirs_cleaner` during normal system cleanup and report suspicious stuck cleanup staging without deleting by default
- add explicit `mo clean --dirs-cleaner` command with `--dry-run` support for deliberate stale staging cleanup
- constrain cleanup to same-filesystem, non-symlink, non-mounted, protected/whitelist-respecting top-level or shallow staging entries with fail-closed scans

## Why this is safe
- normal `mo clean` remains report-only for dirs_cleaner findings
- cleanup never removes `/private/var/dirs_cleaner` itself
- cleanup uses Mole's existing sudo/session and `safe_sudo_remove` path, with a dirs_cleaner-specific validation context
- removal candidates must pass full recursive same-filesystem newest-mtime staleness proof; scan failures skip instead of deleting
- non-empty top-level buckets are not removed as a unit; Mole checks shallow children instead

## Testing
- `bash -n bin/clean.sh lib/clean/system.sh lib/core/file_ops.sh lib/core/help.sh tests/clean_system_maintenance.bats tests/core_safe_functions.bats tests/cli.bats`
- `git diff --check main...HEAD`
- `GOPATH=/private/tmp/mole-gopath GOMODCACHE=/private/tmp/mole-gopath/pkg/mod GOCACHE=/private/tmp/mole-gocache ./scripts/check.sh --no-format`
- `GOPATH=/private/tmp/mole-gopath GOMODCACHE=/private/tmp/mole-gopath/pkg/mod GOCACHE=/private/tmp/mole-gocache go test ./cmd/...`
- `GOPATH=/private/tmp/mole-gopath GOMODCACHE=/private/tmp/mole-gopath/pkg/mod GOCACHE=/private/tmp/mole-gocache MOLE_GO_TEST_CACHE=/private/tmp/mole-go-build-cache ./scripts/test.sh`

Note: `bats`, `shellcheck`, and `shfmt` were not installed in this environment; the repo scripts skipped/fell back as designed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,8 @@ Space freed: 95.5GB | Free space now: 223.5GB
 
 Note: In `mo clean` -> Developer tools, Mole removes unused CoreSimulator `Volumes/Cryptex` entries and skips `IN_USE` items.
 
+During system cleanup, Mole also audits `/private/var/dirs_cleaner` for unusually large or stale macOS cleanup staging entries. The default check is report-only: it shows size, age, ownership metadata, and review commands. To explicitly clean stale top-level or shallow staging entries, preview with `mo clean --dirs-cleaner --dry-run`, then run `mo clean --dirs-cleaner`.
+
 ### Smart App Uninstaller
 
 ```bash
--- a/SECURITY_AUDIT.md
+++ b/SECURITY_AUDIT.md
@@ -79,6 +79,7 @@ Some subpaths under otherwise protected roots are explicitly allowlisted for bou
 - `/private/var/db/DiagnosticPipeline`
 - `/private/var/db/powerlog`
 - `/private/var/db/reportmemoryexception`
+- `/private/var/dirs_cleaner/<top-level-or-shallow-child>` (only via explicit stale staging cleanup; never the parent)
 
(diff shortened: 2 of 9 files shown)


tw93:
@richardgetz Thanks for the very careful work and the detailed safety thinking. I am going to close this PR in its current shape because the explicit cleanup path is too risky for Mole’s current boundary.

The report-only audit idea for `/private/var/dirs_cleaner` is worth discussing, but I do not want to accept a `mo clean --dirs-cleaner` command that performs sudo deletion under `/private/var/dirs_cleaner`, or broaden `safe_sudo_remove` around that cleanup context. I also reproduced local test failures in the new dirs_cleaner safety coverage, so this is not ready as-is.

If you want to continue this, the mergeable shape would be a much smaller PR that only adds a read-only audit: no deletion command, no `safe_sudo_remove` context changes, and focused tests for the warning output.

richardgetz:
It started as read-only, but I added the explicit cleanup for others assuming that would be the end goal for anyone else facing stuck "to be deleted" files. Is the expectation to never clean out system files that are meant for deletion? Tbh not worth my time if the tool isn't managing the deletion. Best of luck!

PR #950, opened 2026-05-23
Harden deletion path validation

M-Hassan-Raza:
Tightens deletion path validation so aliased critical paths and protected symlink targets are rejected before cleanup runs.

The root cause was raw path comparison plus mole_delete skipping symlink validation. This normalizes deletion policy checks, validates symlink targets, and adds coverage for fuzz corpus paths, generated control characters, safe user cache paths, and file-ops-only safe_find_delete.

Validated locally with ./scripts/check.sh --format, ./scripts/check.sh --no-format, MOLE_TEST_NO_AUTH=1 bats tests/path_validation_fuzz.bats tests/core_safe_functions.bats tests/file_ops_mole_delete.bats, and MOLE_TEST_NO_AUTH=1 MOLE_TEST_JOBS=2 BATS_FORMATTER=tap ./scripts/test.sh.

### DIFF
--- a/lib/core/file_ops.sh
+++ b/lib/core/file_ops.sh
@@ -63,7 +63,7 @@ format_duration_human() {
 # Path Validation
 # ============================================================================
 
-_mole_deletion_policy_path() {
+_mole_normalize_deletion_policy_path() {
     local path="$1"
     local slash="/"
     local double_slash="//"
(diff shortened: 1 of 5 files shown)


tw93:
@M-Hassan-Raza thanks for the contribution. This has been merged into main and will go out with the next release.

PR #957, opened 2026-05-23
Fix stable channel update switch

M-Hassan-Raza:
Plain mo update now switches nightly or dev installs back to the stable tagged installer even when the current version already matches the latest release. The update test covers that same-version channel switch path.


tw93:
@M-Hassan-Raza thanks for the fix. This has been merged into main and will go out with the next release.

PR #967, opened 2026-05-24
Add mo history for recent cleanup activity

M-Hassan-Raza:
Before this, checking what Mole did recently meant reading raw operations and deletion logs directly. That made simple use cases awkward, like a small dashboard, a scheduled automation report, or just answering what the last cleanup removed, skipped, or failed on.

This adds a read-only mo history command that summarizes recent operation sessions and deletion audit entries. It also supports --json for scripts and --limit for bounded output.

The command is wired into help, completions, and docs. It is intentionally side-effect free, so asking for history does not create normal Mole log files. The parser is conservative around malformed log lines, unfinished sessions, bad limits, and odd path characters in JSON output.

Tested with:
MOLE_TEST_NO_AUTH=1 bats tests/history.bats tests/cli.bats tests/completion.bats
MOLE_TEST_NO_AUTH=1 ./scripts/test.sh
./scripts/check.sh --format


M-Hassan-Raza:
I know this is a rather big PR but I've done all the validation I possibly can at my end (including extensive handling of edge cases). But still I might've missed something. I don't expect this to merge soon so whenever you get the time to review, I'll be following up on any feedback I receive!

tw93:
@M-Hassan-Raza thank you for the PR. I misframed this earlier: the direction is acceptable, and `mo history` can be considered for merge.

The main boundary is that it should not become part of the primary entry/menu surface. README and help documentation are okay; I will review it with that expectation and keep the command discoverable without making it a main-entry workflow.

M-Hassan-Raza:
> @M-Hassan-Raza thank you for the PR. I misframed this earlier: the direction is acceptable, and `mo history` can be considered for merge.
> 
> The main boundary is that it should not become part of the primary entry/menu surface. README and help documentation are okay; I will review it with that expectation and keep the command discoverable without making it a main-entry workflow.

I was initially confused as well but now it makes perfect sense. Thanks so much for taking the time to bring this PR upto par with your commits :)

PR #983, opened 2026-05-26
fix(clean): silence protected symlink clean warnings

byronwang2005:
## Summary

- Silence protected symlink validation warnings for silent cleanup paths, so `mo clean` no longer prints repeated `Symlink points to protected system path` messages into the spinner line when skipping Office tmp symlinks.

## Safety Review

- Yes, this affects cleanup behavior and symlink handling.
- The safety boundary is unchanged: protected symlinks are still rejected and not deleted. The change only suppresses terminal stderr when `safe_remove` is called in `silent=true` mode, which is used by routine cleanup flows such as `safe_clean`.
- Direct validation and non-silent deletion paths still emit explicit validation errors.

## Tests

- `MOLE_TEST_NO_AUTH=1 bats tests/core_safe_functions.bats tests/file_ops_mole_delete.bats`
- `MOLE_TEST_NO_AUTH=1 bats tests/clean_user_core.bats`
- `./scripts/check.sh`
- Started `MOLE_TEST_NO_AUTH=1 ./scripts/test.sh`; it was interrupted after user confirmed local test pass.

## Safety-related changes

- None.


byronwang2005:
Thanks for your continued OSS work. I’ve used several of your projects and they’ve been solid and useful. Also looking forward to Luo’s progress!

tw93:
@byronwang2005 Thanks. The `safe_remove` change is good. Could you drop `CONTRIBUTORS.svg` from this PR though? That file is auto-regenerated by `.github/workflows/update-contributors.yml`, so manual edits get overwritten. After that it's good to merge.

PR #1004, opened 2026-05-29
fix: resolve "terminal is required" error when uninstalling Homebrew casks that run embedded sudo installers

### LINKED ISSUE #1003: [BUG] Permission issues when removing some Homebrew apps
## Describe the bug

Removing certain apps managed by Homebrew can be hindered by permission restrictions, resulting in the failure of the uninstall command.

## Steps to reproduce

1. Run command: `mo uninstall wireshark`
2. Enter password
3. See error

## Debug logs

```
Files to be removed:
◎ Homebrew apps will be fully cleaned, --zap removes configs and data

◎ Wireshark [Brew] , 449.3MB
  ✓ /Applications/Wireshark.app
  ◎ System: /Library/Application Support/Wireshark

➤ Remove 1 app, 449.3MB  Enter confirm, ESC cancel: 
/ Uninstalling Wireshark [Brew]...[DEBUG] Attempting brew uninstall --cask --zap wireshark-app
| Removing Wireshark (449.3MB)...[DEBUG] App size: 0GB, timeout: 300s
[TIMEOUT] Perl fallback, 300s: brew uninstall --cask --zap wireshark-app
- Removing Wireshark (449.3MB)...==> Running uninstall script /usr/sbin/installer
sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
sudo: a password is required
Error: Failure while executing; `/usr/bin/sudo -E -- /usr/sbin/installer -pkg /opt/homebrew/Caskroom/wireshark-app/4.6.6/Remove\ Wireshark\ from\ the\ system\ path.pkg -target /` exited with 1. Here's the output:
sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
sudo: a password is required

/ Removing Wireshark (449.3MB)...[DEBUG] brew uninstall timeout or failed with exit code: 1
- Removing Wireshark (449.3MB)...[DEBUG] brew uninstall failed: cask_gone=false app_gone=false
☻ Wireshark failed: brew uninstall failed, package still installed
   ☞ Run brew uninstall --cask --zap wireshark-app


Uninstall incomplete
• Failed: Wireshark brew uninstall failed, package still installed
☞ Run brew uninstall --cask --zap wireshark-app
```

## Environment

```
Mole version 1.39.1
macOS: 15.7.7
Architecture: arm64
```

manuelguido:
## Problem

Uninstalling certain Homebrew casks (e.g. Wireshark) fails with:
sudo: a terminal is required to read the password
sudo: a password is required
Error: Failure while executing; /usr/bin/sudo … /usr/sbin/installer … exited with 1.

Homebrew cask uninstall scripts can call `sudo` directly for embedded
`.pkg` installers. Mole runs `brew uninstall` through a timeout wrapper
whose Perl fallback (`run_with_timeout`) detaches the controlling
terminal, so the inner `sudo` has no TTY and cannot prompt for a password.

### `lib/core/sudo.sh` — `create_sudo_askpass_helper()`

Adds a new helper that writes a minimal temp script (`mktemp`, mode 700)
which shows a native macOS GUI password dialog via `osascript`. The
password is never written to disk; `sudo` only invokes the helper when it
actually needs one (a still-valid timestamp skips the prompt entirely).
The function is a no-op in test mode and on non-macOS systems.

### `lib/uninstall/brew.sh` — `brew_uninstall_cask()`

Before calling `brew uninstall`, exports `SUDO_ASKPASS` to the helper
path. Homebrew automatically appends `sudo -A` when `SUDO_ASKPASS` is
set, so any nested `sudo` inside a cask script can obtain the password
through the GUI without needing a TTY.

In the `sudo -u "$SUDO_USER" env …` branch the variable is re-injected
explicitly via the inner `env` list (the outer `sudo` strips the
environment). Cleanup (temp file removal + `SUDO_ASKPASS` restore) is
done inline after `brew` exits — not via a `RETURN` trap, which in bash
3.2 is not function-scoped and would re-fire on the caller's return with
locals out of scope, aborting under `set -u`.

### `tests/brew_uninstall.bats`

Two new tests:
- `brew_uninstall_cask exports SUDO_ASKPASS so inner sudo prompts work` — verifies the variable is visible to `brew` during uninstall.
- `brew_uninstall_cask restores a pre-existing SUDO_ASKPASS value` — verifies the prior value is restored after the call.

Resolves #1003

### DIFF
--- a/lib/core/sudo.sh
+++ b/lib/core/sudo.sh
@@ -310,3 +310,46 @@ stop_sudo_session() {
     fi
     MOLE_SUDO_ESTABLISHED="false"
 }
+
+# ============================================================================
+# Askpass Helper (for tools that invoke sudo internally)
+# ============================================================================
+
+# Create a temporary SUDO_ASKPASS helper script.
+#
+# Some tools (notably Homebrew cask uninstall scripts) invoke `sudo` directly
+# for embedded installer packages. When Mole runs these tools under its timeout
+# wrapper, the child may not inherit Mole's controlling terminal or its
+# tty-scoped sudo timestamp, so sudo fails with "a terminal is required to read
+# the password". Pointing SUDO_ASKPASS at this helper lets sudo obtain the
+# password through a GUI prompt without a terminal. The password is never
+# stored on disk; the helper prompts on demand only when sudo actually needs it.
+#
+# Prints: absolute path to the helper script on success.
+# Returns: 0 on success, 1 if a helper cannot be created (e.g. test mode).
+create_sudo_askpass_helper() {
+    # Never trigger real password or GUI prompts during tests.
+    if [[ "${MOLE_TEST_MODE:-0}" == "1" || "${MOLE_TEST_NO_AUTH:-0}" == "1" ]]; then
+        return 1
+    fi
+
+    # The GUI prompt requires osascript (macOS only).
+    command -v osascript > /dev/null 2>&1 || return 1
+
+    local helper
+    helper=$(mktemp "${TMPDIR:-/tmp}/mole-askpass.XXXXXX") || return 1
+
+    cat > "$helper" << 'ASKPASS'
+#!/bin/bash
+/usr/bin/osascript \
+    -e 'display dialog "Mole needs administrator access to finish removing this app." default answer "" with title "Mole" with icon caution with hidden answer' \
+    -e 'text returned of result' 2> /dev/null
+ASKPASS
+
+    if ! chmod 700 "$helper" 2> /dev/null; then
+        rm -f "$helper"
+        return 1
+    fi
+
+    echo "$helper"
+}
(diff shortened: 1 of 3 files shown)


tw93:
@manuelguido Thank you for the clear write-up. Your root-cause analysis (the Perl fallback detaching the TTY) is exactly right.

I fixed #1003 on main from the other direction: the timeout fallback now keeps the controlling terminal (`setpgid(0,0)` instead of `setsid()`), so the nested `sudo` can read the password on the existing terminal and reuse the already-cached credential, without adding an `osascript` password dialog (aa9a3711). Since that resolves the same root cause, I'm closing this as superseded.

I really appreciate the contribution and the tests; your TTY analysis directly informed the fix. Thank you!

manuelguido:
@tw93 Perfect! Way cleaner than my approach. Keeping the TTY directly makes much more sense. Thank you for this great project!

I stopped using other tools to clean my mac since I found this one 🔥

PR #1013, opened 2026-05-30
Paint mo status from fast metrics first

M-Hassan-Raza:
mo status now renders from fast CPU, memory, disk, network, and disk I/O collectors immediately, then enriches slower hardware and static metadata on the full refresh path.

This keeps process alerts fresh without making Bluetooth, GPU, hardware, and trash sizing block first paint.

Checks:
go test ./...
go test -race ./cmd/status
make build
./scripts/check.sh --format

### DIFF
--- a/cmd/status/main.go
+++ b/cmd/status/main.go
@@ -14,7 +14,11 @@ import (
 	"github.com/charmbracelet/lipgloss"
 )
 
-const refreshInterval = time.Second
+const (
+	refreshInterval      = time.Second
+	processWatchInterval = refreshInterval
+	slowRefreshInterval  = 30 * time.Second
+)
 
 var (
 	// Command-line flags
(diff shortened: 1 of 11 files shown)


tw93:
@M-Hassan-Raza thank you, this is merged.

I added the missing `CollectProcesses()` regression coverage before merging, so live `TopProcesses` and `ProcessAlerts` stay fresh while the slow enrichment fields can still reuse cache.

M-Hassan-Raza:
> @M-Hassan-Raza thank you, this is merged.
> 
> I added the missing `CollectProcesses()` regression coverage before merging, so live `TopProcesses` and `ProcessAlerts` stay fresh while the slow enrichment fields can still reuse cache.

Lmao i first added that and then intentionally removed to keep the PR bloat free :D Thanks for merging

PR #1038, opened 2026-06-02
fix(status): report process resident memory

### LINKED ISSUE #1030: [Mac App Bug] Process memory values look much lower than Activity Monitor
Hi, thanks for building Mole! I noticed a possible issue on the Mac app Status page.

## What happened

On the Status page, the process list's **Memory** column appears to show much smaller values than macOS Activity Monitor for the same moment.

For example, Activity Monitor shows several top processes using memory in the GB range, such as roughly:

- one process around 6.70 GB
- several processes around 1.8-2.8 GB

But Mole's process list on the Status page shows the Memory column mostly around a few hundred MB, for example about 435 MB, 403 MB, 351 MB, 234 MB, etc.

The overall Memory card still shows high memory usage, so this looks specific to the per-process memory values in the process list.

## Expected behavior

The per-process memory values in Mole should be close to the values shown by Activity Monitor, or the UI should make it clear if Mole intentionally uses a different memory metric.

## Actual behavior

The per-process memory values shown in Mole look much lower than Activity Monitor, which makes it hard to identify high-memory processes from the Status page.

## Environment from the screenshot

- macOS: 26.5
- Apple Silicon: M5
- Mole Mac app Status page
- Memory pressure shown in Mole: 56%
- Overall memory usage shown in Mole: 77%

## Notes

I only noticed this from a screenshot comparison, so I may be missing some implementation detail. If Mole intentionally displays a different metric than Activity Monitor, maybe the label or tooltip could clarify that. Thanks!

RajvardhanPatil07:
## Summary

- Add `memory_bytes` to status process snapshots from `ps` RSS values.
- Keep the existing `memory` percent field for compatibility and render resident memory in the process card when available.

Fixes #1030

## Safety Review

- Does this change affect cleanup, uninstall, optimize, installer, remove, analyze delete, update, or install behavior? No.
- Does this change affect path validation, protected directories, symlink handling, sudo boundaries, or release/install integrity? No.

## Tests

- `/Users/raj/.gimme/versions/go1.26.3.darwin.arm64/bin/go test ./cmd/status ./internal/units`
- `/Users/raj/.gimme/versions/go1.26.3.darwin.arm64/bin/go test ./...` failed in existing `cmd/analyze` `TestDeletePathWithProgress` with `timeout moving to Trash`; touched status package passed.

## Safety-related changes

- None.

### DIFF
--- a/cmd/status/metrics.go
+++ b/cmd/status/metrics.go
@@ -102,12 +102,13 @@ type DiskIOStatus struct {
 }
 
 type ProcessInfo struct {
-	PID     int     `json:"pid"`
-	PPID    int     `json:"ppid"`
-	Name    string  `json:"name"`
-	Command string  `json:"command"`
-	CPU     float64 `json:"cpu"`
-	Memory  float64 `json:"memory"`
+	PID         int     `json:"pid"`
+	PPID        int     `json:"ppid"`
+	Name        string  `json:"name"`
+	Command     string  `json:"command"`
+	CPU         float64 `json:"cpu"`
+	Memory      float64 `json:"memory"` // Percent of physical memory, kept for compatibility.
+	MemoryBytes uint64  `json:"memory_bytes,omitempty"`
 }
 
 type CPUStatus struct {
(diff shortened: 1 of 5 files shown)


tw93:
@RajvardhanPatil07 Thanks for the clean implementation. RSS display and the backward-compatible fallback for the old 5-column ps shape are both well handled. Merged.

PR #1039, opened 2026-06-02
fix(clean): protect Raycast cache state

### LINKED ISSUE #1033: [BUG]
Hey @tw93 

I have really been loving Mole for a while now, but since the new Raycast Beta update (https://raycast.com/new) I believe they've moved some folders or caches to different directories, and those directories ARE getting cleaned. Could you possibly investigate the new Raycast and whitelist some directories there? I'm not sure which ones, but I think AI chats, clipboard histories (perhaps, not sure), and other things are getting messed up.

https://github.com/tw93/Mole/issues/174

Maybe you could also invent some new file format/schema thing where inside an app's package contents, the developer must attach some `.mole` file where they list all the directories that should be excluded from the cleanup, so you don't have to whitelist stuff manually anymore.

Thanks,
Visual-Studio-Coder

RajvardhanPatil07:
## Summary

- Stop deleting Raycast `urlcache` and `fsCachedData` during launcher cleanup.
- Protect the Raycast cache root during normal cleanup so moved beta state is not removed by broader cache paths.
- Add focused coverage for the launcher skip and path protection behavior.

Fixes #1033

## Safety Review

- Does this change affect cleanup, uninstall, optimize, installer, remove, analyze delete, update, or install behavior? Yes, cleanup now skips Raycast cache state instead of deleting selected Raycast cache subfolders.
- Does this change affect path validation, protected directories, symlink handling, sudo boundaries, or release/install integrity? Yes, Raycast cache paths are added to normal cleanup protection only. Uninstall mode is left unchanged for user-selected Raycast removal.

## Tests

- `/bin/bash -n lib/core/app_protection.sh lib/clean/app_caches.sh tests/clean_app_caches.bats tests/core_common.bats`
- Manual `/bin/bash` scenario for `clean_launcher_apps` with Raycast `urlcache` and `fsCachedData` present.
- Manual `/bin/bash` scenario for `should_protect_path` on a Raycast cache child path.

Bats, shfmt, and shellcheck were not installed on this machine, so the exact Bats files could not be run locally.

## Safety-related changes

- Raycast cache paths are protected during normal cleanup to avoid removing clipboard, chat, or other state that Raycast stores under cache-shaped directories.

### DIFF
--- a/lib/clean/app_caches.sh
+++ b/lib/clean/app_caches.sh
@@ -620,9 +620,10 @@ clean_note_apps() {
 clean_launcher_apps() {
     safe_clean ~/Library/Caches/com.runningwithcrayons.Alfred/* "Alfred cache"
     safe_clean ~/Library/Caches/cx.c3.theunarchiver/* "The Unarchiver cache"
-    # Raycast: only clean network and FS caches; Clipboard subfolder contains user's clipboard history.
-    safe_clean ~/Library/Caches/com.raycast.macos/urlcache/* "Raycast URL cache"
-    safe_clean ~/Library/Caches/com.raycast.macos/fsCachedData/* "Raycast FS cache"
+    if [[ -d "$HOME/Library/Caches/com.raycast.macos" ]]; then
+        echo -e "  ${GRAY}${ICON_WARNING}${NC} Raycast cache · skipped by default"
+        note_activity
+    fi
 }
 # Remote desktop tools.
 clean_remote_desktop() {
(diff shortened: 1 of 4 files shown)


tw93:
@RajvardhanPatil07 Thanks. I simplified the approach before merging: removed the two safe_clean lines directly instead of adding should_protect_path overrides, since the only cleanup entry point for Raycast cache was in clean_launcher_apps. Simpler diff, same result.

PR #1072, opened 2026-06-06
fix: check scan deadline inside nested-dir loop to prevent hung Project artifacts (#1053)

### LINKED ISSUE #1053: [BUG] Stuck on "Project artifacts"
## Describe the bug

Its getting stuck on "Project artifacts" no errors or anything just stuck

## Steps to reproduce

1. Run command: `mo clean`

## Debug logs

➤ Device backups & firmware
  ✓ Nothing to clean

➤ Time Machine
  ✓ No incomplete backups found
  ✓ Nothing to clean

➤ Large files
  ◎ Xcode archives (review only): 6.86GB, Path: /Users/sukuna/Library/Developer/Xcode/Archives

➤ System Data clues
  • Xcode Archives: 6.86GB
  ↳ Path: ~/Library/Developer/Xcode/Archives
  ☞ Review: mo analyze, Device backups, docker system df

➤ Project artifacts

</details>

## Environment

Mole version 1.41.0
macOS: 26.5
Architecture: arm64
Kernel: 25.5.0
SIP: Enabled
Disk Free: 45.47GB
Install: Manual
Shell: /bin/zsh

Image

Puhavik:
## Problem

Fixes #1053 — `mo clean` would hang indefinitely on the **Project artifacts** section.

`probe_project_artifact_hints` had a wall-clock budget (`MOLE_TIMEOUT_HINT_SCAN_SEC`, default 15 s), but the innermost `while` loop over nested subdirectories had no deadline check. When a single scan root is used, the outer-root guard fires only at the **top** of each root iteration — once that iteration is entered it never re-checks the budget. A slow `hint_collect_child_dirs_with_timeout` call (large monorepo, cold disk) pushed `SECONDS` past the deadline, but the nested-dir loop ran all the way to `max_nested_per_project` (120) iterations before control returned to any existing guard.

## Fix

Two new checkpoints in `lib/clean/hints.sh`:

1. **Before** `hint_collect_child_dirs_with_timeout` for nested dirs — skips the `find` call entirely if the budget is already exhausted.
2. **At the top of each nested-dir iteration** — stops immediately when `SECONDS` crosses the deadline mid-loop.

Also removes a dead `[[ "$stop_scan" == "true" ]] && break` that followed the `for target_names` loop inside the nested-dir `while`: at that point `stop_scan` could never be true (it is only set in the `project_dir` loop above, which breaks before reaching nested processing).

## Test

New bats test in `tests/clean_hints.bats`:

- Creates one scan root with one project containing two nested sub-projects (each with a `build/` artifact).
- Mocks `hint_collect_child_dirs_with_timeout` to sleep 2 s on the nested call, advancing `SECONDS` past the 1 s budget before the nested-dir loop starts.
- **New code**: deadline fires on the first nested-dir iteration → `count=0`, `skipped=true`.
- **Old code**: nested loop ran without a check → `count=2`, `skipped=false`.

```
ok 6 probe_project_artifact_hints respects budget inside nested-dir loop (#1053)
```

All 24 hints tests pass.

### DIFF
--- a/lib/clean/hints.sh
+++ b/lib/clean/hints.sh
@@ -410,6 +410,13 @@ probe_project_artifact_hints() {
             done
             [[ "$stop_scan" == "true" ]] && break
 
+            if [[ $SECONDS -ge $scan_deadline ]]; then
+                PROJECT_ARTIFACT_HINT_TRUNCATED=true
+                PROJECT_ARTIFACT_HINT_SCAN_SKIPPED=true
+                stop_scan=true
+                break
+            fi
+
             local nested_count=0
             nested_dirs_file=$(mktemp_file "project_artifact_nested") || {
                 PROJECT_ARTIFACT_HINT_SCAN_SKIPPED=true
@@ -424,6 +431,12 @@ probe_project_artifact_hints() {
             fi
 
             while IFS= read -r -d '' nested_dir; do
+                if [[ $SECONDS -ge $scan_deadline ]]; then
+                    PROJECT_ARTIFACT_HINT_TRUNCATED=true
+                    PROJECT_ARTIFACT_HINT_SCAN_SKIPPED=true
+                    stop_scan=true
+                    break
+                fi
                 [[ -d "$nested_dir" ]] || continue
 
                 local nested_name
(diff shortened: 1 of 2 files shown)


tw93:
@Puhavik Thanks for the careful fix and the focused regression test. The nested-dir deadline gap was real, and this is now merged into `main` for the next release.

PR #1079, opened 2026-06-08
fix: read system service plists with sudo

honghoker:
## Summary
- Read system service plist `ProgramArguments:0` / `Program` values with `sudo PlistBuddy`.
- Reject non-absolute plist read output before using it as a service binary path.
- Add focused coverage for unreadable LaunchDaemon plists that point to an existing binary.

## Safety Review
 - Does this change affect cleanup, uninstall, optimize, installer, remove, analyze delete, update, or install behavior? Yes, `mo clean` orphaned system service detection now reads
  LaunchDaemon/LaunchAgent plist contents through the same sudo boundary used to discover those plist files.
- Does this change affect path validation, protected directories, symlink handling, sudo boundaries, or release/install integrity? Yes, sudo boundary handling changes for reading `/
  Library/LaunchDaemons` and `/Library/LaunchAgents` plist contents only. Deletion, path validation, symlink handling, protected-path checks, dry-run behavior, `launchctl unload`, and `safe_sudo_remove` are unchanged.

## Tests
- Manual `/bin/bash` scenario for an unreadable system-service plist whose `Program` points to an existing binary.
- Manual `/bin/bash` scenario for `PlistBuddy` failure text on stdout, verifying it is not treated as a binary path.

Bats was not available locally, so the new Bats case was covered with the equivalent manual shell scenario.

## Safety-related changes
- System service plist reads now use `sudo PlistBuddy`, matching the existing `sudo find` scan boundary.
- Non-absolute plist read output is rejected before orphan detection, preventing `PlistBuddy` failure text from being interpreted as a missing binary path.

### DIFF
--- a/lib/clean/apps.sh
+++ b/lib/clean/apps.sh
@@ -526,7 +526,8 @@ clean_orphaned_system_services() {
     }
 
     # Read a launchd program path from a system plist.
-    # The plist itself was discovered with sudo, so read it with sudo too:
+    # The plist itself was discovered with sudo, so read it with sudo too (the
+    # caller already cleared a `sudo -n true` probe, so keep it non-interactive):
     # unreadable root-owned plists make PlistBuddy print a non-path "File Doesn't
     # Exist, Will Create..." message on stdout, which must never be treated as a
     # missing binary path.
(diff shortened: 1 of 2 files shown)


tw93:
@honghoker Thank you, this is a sharp catch. An unreadable root-owned plist makes PlistBuddy print "File Doesn't Exist, Will Create..." to stdout, and without your absolute-path guard that text would have been treated as a missing binary and flagged the plist as a false orphan. Merged.

One small adjustment before merging: I changed the read from `sudo` to `sudo -n` PlistBuddy. We just landed #1085, which routes all of `mo clean`'s privileged calls through non-interactive sudo so nothing can prompt for a password behind the spinner (#1084). Plain `sudo` here would have reopened that hole during the orphan scan; `sudo -n` is safe because `clean_orphaned_system_services` already clears a `sudo -n true` probe before it runs. I also updated your test's sudo mock to absorb the `-n` flag.

It's on `main` now and ships in the next release; you can try it immediately with `mo update --nightly`. Thanks again for hardening this path.

PR #1086, opened 2026-06-08
fix: clean up Docker privileged helpers and LaunchDaemon plists after uninstall

### LINKED ISSUE #1082: [BUG] Docker uninstall leaves privileged helper files and background activity items behind
## Describe the bug

After uninstalling Docker Desktop using Mole, Docker-related background items were still present on macOS.

Docker / Docker Inc continued to appear in:

System Settings → General → Login Items & Extensions → App Background Activity

On checking manually, Docker LaunchDaemons and PrivilegedHelperTools were still present under `/Library`.

## Steps to reproduce

1. Install Docker Desktop on macOS.
2. Uninstall Docker Desktop using Mole.
3. Open:

   System Settings → General → Login Items & Extensions → App Background Activity

4. Notice Docker / Docker Inc still appears.
5. Run:

```bash
sudo find /Library/LaunchDaemons /Library/LaunchAgents ~/Library/LaunchAgents /Library/PrivilegedHelperTools -iname "*docker*" -print
```

6. The following files are still present:

```text
/Library/LaunchDaemons/com.docker.vmnetd.plist
/Library/LaunchDaemons/com.docker.socket.plist
/Library/PrivilegedHelperTools/com.docker.socket
/Library/PrivilegedHelperTools/com.docker.vmnetd
```

## Expected behavior

Mole should either remove these Docker-related LaunchDaemons and PrivilegedHelperTools during uninstall, or detect and show them as leftover files that can be removed.

## Debug logs

I did not capture the debug output at the time of uninstall.

```bash
mo <command> --debug
```

<details>
<summary>Debug output</summary>

```text
Not available
```

</details>

## Environment

Please fill this with your local output:

```text
Mole version 1.41.0 macOS: 26.5.1 Architecture: arm64 Kernel: 25.5.0 SIP: Enabled Disk Free: 109.27GB Install: Homebrew Shell: /bin/zsh
```

## Additional context

Manually removing the leftover files and rebooting cleared Docker from App Background Activity.

Commands used:

```bash
sudo rm -f /Library/LaunchDaemons/com.docker.vmnetd.plist
sudo rm -f /Library/LaunchDaemons/com.docker.socket.plist
sudo rm -f /Library/PrivilegedHelperTools/com.docker.vmnetd
sudo rm -f /Library/PrivilegedHelperTools/com.docker.socket
```

After this, Docker no longer appeared in macOS App Background Activity.

These files appear to be Docker Desktop privileged helper tools and LaunchDaemons, so Mole may need elevated permissions to detect/remove them.

imrajyavardhan12:
## Summary

Fixes #1082

Docker uninstall was leaving behind privileged helper files (`/Library/PrivilegedHelperTools/com.docker.vmnetd`, `com.docker.socket`) and their corresponding LaunchDaemon plists, causing them to appear in System Settings > Login Items > Background Activity.

## Root Cause

Two distinct bugs in `clean_orphaned_system_services()`:

1. **Overly broad data protection**: `should_protect_path()` blocked cleanup of orphaned Docker files because `should_protect_data()` matched `com.docker.*` on the filename, even though the parent app was already verified as uninstalled by the orphan detection logic.

2. **Incomplete orphan detection**: LaunchDaemon plists were not detected as orphaned when their target binary still existed in `/Library/PrivilegedHelperTools/`. The `_plist_is_orphaned()` function only checked binary existence, not whether the parent app was still installed.

### Fix 1: Enable uninstall mode during orphaned services cleanup (`lib/clean/apps.sh:718-754`)
- Set `MOLE_UNINSTALL_MODE=1` before the cleanup loop to skip the `should_protect_data()` filename check in `should_protect_path()`
- Added trap (`RETURN ERR INT TERM`) to ensure `MOLE_UNINSTALL_MODE` is always unset, even on error or interrupt
- Still protects truly system-critical components (`SYSTEM_CRITICAL_BUNDLES`)

### Fix 2: Detect orphaned PrivilegedHelperTools binaries (`lib/clean/apps.sh:567-585`)
- Enhanced `_plist_is_orphaned()` to check if a plist's binary lives in `/Library/PrivilegedHelperTools/`
- When the binary exists, uses `bundle_has_installed_app()` to verify the parent app is still installed
- Marks the plist as orphaned if the parent app is gone, even though the binary still exists on disk

## Testing

- ✅ `clean_apps.bats`: 25/25 tests passed (including all orphaned system services tests)
- ✅ `clean_app_caches.bats`: 36/36 tests passed
- ✅ `uninstall_safety.bats`: 13/13 tests passed (including Docker auth token preservation)
- ✅ `scripts/check.sh`: ShellCheck + syntax check passed
- ✅ `go test ./...`: passed

### DIFF
--- a/lib/clean/apps.sh
+++ b/lib/clean/apps.sh
@@ -718,14 +718,13 @@ clean_orphaned_system_services() {
         local failed_count=0
         local removed_kb=0
 
-        # Treat orphaned services like uninstall targets: we've already verified
-        # the parent app is gone, so skip data-protection checks that would block
-        # cleanup of legitimately orphaned files (e.g., Docker helpers). See #1082.
-        # Use trap to ensure cleanup even on error/interrupt.
-        MOLE_UNINSTALL_MODE=1
-        trap 'unset MOLE_UNINSTALL_MODE; trap - RETURN ERR INT TERM' RETURN ERR INT TERM
         for orphan_file in "${orphaned_files[@]}"; do
-            if should_protect_path "$orphan_file"; then
+            # Orphans were already verified to have no installed parent app, so
+            # bypass the data-protection filename check (which would otherwise block
+            # legitimately orphaned files like Docker helpers) for this single call.
+            # MOLE_UNINSTALL_MODE is scoped to the call and never leaks to later
+            # cleanup sections; SYSTEM_CRITICAL_BUNDLES stay protected. See #1082.
+            if MOLE_UNINSTALL_MODE=1 should_protect_path "$orphan_file"; then
                 debug_log "Skipping protected orphaned service: $orphan_file"
                 skipped_protected_count=$((skipped_protected_count + 1))
                 continue
@@ -750,8 +749,6 @@ clean_orphaned_system_services() {
                 fi
             fi
         done
-        unset MOLE_UNINSTALL_MODE
-        trap - RETURN ERR INT TERM
 
         local orphaned_kb_display
         if [[ $removed_kb -gt 1024 ]]; then
--- a/lib/clean/apps.sh
+++ b/lib/clean/apps.sh
@@ -567,8 +567,22 @@ clean_orphaned_system_services() {
         local binary
         binary=$(_plist_binary_path "$plist") || return 1 # no Program key → skip
 
-        # If the binary still exists, the service is healthy.
-        [[ -e "$binary" ]] && return 1
+        # If the binary still exists, check if it's in PrivilegedHelperTools.
+        # If so, verify the parent app is still installed. If the parent app
+        # is gone, the binary itself is orphaned, so this plist is too. See #1082.
+        if [[ -e "$binary" ]]; then
+            if [[ "$binary" == /Library/PrivilegedHelperTools/* ]]; then
+                local helper_bundle_id
+                helper_bundle_id=$(basename "$binary")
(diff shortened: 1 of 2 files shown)


tw93:
@imrajyavardhan12 Thank you, this is a well-diagnosed fix. You correctly pinned both halves of #1082: the orphaned LaunchDaemon plists whose binary still sits in `/Library/PrivilegedHelperTools`, and the `should_protect_data` match on `com.docker.*` that was blocking cleanup. Merged.

I made two small adjustments before merging:

1. **Scoped the uninstall-mode bypass.** Instead of setting `MOLE_UNINSTALL_MODE=1` for the whole cleanup loop plus a `trap ... RETURN ERR INT TERM`, I moved it to a per-call prefix: `MOLE_UNINSTALL_MODE=1 should_protect_path "$orphan_file"`. The trap variant cleared `bin/clean.sh`'s own `INT`/`TERM` handlers for the rest of the run, so a Ctrl-C after this section would have skipped cleanup. The per-call form can't leak into later cleanup sections (verified the flag is unset after the call on bash 3.2).

2. **Added a regression test** in `tests/clean_apps.bats` that a data-protected orphan (`com.getpostman.*`, same `should_protect_data` branch as Docker) is removed rather than skipped, routed through `/Library/LaunchDaemons` so it runs on CI.

It's on `main` now and ships in the next release; you can grab it immediately with `mo update --nightly`. Nice work.

Issue #1062, opened 2026-06-05
[FEATURE] See folder structure before sizes are in


philippb:
## Feature description
<img width="488" height="335" alt="Image" src="https://github.com/user-attachments/assets/670f5f69-1f97-48ef-9bc2-99abe1c3318d" />

When opening a folder, I'd love to see all the sub folders right away even before the final sizes are in, so I can navigate.
I'm using Mole heavy for dev workflows where there are large folders with many files of build artifacts. So it takes a long time to see the content, and sometime I just know where to navigate and then to see a more narrow picture.

## Proposed solution
When you open a folder, you see the subfolders right away, with a basic `ls`

Then as the scanner is running, I can see in real time the folder size grow with an animation that indicates that the folder is still getting scanned. 
Possibly the indicators to items that are getting scanned can show the % of total files already scanned.

When I navigate to a subfolder, it stops scanning the other folders on the same level and just continues with the content of the subfolder.

## Implementation
I'm happy to take a spin at this and open a PR if you want me to.

tw93:
@philippb Thanks, this is a nice idea: an `ls`-first listing with sizes filling in as the scan runs, and sibling scans pausing when you drill in. A PR would be very welcome. The relevant code is the Bubble Tea model in `cmd/analyze/` (`update.go` holds the navigation/message chain). Happy to review.

philippb:
Playing around with this... the UX is a little tricky. I'm right now thinking that keeping the folders updating and changing order while the load is the best.
I keep the cursor on the index, even if the folder moves. For most sub folders on my machine it settles in pretty quickly. 

https://screen.studio/share/TG62p8YU

Let me know what you think. Happy to refine or just open a draft PR for you to play with it. 
To me it feels better than waiting for the results....

tw93:
@philippb Thanks for sharing the demo. I watched it, and I agree this feels much better than waiting for the full scan before showing anything.

I think this is worth exploring as a draft PR. One constraint I would like to keep clear: while rows are updating and reordering, the focused item should stay tied to the path, not only the visible index, so Enter, Open, and Delete cannot accidentally act on a different folder after a reorder.

A good first version would be:
1. show immediate children with pending sizes,
2. update each row as its child scan finishes,
3. keep navigation responsive while scanning,
4. preserve focus and multi-select by path across reorders,
5. when drilling into a child, prioritize that path and avoid spending more work on siblings where practical.

Please open a draft PR when you are ready. I can review the UX and code path from there.

[...]

PR #1095, opened 2026-06-09
Draft: Explore ls-first analyze navigation

philippb:
## Summary

Draft PR for #1062. This explores an ls-first navigation flow for `mo analyze`: rows render from `os.ReadDir` immediately, then child directory sizes fill in live as scans progress.

This follows the issue discussion with @tw93:
- show immediate children with pending sizes instead of waiting for the full scan,
- update rows as child scans make progress,
- keep navigation responsive while scanning,
- preserve or experiment with focus behavior across live reorders,
- when drilling into a child, cancel/suppress the parent live session and avoid saving a stale completed parent snapshot.

Demo from the issue thread:
https://screen.studio/share/TG62p8YU

## UX options to try

The current defaults favor continuous sort + cursor-by-index because that makes the live scan easy to feel out while rows settle:

- `MOLE_ANALYZE_LIVE_SORT=continuous` (default): rows keep sorting by current known size.
- `MOLE_ANALYZE_LIVE_SORT=freeze-on-move`: rows sort live until the first up/down key, then freeze.
- `MOLE_ANALYZE_LIVE_CURSOR=index` (default): selection stays on the visible row number as rows move.
- `MOLE_ANALYZE_LIVE_CURSOR=path`: selection follows the selected path across reorders.

While a scan is active:

- `s` toggles live sort mode.
- `c` toggles cursor anchoring.

Example:

```bash
MOLE_ANALYZE_LIVE_SORT=freeze-on-move MOLE_ANALYZE_LIVE_CURSOR=path mo analyze ~
```

## Implementation notes

- Adds a live scan session for non-overview TUI scans.
- Cache hits and JSON/overview behavior stay on existing paths.
- Live child progress events update row size, bar, percent, and scanning marker.
- Completed child scans warm child cache entries.
- Session ids/path checks ignore stale events after navigation.
- Active scans are canceled on navigation, refresh, back, and delete; completed parent cache is only written for the active completed session.
- The row divider before the folder icon spins for folders that are still scanning.

## Review focus

This is intentionally draft UX work. I’d especially like feedback on:

- whether the default should be cursor-by-index or cursor-by-path,
- whether rows should continuously reorder or freeze after navigation,
- whether the spinner divider is the right affordance for still-scanning rows,
- whether the live progress cadence feels too noisy or too slow.

One implementation constraint still worth reviewing: deep filesystem walks use the existing scanner internals, so cancellation suppresses stale UI/events and avoids writing a completed parent cache, but some underlying IO may continue until the current subtree walk returns.

## Verification

- `MOLE_SKIP_FINDER_TESTS=1 go test -count=1 ./cmd/...`
- `go test -race -run 'TestLiveScan|TestOverviewHome|TestScanningView' -count=1 ./cmd/analyze`
- `git diff --check`

The unskipped full `go test -count=1 ./cmd/...` hit Finder busy errors in existing Trash/Finder-dependent tests locally (`-15260`).

### DIFF
--- a/cmd/analyze/analyze_test.go
+++ b/cmd/analyze/analyze_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"slices"
 	"strconv"
 	"strings"
 	"sync/atomic"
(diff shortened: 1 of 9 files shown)


philippb:
@tw93 play around with it and let me know what you think.

tw93:
@philippb thanks for the thoughtful PR. I pushed a maintainer update to this branch to rebase it onto current main, keep the live scan rows working with the current filter backing lists, and make the selection path-based so live sorting cannot move the cursor onto a different entry.

CI is green now. The PR is mergeable, and I am leaving it as draft until you want to mark it ready.

philippb:
Glad you merged it.
Happy to contribute!! I use mo a lot.

One thing that comes to mind... how to make it better for coding agents to
use it. :)

Glad this helped

PR #1113, opened 2026-06-13
Avoid recursive container stub cleanup

M-Hassan-Raza:
Summary:
Change orphaned container-stub cleanup to remove only the verified metadata plist, then rmdir the container instead of recursively deleting the directory.

Why:
The old path verified a metadata-only container, then ran rm -rf on the directory. If anything appeared after verification, that recursive delete could remove content that was no longer part of the verified stub.

Validation:
./scripts/check.sh --format
./scripts/check.sh
env -u NO_COLOR MOLE_TEST_NO_AUTH=1 bats tests/clean_apps.bats
MOLE_TEST_NO_AUTH=1 bats --filter "clean_orphaned_container_stubs preserves content that appears during removal" tests/clean_apps.bats
env -u NO_COLOR MOLE_TEST_NO_AUTH=1 ./scripts/test.sh

### DIFF
--- a/lib/clean/apps.sh
+++ b/lib/clean/apps.sh
@@ -801,6 +801,23 @@ clean_orphaned_system_services() {
 # A stub container contains only .com.apple.containermanagerd.metadata.plist
 # with no Data/ subdirectory — it holds no user data and is safe to remove.
 # Only targets a hardcoded allowlist of apps known to leave such stubs.
+_remove_verified_container_stub() {
+    local container_dir="$1"
+    local metadata_plist="$2"
+
+    [[ -d "$container_dir" ]] || return 1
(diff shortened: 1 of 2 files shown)


tw93:
@M-Hassan-Raza Merged, thanks. Replacing the `rm -rf` with a targeted `rm -f` on the metadata plist plus a `rmdir` is the right call: if anything repopulates the container during the mdfind/whitelist window, `rmdir` refuses the non-empty directory and the new content survives instead of getting recursively deleted.

The regression test that injects raced content and asserts it's preserved is a nice touch, it fails on the old path and passes on the new one. Verified locally with the format check and the full `clean_apps.bats` suite (28/28 green).

PR #1138, opened 2026-06-19
feat(status): add --watch streaming NDJSON mode

braianj:
## What

Adds a continuous **`--watch`** streaming mode to `mo status`, alongside the existing one-shot `-json` and interactive TUI modes.

- `--watch` — stream one full `MetricsSnapshot` per line as **NDJSON** from a single long-lived `Collector`, so network/disk-IO **rates stay accurate** across ticks (a warm collector keeps the previous counters).
- `--interval <d>` — collection cadence for `--watch` (e.g. `1s`, `2s`; defaults to `1s`).
- `--listen <socket>` — serve the same NDJSON stream over a **unix socket** (mode `0666`) instead of stdout, so a root `LaunchDaemon` can feed an unprivileged client fully-populated GPU/temp/fan metrics.

The watch loop mirrors the TUI's collection cadence: a fast first paint, a full collect on the next tick to prime the enrichment cache (GPU/thermal/batteries/disks), then mostly fast collects with periodic process/full refreshes. The socket is removed on `SIGINT`/`SIGTERM` (e.g. `launchctl bootout`).

## Why

The Topo / MoleUI menu-bar app consumes `status-go --watch [--interval] [--listen]` as its metrics source. Vanilla `mo status` only supports a one-shot `-json` dump, so the app's helper exited immediately with `flag provided but not defined: -watch`, leaving the UI stuck on a loading state and the privileged metrics daemon in a crash/restart loop.

## How it was verified

- `gofmt`, `go vet ./cmd/status`, `go build`, `go test ./cmd/status` — all pass.
- **stdout mode** (`--watch --interval 1s`): each line parses as standalone JSON with live `cpu`/`memory`/`top_processes`; no stderr noise.
- **socket mode** (`--watch --listen <sock>`): socket created `srw-rw-rw-`, an `AF_UNIX` client reads valid NDJSON snapshots; `SIGTERM` removes the socket.
- End-to-end: rebuilt the helper, the real privileged `LaunchDaemon` came up and served its socket instead of crash-looping.

## Notes

- No changes to existing flags or output shape — `-json`/TUI behavior is unchanged.
- New code is isolated in `cmd/status/watch.go`; `main.go` only wires the flags and dispatch.

### DIFF
--- a/cmd/status/main.go
+++ b/cmd/status/main.go
@@ -26,6 +26,12 @@ var (
 	procCPUThreshold = flag.Float64("proc-cpu-threshold", 100, "alert when a process stays above this CPU percent")
 	procCPUWindow    = flag.Duration("proc-cpu-window", 5*time.Minute, "continuous duration a process must exceed the CPU threshold")
 	procCPUAlerts    = flag.Bool("proc-cpu-alerts", true, "enable persistent high-CPU process alerts")
+
+	// Watch mode (consumed by MoleUI): stream NDJSON (one snapshot per line)
+	// from a single warm collector, optionally over a unix socket.
+	watchMode     = flag.Bool("watch", false, "stream metrics continuously as newline-delimited JSON instead of the one-shot TUI/JSON")
+	watchListen   = flag.String("listen", "", "with --watch, serve the NDJSON stream over this unix socket path instead of stdout")
+	watchInterval = flag.String("interval", "", "with --watch, collection interval (e.g. 1s, 2s); defaults to 1s")
 )
 
 func shouldUseJSONOutput(forceJSON bool, stdout *os.File) bool {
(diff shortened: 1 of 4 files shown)


tw93:
@braianj thank you for the PR and for explaining the menu-bar use case.

I pushed a maintainer revision that keeps `--watch` as a stdout NDJSON stream, shares the same collection cadence as the TUI, and adds Go plus CLI coverage.

I removed `--listen` for now because a unix socket would be a new IPC contract for `mo status`, and the original path handling removed the requested path before binding. That is too broad for this CLI surface today. Stdout `--watch` gives us the useful long-lived collector without adding a socket lifecycle to support.

Verified: `./scripts/check.sh --format`, `go test ./...`, `go vet ./cmd/status`, `MOLE_TEST_NO_AUTH=1 bats tests/cli.bats`, plus a wrapper smoke for `./mole status --watch`.

Issue #1145, opened 2026-06-22
[BUG] Mole does not recognize Anki's user Application Support files when uninstalling it


zach12211:
## Describe the bug

When uninstalling the flash card app [Anki](https://apps.ankiweb.net), Mole does not recognize its user Application Support directories:

`/Users/zachary/Library/Application Support/Anki2`
`/Users/zachary/Library/Application Support/AnkiProgramFiles`

## Steps to reproduce

1. Run command: `mo uninstall Anki`
2. Enter `y` to proceed with Anki uninstallation
3. See error

## Expected behavior

I expect the Anki Application Support files to be found and deleted.

## Debug logs

```
----------------------------------------------------------------------
Mole Debug Session, 2026-06-22 17:24:42
----------------------------------------------------------------------
User: zachary
Hostname: Zacharys-MacBook-Pro.local
Architecture: arm64
Kernel: 24.6.0
macOS: 15.7.7, 24G720
Shell: /bin/zsh, xterm-256color
Sudo Access: Required
----------------------------------------------------------------------
[2026-06-22 17:28:33] DEBUG: [DRY RUN] Would unload launch services for bundle: net.ankiweb.anki
[2026-06-22 17:28:33] DEBUG: [DRY RUN] Would remove login item: Anki
[2026-06-22 17:28:33] DEBUG: [DRY RUN] Would terminate running app: Anki
[2026-06-22 17:28:33] DEBUG: [DRY RUN] Would delete (trash): /Applications/Anki.app
[2026-06-22 17:28:33] DEBUG: [DRY RUN] Would delete (trash): /Users/zachary/Library/Caches/Anki
[2026-06-22 17:28:34] DEBUG: [DRY RUN] Would clear defaults domain: net.ankiweb.anki
[2026-06-22 17:28:34] INFO: [DRY RUN] Would refresh LaunchServices and update Dock entries
```

## Environment

Please run `mo update` to ensure you are on the latest version, then paste the output of `mo --version` below:

```
Mole version 1.44.0
macOS: 15.7.7
Architecture: arm64
Kernel: 24.6.0
SIP: Enabled
Disk Free: 102.88GB
Install: Manual
Shell: /bin/zsh
```

## Additional context

<img width="638" height="439" alt="Image" src="https://github.com/user-attachments/assets/9f38ac66-0ffc-441a-976d-2b46f09606a5" />

tw93:
[...]

PR #1146, opened 2026-06-23
fix(uninstall): detect Anki support files

chaochaoweb3:
## Summary
- detect Anki's exact `~/Library/Application Support/AnkiProgramFiles` support directory during uninstall
- keep `~/Library/Application Support/Anki2` out of automatic uninstall cleanup because it contains decks, media, and backups
- cover both the collected support path and preserved profile path with a naming-variants regression test

Addresses #1145.

## Maintainer update
I narrowed the original patch before merge so Mole cleans the launcher-managed support files without deleting Anki profile data.

## Validation
- `bash -n lib/core/app_protection.sh`
- `git diff --check`
- `MOLE_TEST_NO_AUTH=1 bats tests/uninstall_naming_variants.bats tests/uninstall_safety.bats tests/bundle_resolver.bats`
- temporary-HOME smoke check for `find_app_files "net.ankiweb.anki" "Anki"` returning only `AnkiProgramFiles`

### DIFF
--- a/lib/core/app_protection.sh
+++ b/lib/core/app_protection.sh
@@ -1116,9 +1116,9 @@ find_app_files() {
         [[ -d ~/.mobiledev ]] && files_to_clean+=("$HOME/.mobiledev")
     fi
 
-    # Anki stores profile data outside the app-name and bundle-id defaults.
+    # Anki's profile directory (Anki2) contains decks, media, and backups.
+    # Only collect the launcher-managed support files here.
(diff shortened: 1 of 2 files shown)


tw93:
@chaochaoweb3 thank you for the patch. I narrowed the cleanup before merging: Mole now detects `~/Library/Application Support/AnkiProgramFiles`, but intentionally leaves `~/Library/Application Support/Anki2` alone because that folder contains decks, media, and backups.

Merged into `main` in `8bc73a7` after running the related uninstall safety tests.

PR #1154, opened 2026-06-24
fix(clean): never delete endpoint-security agent caches

ImL1s:
## Summary

`mo clean` and `mo analyze` deleted rebuildable GPU/Metal shader caches, code-signature clones, and old user-owned files under every app's per-user Darwin folder (`/private/var/folders/.../C`, `.../X`, `.../T`). On a Mac managed with CrowdStrike Falcon this removed files inside the Falcon sensor's own cache container (`com.crowdstrike.falcon.App/com.apple.metalfe`), which tripped the sensor's tamper protection (IOA `MacFalconSensorTamper`, MITRE T1562.001) and was flagged as malware by the company's security team.

The fix follows the repo's data/logic split and protects every deletion path:

- `app_protection_data.sh`: new `ENDPOINT_SECURITY_BUNDLE_PREFIXES` array of endpoint-security / EDR / MDM vendor bundle ids (CrowdStrike, SentinelOne, ESET, Jamf, GlobalProtect, Cisco Secure Client), mirroring the vendors already in `OFFICIAL_UNINSTALLER_RULES`.
- `app_protection.sh`: `is_endpoint_security_cache_path()` iterates the array and matches a vendor id under `/private/var/folders` (or the `/var/folders` symlink form).
- **`file_ops.sh`: `validate_path_for_deletion()` rejects these paths before its `/private/var/folders` allowlist** — so every `safe_remove` / `safe_sudo_remove` caller is covered, not only the cleanup sweeps.
- `lib/clean/system.sh` (GPU-cache + `code_sign_clone` sweeps), `lib/clean/user.sh` (`_clean_darwin_user_runtime_dir`), and `should_protect_path()` also pre-check the predicate.
- **`cmd/analyze/delete.go`: `mo analyze` refuses to move these paths to Trash** (`validateTrashTarget` / `isProtectedAnalyzeDeletePath`), mirroring the shell predicate.

The match is protection-only: rebuildable caches for all other apps are still reclaimed.

## Safety Review

- Affects `clean` (system + user cleanup) and `analyze` (interactive Trash) deletion behavior.
- Affects protected-directory policy: it **adds** protection (skips deletions) for EDR/MDM agent paths under `var/folders`, enforced centrally in `validate_path_for_deletion()` so no future caller can regress. It does not broaden any deletion matcher, remove an existing protection, or change sudo / Trash / dry-run / operation-logging behavior.
- The predicate is anchored to `/private/var/folders/*` and `/var/folders/*` (an unrelated `.../var/folders/...` path cannot match) and is protection-only: the worst case is leaving a few MB of a security agent's rebuildable cache unreclaimed; it can never cause an unwanted deletion. Non-security products (e.g. `com.cisco.webex`) are not matched.

## Tests

- `./scripts/check.sh` (shfmt, shellcheck, go vet, syntax) and `MOLE_TEST_NO_AUTH=1 ./scripts/test.sh` (full bats + Go): pass.
- `core_safe_functions.bats`: predicate matches every vendor across `C/` / `X/` / `T/`, rejects normal apps / `com.cisco.webex` / non-Darwin `.../var/folders/...` / non-`var/folders` paths; `validate_path_for_deletion` rejects the Falcon path but allows a normal Darwin cache shard.
- `clean_system_maintenance.bats`: GPU-cache and `code_sign_clone` sweeps skip EDR paths while reclaiming a normal app's cache.
- `clean_user_core.bats`: `_clean_darwin_user_runtime_dir` skips an EDR agent's user-owned cache while reclaiming a normal old cache file.
- `cmd/analyze/delete_test.go`: `validateTrashTarget` rejects EDR Darwin caches and allows a normal Darwin cache.

## Safety-related changes

- Adds `ENDPOINT_SECURITY_BUNDLE_PREFIXES` (data) + `is_endpoint_security_cache_path()` (shell) + a Go mirror, enforced at `validate_path_for_deletion()` (covers all shell deletion callers), the three `clean` sweeps, `should_protect_path()`, and `mo analyze`'s Trash validation. Protection-only: no new deletion surface, no change to sudo / Trash / dry-run / logging.

### DIFF
--- a/cmd/analyze/delete.go
+++ b/cmd/analyze/delete.go
@@ -169,12 +169,24 @@ func validateTrashTarget(path string) error {
 }
 
 func isProtectedAnalyzeDeletePath(path string) bool {
-	home := os.Getenv("HOME")
-	if home == "" || path == "" {
+	if path == "" {
 		return false
 	}
 
 	cleanPath := filepath.Clean(path)
+
+	// EDR / Darwin-cache protection is based on the absolute path and does not
+	// depend on HOME, so check it first: an unset HOME must not let a Falcon
+	// cache slip through (e.g. `env -u HOME mo analyze`).
+	if isEndpointSecurityCachePath(cleanPath) {
+		return true
+	}
+
+	home := os.Getenv("HOME")
+	if home == "" {
+		return false
+	}
+
 	orbstackState := filepath.Join(home, ".orbstack")
 	if cleanPath == orbstackState || strings.HasPrefix(cleanPath, orbstackState+string(filepath.Separator)) {
 		return true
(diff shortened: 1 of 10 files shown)


tw93:
@ImL1s thank you for the detailed fix and context. I added one test update before merging:

- tightened the safe_sudo_remove protected-path return-code test to use the real CrowdStrike var/folders cache shape
- added a Go test to keep mo analyze's endpoint-security prefix list in sync with ENDPOINT_SECURITY_BUNDLE_PREFIXES from the shell data

Merged to main in 1eb04fe. This will be included in the next release; users can try the main build via mo update --nightly after Nightly refreshes.

Issue #1218, opened 2026-07-13
[BUG] Uninstall can be suspended with SIGTTIN after removal completes


Carsonn-Liu:
## Describe the bug

After an application is removed successfully with `mole uninstall`, Mole can be suspended when the final return/exit prompt tries to read from the terminal:

```text
Uninstall complete
Removed 1 app, freed 16.6MB: Adobe Downloader

Press Enter to return to the app list, press q to exit (5) [1]  + 9008 suspended (tty input)  mole uninstall
```

The command was run normally in the foreground, not with `&`. Running `fg` resumes it, confirming the process is stopped by `SIGTTIN` rather than crashed.

## Steps to reproduce

1. Run `mole uninstall` in a foreground zsh session.
2. Select and remove an application.
3. Wait for the completed-uninstall screen and its return/exit countdown.
4. Observe zsh report `suspended (tty input)` when Mole attempts to read the next key.

This was observed while uninstalling Adobe Downloader. Because the app was already removed, I did not repeat a destructive uninstall solely to reproduce the final prompt.

## Expected behavior

Once removal has completed, Mole should keep the interactive prompt working when it still owns the foreground terminal, or exit cleanly if terminal ownership has returned to the parent shell. It should not leave a suspended job after a successful uninstall.

## Debug logs

The relevant terminal output is included above. Command resolution was checked to rule out aliases and wrapper functions:

```text
mole is /opt/homebrew/bin/mole
mo is /opt/homebrew/bin/mo
TERM_PROGRAM=ghostty
```

## Environment

```text
Mole:          1.46.0
Install:       Homebrew
macOS:         26.5.2
Kernel:        25.5.0
Architecture:  arm64
SIP:           Enabled
Shell:         /bin/zsh
Terminal:      Ghostty
Disk free:     396.32 GB
```

## Additional context

The application and user-level data had already been removed before the suspension. Two privileged helper remnants were displayed as review-only; they are unrelated to the terminal-state issue.

A fix and regression tests are already proposed in PR #1213.

wbpluto:
This issue has reappeared here with me：


Mole version 1.46.0
macOS: 26.5.2
Architecture: arm64
Kernel: 25.5.0
SIP: Enabled
Disk Free: 50.79GB
Install: Homebrew
Shell: /bin/zsh


<img width="612" height="311" alt="Image" src="https://github.com/user-attachments/assets/c25e5ff3-86a1-4bce-b820-0dbf2bfcb294" />

[...]

PR #1241, opened 2026-07-16
fix(uninstall): avoid SIGTTIN at confirm prompt from background tty handoff (#1222)

### LINKED ISSUE #1222: [BUG] Uninstall suspended with SIGTTIN at the confirmation prompt: background timeout workers steal the tty foreground pgrp
## Describe the bug

`mo uninstall <app>` is suspended with `SIGTTIN` at the **pre-uninstall confirmation prompt**, before anything is removed:

```text
➜ mo uninstall ipaverse
◎ Matched 1 app(s):
1. ipaverse  6.5MB  |  Last: 3w ago

Proceed with uninstallation? [y/N] zsh: suspended (tty input)  mo uninstall ipaverse
```

The command was run normally in the foreground, not with `&`.

This is **related to but distinct from #1218 / PR #1213**. #1218 fires at the *post-removal* "Press Enter to return" countdown, after the app is gone. This one fires at the *confirmation* `read` in `bin/uninstall.sh:1519`, so nothing is ever uninstalled. More importantly, I believe I've found the underlying source of the terminal theft, which PR #1213 does not address (details at the bottom).

## Root cause

Mole hands the controlling terminal's foreground process group away **from background subshells**, and then reads from the tty in the foreground script.

1. On a machine with **no `gtimeout`/`timeout`** (i.e. no coreutils), `run_with_timeout` takes the Perl fallback in `lib/core/timeout.sh:155-243`.
2. That helper deliberately runs the timed child in a new process group and hands it the terminal via `tcsetpgrp` — the #1201 fix, so nested `sudo` can prompt.
3. But `bin/uninstall.sh` calls `run_with_timeout` from **background** subshells that have no need for the tty at all:
   - the disowned metadata-refresh job (`bin/uninstall.sh:293`/`:348`), which fans out up to 4 parallel workers, each running `run_with_timeout 0.2 mdls ...` (`:279`);
   - the parallel scan workers (`:711` → `:77`, `:96`, `:113-117`).

   These blocks redirect stdout/stderr (`) > /dev/null 2>&1 &`) but **not stdin**, so stdin is still the tty, `-t STDIN` is true in the Perl helper, and every worker performs the `tcsetpgrp` handoff.
4. While the main script sits at `read -r confirm` (`bin/uninstall.sh:1519`), one of those background helpers takes the terminal's foreground group. The main script is now a background process reading the tty → `SIGTTIN` → zsh reports `suspended (tty input)`.

The refresh job is also **disowned**, so it keeps doing this after the main script has moved on.

## Evidence

I instrumented this with a pty harness that emulates zsh's job control exactly (session leader + job in its own pgrp made foreground) and sampled `tcgetpgrp()` on the pty throughout the run.

**Cold metadata cache** (`~/.cache/mole/uninstall_app_metadata_v1` removed), which is what makes the refresh job run:

- The terminal's foreground process group is reassigned **60+ times in ~8 seconds**, bouncing between mole's pgrp and short-lived helper pgrps, and repeatedly landing on process groups that are **already dead**.
- At the moment the confirm prompt was printed, the foreground pgrp was a helper's, not mole's.
- The suspend reproduced **2/2**: job stopped by signal 21 (`SIGTTIN`), with mole's job pgrp ≠ the tty's foreground pgrp.

**Control — same harness, same cold cache, only the Perl timeout path disabled** (`MO_TIMEOUT_INITIALIZED=1 MO_TIMEOUT_BIN= MO_TIMEOUT_PERL_BIN=`, which falls back to the shell timeout implementation that never calls `tcsetpgrp`):

- The terminal is never stolen; the foreground pgrp at the prompt is the job's own pgrp.
- **2/2 clean runs**, prompt answered normally.

That isolates the cause to the Perl helper's tty handoff being invoked from background subshells.

This also explains the intermittency: the guilty refresh job only fires when the metadata cache is cold or stale. With a warm cache the same command frequently succeeds.

## Steps to reproduce

On a machine **without coreutils** (so `command -v gtimeout timeout` finds nothing):

```sh
rm -f ~/.cache/mole/uninstall_app_metadata_v1        # force the background refresh job
mo uninstall <some-app>                              # in a foreground zsh job
# -> suspended (tty input) at "Proceed with uninstallation? [y/N]"
```

Installing coreutils makes it disappear, because `MO_TIMEOUT_BIN=gtimeout` is used and the Perl handoff never runs.

## Suggested fix

The background call sites don't need the terminal — they run `mdls` and `du`. Redirecting their stdin would make the Perl helper's `-t STDIN` check false and skip the handoff entirely, while preserving the #1201 foreground-sudo behaviour:

```diff
-    ) > /dev/null 2>&1 &
+    ) > /dev/null 2>&1 < /dev/null &
```

(and likewise for the parallel metadata workers)

Alternatively, gate the handoff behind an explicit opt-in that only the interactive, sudo-capable call sites set, rather than doing it for every timed command that happens to inherit a tty on stdin.

## Why PR #1213 doesn't cover this

PR #1213 (for #1218) adds `mole_tty_is_foreground` before the **post-uninstall countdown** read, and adds a validated retry around `tcsetpgrp`. Neither helps here:

- The confirm `read` at `bin/uninstall.sh:1519` has no such guard, and "exit cleanly if not foreground" wouldn't be the right behaviour for a confirmation prompt anyway.
- The retry doesn't help because the handoff isn't *failing* — it *succeeds*. The bug is that a background worker performs it at all while the foreground script is about to read.

So even with #1213 merged, I'd expect this confirm-prompt suspension to remain on machines without coreutils.

## Environment

```text
Mole:          1.46.0
Install:       Homebrew
macOS:         26.5.2
Kernel:        25.5.0
Architecture:  arm64
Shell:         /bin/zsh
coreutils:     NOT installed (no gtimeout/timeout -> Perl timeout fallback active)
```

Happy to share the pty harness script if it's useful for a regression test — it deterministically reproduces the `SIGTTIN` stop on a cold cache.

sownhere:
## Summary

Fixes #1222. `mo uninstall <app>` was suspended with `SIGTTIN` at the **pre-uninstall confirmation prompt** — before anything was removed — on machines **without coreutils** (`gtimeout`/`timeout` absent, so the Perl timeout fallback is active).

## Root cause

The Perl timeout fallback in `lib/core/timeout.sh` deliberately hands the controlling terminal's foreground process group to its timed child whenever stdin is a tty, so nested `sudo` inside brew cask uninstall scripts can prompt (the #1201 fix). But `bin/uninstall.sh` calls `run_with_timeout` from **background** workers that never need the terminal:

- the disowned metadata-refresh subshell and its nested per-app workers (`mdls` / `du`);
- the parallel scan workers (`process_app_metadata` → `mdls`).

Those blocks redirected stdout/stderr but not **stdin**, so when a worker still had the tty on stdin, the Perl helper's `-t STDIN` check was true and it performed the `tcsetpgrp` handoff — stealing the foreground process group. While the main script sat at the `read` confirmation prompt, it was now a background process reading the tty → `SIGTTIN` → `zsh: suspended (tty input)`.

This is distinct from #1218 / PR #1213 (which guards the *post*-removal countdown): the confirm-prompt `read` has no such guard, and the handoff here isn't *failing* — it *succeeds* from a context that should never touch the terminal.

Full credit to @oichtental for the precise diagnosis and the suggested fix.

## Fix

Redirect stdin from `/dev/null` on the two background call sites so `-t STDIN` is false and the handoff is skipped:

- the disowned metadata-refresh subshell (its nested workers inherit the redirected stdin);
- the parallel scan workers.

The interactive sudo handoff (#1201) lives on **non-background** call sites (`lib/uninstall/brew.sh`) that keep the tty on stdin and are untouched, so it is unaffected.

## Tests

New coverage in `tests/core_timeout.bats` (with an `expect` pty harness), verifying both halves of the contract:

- the Perl fallback **still** hands the tty to its child when stdin is a tty (#1201 preserved);
- the Perl fallback **never** hands off when stdin is `/dev/null` (#1222 fixed);
- a source guard that both background call sites in `bin/uninstall.sh` redirect stdin.

The source guard was confirmed to fail on the pre-fix tree and pass after.

## Verification

```
MOLE_TEST_NO_AUTH=1 bats tests/core_timeout.bats           # 27 passed
MOLE_TEST_NO_AUTH=1 bats tests/uninstall.bats tests/uninstall_scan_bash32.bats  # 82 passed
./scripts/check.sh --format                                 # clean (shfmt + shellcheck)
```

### DIFF
--- a/bin/uninstall.sh
+++ b/bin/uninstall.sh
@@ -345,7 +345,14 @@ start_uninstall_metadata_refresh() {
         uninstall_release_metadata_lock "$MOLE_UNINSTALL_META_CACHE_LOCK"
         rm -f "$updates_file" "$refresh_merged_file"
         rm -f "$refresh_file" 2> /dev/null || true
-    ) > /dev/null 2>&1 &
+        # Redirect stdin from /dev/null so the perl timeout fallback does not see
+        # a tty on stdin and hand the controlling terminal to its timed child.
+        # This background refresh (and its nested workers, which inherit this
+        # stdin) never needs the terminal; leaving stdin as the tty lets a worker
+        # steal the foreground process group and stop the foreground prompt with
+        # SIGTTIN (issue #1222). The interactive sudo handoff (#1201) is on
+        # non-background call sites and is unaffected.
+    ) > /dev/null 2>&1 < /dev/null &
     disown "$!" 2> /dev/null || true
 
 }
@@ -708,7 +715,10 @@ _scan_resolve_uncached() {
     if ((total_apps > 0)); then
         for app_data_tuple in "${app_data_tuples[@]}"; do
             ((app_count++))
-            process_app_metadata "$app_data_tuple" "$scan_raw_file" &
+            # Redirect stdin from /dev/null so the perl timeout fallback used by
+            # process_app_metadata does not hand the controlling terminal to its
+            # timed mdls/du child from this background worker (issue #1222).
+            process_app_metadata "$app_data_tuple" "$scan_raw_file" < /dev/null &
             pids+=($!)
             update_scan_status "Scanning applications..." "$app_count" "$total_apps"
 
--- a/tests/core_timeout.bats
+++ b/tests/core_timeout.bats
@@ -260,6 +260,67 @@ setup() {
 	[[ "$output" == *"READ-AFTER:typed-after"* ]]
 }
 
+# Issue #1222: the perl fallback hands the controlling terminal to its timed
+# child whenever stdin is a tty (the #1201 behaviour). When it is invoked from a
+# background metadata/scan worker in bin/uninstall.sh that still has the tty on
+# stdin, that handoff steals the foreground process group from the foreground
+# script, which then stops with SIGTTIN at the confirmation prompt. Redirecting
+# the worker's stdin from /dev/null makes -t STDIN false and skips the handoff.
+# These two tests pin both halves of the contract: the handoff still happens for
+# interactive (tty) callers, and never happens once stdin is /dev/null.
+_tty_bg_field() {
+	# Extract NAME=<digits> from the fixture output (single line each).
+	printf '%s\n' "$2" | sed -n "s/.*${1}=\\([0-9][0-9]*\\).*/\\1/p" | head -1
+}
+
+@test "run_with_timeout: perl fallback hands tty to child when stdin is a tty (#1201/#1222)" {
+	if [[ "$(uname -s)" != "Darwin" || ! -x /usr/bin/expect || ! -x /usr/bin/perl ]]; then
+		skip "macOS expect/perl required"
+	fi
+
+	run /usr/bin/expect "$PROJECT_ROOT/tests/timeout_tty_background.exp" "$PROJECT_ROOT" tty
+
+	[ "$status" -eq 0 ] || return 1
+	local child fg caller
+	child=$(_tty_bg_field CHILD_PGRP "$output")
+	fg=$(_tty_bg_field FG "$output")
+	caller=$(_tty_bg_field CALLER_PGRP "$output")
+	[[ -n "$child" && -n "$fg" && -n "$caller" ]] || return 1
+	# The timed child captured the terminal's foreground process group.
+	[ "$fg" = "$child" ] || return 1
+	[ "$fg" != "$caller" ] || return 1
+}
+
+@test "run_with_timeout: perl fallback keeps tty with caller when stdin is /dev/null (#1222)" {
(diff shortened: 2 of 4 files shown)


oichtental:
Thanks for the quick turnaround, and for the credit — much appreciated.

The fix and the test split look right to me. I especially like that the bats coverage pins both directions of the contract rather than just the bug: asserting the Perl fallback *still* hands off when stdin is a tty is what keeps a future cleanup from silently re-breaking #1201 while "fixing" something else. The source guard on the two background call sites is a nice touch too — it'll catch a new background worker that forgets the redirect.

LGTM.

Issue #88, opened 2025-11-28
Feature request: Application Support, Group containers cache folders


mkhalila:
Hey there, 

Firstly once again what an amazing tool, thank you so much. I've been running mole for a few weeks now.

When I run OmniDiskSweeper, I can see "Application Support" or "Group Containers" caches/logs for many apps that mole doesn’t clear up (maybe by design?). These take up significant space. Examples include:

Comet browser: 
- 160mb: '/Users/khalil/Library/Application Support/Comet/Default/Service Worker/CacheStorage'
- 12mb: '/Users/khalil/Library/Application Support/Comet/Default/Service Worker/ScriptCache'
- 20mb: '/Users/khalil/Library/Application Support/Comet/Crashpad/completed'

Lark: 
- 550mb stale update from 2 weeks ago: '/Users/khalil/Library/Application Support/LarkInternational/update/update_downloading'
- 463mb cache: '/Users/khalil/Library/Application Support/LarkInternational/aha/users/0f416d.../profile_explorer/Service Worker/CacheStorage'

Zed editor:
- 126mb cache: '/Users/khalil/Library/Application Support/Zed/node/node-v24.11.0-darwin-arm64/cache'

Asana:
- 498mb cache: '/Users/khalil/Library/Application Support/Asana/Cache/Cache_Data'
- 165mb: '/Users/khalil/Library/Application Support/Asana/Code Cache/js'
- 75mb: '/Users/khalil/Library/Application Support/Asana/Service Worker/CacheStorage'

Whatsapp:
- 33mb logs: '/Users/khalil/Library/Group Containers/group.net.whatsapp.WhatsApp.shared/Logs'

I appreciate some of these tools may not be as popular as others but maybe patterns like `Application Support/...Service Worker/CacheStorage` or `CacheStorage`, or `Cache_Data`, `Group Containers/.../Logs` etc., could be extrapolated

tw93:
Thank you very much for your suggestion, I didn't consider this, it's a great suggestion, I will add it in the next version.

tw93:
It has been updated to 1.10.17. What you said is supported. Thank you for your hard work. Mo update to update to the latest version, and then use it. I wish you a happy use.

mkhalila:
Thank you so much. That was so swift. Happy to help!

PR #1246, opened 2026-07-18
fix(clean): cover Zed bundled npm cache

hhh2210:
## Summary

Zed now keeps its bundled npm cache at `~/Library/Application Support/Zed/node/cache`. Mole only cleaned `~/Library/Caches/Zed`, so the npm cache was left behind. On my Mac, the current path had grown to 5.0 GB and contained the usual npm `_cacache` and `_npx` directories.

This change cleans only the contents of `node/cache` through `safe_clean`. It does not touch Zed's `db` directory or other editor state. The existing Zed test now checks the exact cache path and verifies that the `db` sibling stays outside the cleanup target.

Refs #88, which reported an earlier versioned form of the same Zed npm cache path.

## Safety Review

- This affects `mo clean` by adding one exact, user-owned cache path.
- Deletion still goes through `safe_clean`, so dry-run, whitelist, path validation, and operation logging keep their existing behavior.
- No path validation, protected-directory, symlink, sudo, install, or release boundaries change.
- This PR does not change System Data detection or clean any other `Application Support` data.

## Tests

- `MOLE_TEST_NO_AUTH=1 bats --filter 'clean_code_editors includes Zed caches' tests/clean_app_caches.bats`
- `git diff --check`
- Full `tests/clean_app_caches.bats`: 41 passed, 1 existing Xcode assertion failed. The same failure reproduces on an unmodified `upstream/main` worktree.

## Safety-related changes

- Adds `~/Library/Application Support/Zed/node/cache/*` to `clean_code_editors()`.
- Keeps `~/Library/Application Support/Zed/db` untouched and covered by a regression assertion.

### DIFF
--- a/lib/clean/app_caches.sh
+++ b/lib/clean/app_caches.sh
@@ -171,8 +171,10 @@ clean_code_editors() {
     safe_clean ~/Library/Application\ Support/Code/WebStorage/*/CacheStorage/* "VS Code webview cache"
     safe_clean ~/Library/Caches/com.sublimetext.*/* "Sublime Text cache"
     safe_clean ~/Library/Caches/Zed/* "Zed cache"
-    # Zed stores its bundled npm cache here; keep editor state under db/ untouched.
+    # Zed npm caches: node/cache (system-node scratch) and node/node-v*/cache
+    # (per-version managed runtime, see #88); keep editor state under db/ untouched.
     safe_clean ~/Library/Application\ Support/Zed/node/cache/* "Zed npm cache"
+    safe_clean ~/Library/Application\ Support/Zed/node/node-v*/cache/* "Zed npm cache"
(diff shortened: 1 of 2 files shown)


tw93:
@hhh2210 Thanks, this is a clean addition and the db/ guard in the test is appreciated.

I pushed one commit on top of your branch: Zed's managed (bundled) node runtime keeps its npm cache under `node/node-v{version}-{os}-{arch}/cache`, which is the exact path reported in #88, while `node/cache` covers the system-node scratch cache. Both variants are now cleaned, with a test asserting the versioned cache is removed and the runtime `bin/` stays untouched.

Full `tests/clean_app_caches.bats` passes 42/42 here, including the Xcode case you saw failing locally.

PR #1254, opened 2026-07-20
feat(analyze): add opt-in --deep scan for root-owned System Data (/private/var/folders)

### LINKED ISSUE #1253: analyze: overview can't see root-owned System Data (idleassetsd temp filling /private/var/folders)
## Problem

`mo analyze`'s overview runs completely unprivileged, so it cannot measure
root-owned trees. In practice this hides the single largest source of "disk full
but nothing to delete" reports on modern macOS: **`/private/var/folders`**.

A background system daemon — most commonly `com.apple.idleassetsd` (the Aerial
screensaver / dynamic-wallpaper downloader) — can get stuck (re)downloading 4K
video assets and leave behind hundreds of aborted temp files:

```
/private/var/folders/zz/…/T/com.apple.idleassetsd/CFNetworkDownload_*.tmp
```

Each is ~1 GB. On one affected machine (macOS 15.7.1) this reached **~221 GB**.
Because the folder is owned by `root` (mode `700`) and lives in the per-user
Darwin temp area, macOS reports it only as opaque **"System Data"**, the Finder
never shows it, and — critically — `mo analyze` shows almost nothing for it: a
plain `du` run as the user hits `permission denied` on the root subtree and
reports ~0. That is actively misleading when hunting for what is filling a disk.

## Why analyze misses it

1. `/private` is not part of the overview at all (only Home, User Library,
   Applications, System Library + `$HOME`-based insights).
2. Even if it were, `analyze` is intentionally sudo-free, so `du` as the user
   cannot read the root-owned tree and undercounts it to ~0.

## Proposal

An **opt-in** `--deep` flag (also `MO_ANALYZE_DEEP=1`) that:

- Surfaces `/private/var/folders` and `/private/var/vm` as hidden-space insights.
- Elevates `du` reads of `/private` paths with `sudo -n`.
- Primes `sudo` **once, up front, before the TUI starts**, so the interface
  never blocks on a password prompt. Declining auth falls back to the normal
  unprivileged view.

This keeps the default behavior byte-for-byte unchanged and the sudo tension
contained to an explicit, documented opt-in, while finally making the biggest
chunk of "System Data" visible inside the existing disk explorer.

## Environment

- macOS 15.7.1 (24E-class), Apple Silicon
- Real-world case: `com.apple.idleassetsd` temp accumulation, ~221 GB

I have a working implementation and will open a PR referencing this issue.

italojs:
Closes #1253

## Summary
`mo analyze`'s overview runs fully unprivileged, so it cannot measure root-owned
trees. The biggest offender is `/private/var/folders`, where background daemons —
notably `com.apple.idleassetsd` (Aerial screensaver / dynamic wallpaper) — leak
hundreds of GB of aborted `CFNetworkDownload_*.tmp` files. macOS buries this in
the opaque "System Data" bucket and the Finder never shows it, so users see a
full disk with nothing to delete. A plain `du` as the user hits "permission
denied" on the root folder and reports ~0 — actively misleading.

## What this adds
An **opt-in** `--deep` flag (also `MO_ANALYZE_DEEP=1`) that:
- Surfaces `/private/var/folders` and `/private/var/vm` as hidden-space insights.
- Elevates `du` reads of `/private` paths with `sudo -n`.
- Primes sudo **once, up front, before the TUI starts**, so the interface never
  blocks on a password prompt. Declining auth falls back to the normal view.

## Safety
- Opt-in only; default behavior is byte-for-byte unchanged.
- Read-only: no deletion paths changed. Root-owned files remain non-deletable
  (analyze never deletes with sudo).
- `sudo` is used only for `/private` targets and only in deep mode. User paths
  (Home, ~/Library, /Applications) are never elevated.
- The `MOLE_TEST_NO_AUTH` guard suppresses sudo, so tests/CI never prompt.

## Product-decision-filter
1. Belongs to `analyze` (disk explorer). ✔
2. Safe by default, opt-in, testable without real auth, explainable in one line. ✔
3. Read-only — nothing is changed, only measured. ✔
4. N/A (visualization only; no removal). ✔
5. Not better as docs/warning — the value is making invisible System Data
   visible inside the existing explorer. ✔

## Known limitation
Overview sizing and folded-dir sizing of root-owned trees go through `du` and are
elevated. A pure Go-walk fallback still can't expand a `700`/root subtree
file-by-file, so some deep subtrees show a correct total but limited drill-down.
Flagged as a follow-up.

## Tests
- New `cmd/analyze/deep_scan_test.go` (path gating, sudo guards, argv building,
  deep-only insight gating). The unit tests never invoke real sudo.
- `go test ./...` → all pass. `gofmt` clean.
- Manual: `mo analyze --deep --json` lists the new insights;
  without `--deep` they are absent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,7 +189,7 @@ Select a location to explore:
 ↑↓→ | Enter | R Refresh | O Open | P Preview | F File | Esc/Q Quit
 ```
 
-> Deep scan: `mo analyze --deep` additionally measures root-owned system areas such as `/private/var/folders`, where background daemons (for example `com.apple.idleassetsd` aerial-wallpaper downloads) can silently accumulate tens of gigabytes that macOS hides inside the opaque "System Data" bucket. It is opt-in because it authenticates with `sudo` once, up front, and then reads without ever blocking the interface on a password prompt; if authentication is declined, `analyze` falls back to the normal, unprivileged view.
+> Note: `mo analyze --deep` also measures root-owned areas such as `/private/var/folders`, which macOS hides inside "System Data". It asks for `sudo` once up front and falls back to the normal view if you decline. These areas are read-only in analyze: they are shown so you can see what is filling the disk, not deleted from here.
 
 ### Live System Status
 
--- a/README.md
+++ b/README.md
@@ -189,6 +189,8 @@ Select a location to explore:
 ↑↓→ | Enter | R Refresh | O Open | P Preview | F File | Esc/Q Quit
 ```
 
+> Deep scan: `mo analyze --deep` additionally measures root-owned system areas such as `/private/var/folders`, where background daemons (for example `com.apple.idleassetsd` aerial-wallpaper downloads) can silently accumulate tens of gigabytes that macOS hides inside the opaque "System Data" bucket. It is opt-in because it authenticates with `sudo` once, up front, and then reads without ever blocking the interface on a password prompt; if authentication is declined, `analyze` falls back to the normal, unprivileged view.
+
 ### Live System Status
 
 Real-time dashboard with health score, hardware info, and performance metrics.
--- a/cmd/analyze/deep_scan.go
+++ b/cmd/analyze/deep_scan.go
@@ -12,36 +12,43 @@ import (
 	"time"
 )
 
-// Deep scan surfaces root-owned system areas — most notably
-// /private/var/folders, where daemons such as com.apple.idleassetsd can
-// silently accumulate hundreds of gigabytes of aborted temporary downloads —
-// that the normal, unprivileged analyze overview cannot measure. macOS folds
-// these bytes into the opaque "System Data" bucket and the Finder never shows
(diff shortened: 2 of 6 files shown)


tw93:
@italojs Thanks for digging into this, your diagnosis in #1253 was spot on: a plain `du` reporting ~0 on a root-owned tree is worse than reporting nothing, and macOS hiding hundreds of GB of idleassetsd temp files in "System Data" is exactly the "disk full but nothing to delete" trap.

After sitting with it, I decided to keep `analyze` fully unprivileged instead of teaching it to `sudo`, so the opt-in `--deep` path has been removed. It only added visibility (analyze never deletes with `sudo`), and I'd rather `analyze` stay a quiet disk explorer than grow a privileged mode.

The problem you found is still fixed, just from the other end: `mo clean` now removes the stale `CFNetworkDownload_*.tmp` files under the idleassetsd temp directory directly, so the space is reclaimed rather than only shown. It's on `main` (nightly) now. Appreciate you surfacing it.

PR #1263, opened 2026-07-22
feat(status): let the CPU card show a configurable number of cores

gloiiire:
## What

The CPU card hard-codes `min(len(cores), 2)`, so `mo status` only ever lists the
two busiest cores regardless of how many the machine has. This makes the count
configurable:

- Press **`c`** to cycle how many cores the card lists: **2 → 4 → 8 → all**.
- The choice is persisted in `~/.config/mole/status_prefs`, so it survives restarts.
- The default stays **2**, so existing users see no change.
- Per-core data was already collected — this only changes rendering.

## Two commits, and why

1. **`refactor(status): multi-key preferences store`** — the preferences file was
   read with a single whole-file string comparison (`cat_hidden=true`), so it could
   only ever hold one preference: writing a second key would clobber the first.
   This refactors it into a small `key=value` map store (`loadPrefs`/`savePref`)
   that preserves other keys on write and sorts output for a stable file. The
   existing cat toggle behaves exactly as before.
2. **`feat(status): let the CPU card show a configurable number of cores`** — adds
   the `cpu_cores` preference (0 = all) and the `c` key on top of that store.

Splitting it this way keeps the store refactor reusable for any future preference.

## Testing

- `./scripts/check.sh` passes (gofmt + go vet + syntax).
- `go test ./cmd/status/` green, including new tests for: prefs round-trip &
  key-preservation, the core-count rendering (top-N / all / clamp), and the `c`
  cycle wrap-around.

No new dependencies; the macOS/Linux build is unaffected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,7 +218,7 @@ Proxy   HTTP · 192.168.1.100             Terminal   ▮▯▯▯▯  12.5%
 
 Health score is based on CPU, memory, disk, temperature, and I/O load, with color-coded ranges.
 
-Shortcuts: In `mo status`, press `k` to toggle the cat and save the preference, and `q` to quit.
+Shortcuts: In `mo status`, press `k` to toggle the cat, `c` to cycle how many CPU cores the card lists (2, 4, 8, all), and `q` to quit. Both preferences are saved.
 
 When enabled, `mo status` shows a read-only alert banner for processes that stay above the configured CPU threshold for a sustained window. Use `--proc-cpu-threshold`, `--proc-cpu-window`, or `--proc-cpu-alerts=false` to tune or disable it.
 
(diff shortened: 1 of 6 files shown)


tw93:
Merged, thanks. The multi-key prefs store is a clean refactor and the per-core data was already there, so this was a small rendering change well contained.

Two things I adjusted before merging: on a many-core Mac "all" grew the card past a short window, and the status frame has no scrollback (`padViewToHeight` only pads), so the lower cards could scroll off. The view now steps the core count back down until the frame fits, without touching the saved preference. I also added `c` to the shortcut line in the README next to `k`, and switched one `strings.Split` to `strings.SplitSeq` to satisfy the `modernize` linter.

PR #1270, opened 2026-07-23
spike: add machine-readable CLI previews for agent workflows

philippb:
## Feedback requested

This is intentionally a draft spike. Let me know if this is the right direction and, especially, whether the proposed machine contract is small enough for Mole.
This is coming from a personal need to having agents clean up local dev environment with build artifacts and other generated data

## What was the problem

Mole already has good read-only JSON surfaces for `analyze`, `status`, and `history`, but agent consumers still hit two practical gaps:

- `analyze --json` always returns the complete indented result, even when an agent only needs the largest few entries.
- destructive commands have safe dry-run behavior, but their preview formats are human-oriented and inconsistent. An agent must scrape terminal prose or, for `clean`, read a separate path-list file.

That makes automated use more verbose and less reliable than the existing safety model deserves.

## Why this is important

Coding agents can generate large, rebuildable artifacts during long sessions. They need to inspect disk usage and preview cleanup without consuming unnecessary context or accidentally turning a formatting option into an authorization path.

A small structured preview contract makes the safety state explicit: whether the command is a dry-run, which scope was scanned, whether candidates are permanent or rebuildable, whether warnings were raised, and whether the result was truncated.

## What is the solution

This spike extends existing commands rather than adding an `mo agent` subsystem:

- `mo analyze --json --limit N [path]` returns the largest `N` top-level entries while preserving complete scan totals and explicit truncation metadata.
- `mo analyze --json --compact [path]` removes optional indentation without changing the decoded document.
- `mo clean --dry-run --json` and `mo purge --dry-run --json` emit a versioned preview receipt on stdout while keeping human diagnostics on stderr.
- Preview receipts include command, mode, scope, summary, warnings, candidates, action, safety classification, and command-specific recoverability.
- `--json` is rejected unless `--dry-run` is also present. Receipts explicitly set `apply_supported: false`.
- Machine candidate records use an internal NUL-delimited stream so unusual whitespace or human preview comment delimiters cannot corrupt paths.
- Shell completions, focused tests, and a durable spike log are included.

The human TUI and existing human-readable output remain unchanged.

## The things we tried and why this is a good initial approach

The spike was developed through five evidence-driven iterations:

1. Added top-N analysis output. On a representative large repository tree, `--limit 10` reduced the response from 14,969 to 5,362 bytes while preserving full totals.
2. Added opt-in compact encoding, reducing the bounded response to 4,215 bytes without changing the decoded JSON.
3. Proved the preview schema first on `clean`, which already had an authoritative dry-run candidate list.
4. Extended the same envelope to `purge`, while retaining its different `rebuild` recoverability semantics and configured-root validation.
5. Added structured safety warnings and deliberately rejected extra machinery that did not earn its complexity.

We considered bounded NDJSON progress streaming, but the observed scans completed in roughly half a second and a final bounded document was simpler. We also considered plan/run IDs, but rejected them because Mole has no durable plan store, status lookup, or apply receipt; an ID would be decorative and could imply authority it does not have.

This seems like a useful initial boundary because it is additive, preview-only, uses existing command concepts, and does not add storage, background state, a new subsystem, or an apply path. The full experiment record and rejected ideas are in `docs/spikes/agent-cli-protocol.md`.

## Safety review

- This changes cleanup and purge preview output only; it does not add or broaden a deletion path.
- JSON preview fails closed unless dry-run is active.
- Purge candidates are recorded only after configured-root safety revalidation.
- Clean and purge fixtures were verified to remain present after real CLI dry-runs.
- No path validation, protected-directory policy, sudo boundary, deletion funnel, or confirmation behavior was weakened.
- No real destructive Mole command was run during development or verification.

## Tests

- 232 touched-surface Bats tests across clean, app cleanup, project caches, purge, completion, and CLI behavior
- `MOLE_SKIP_FINDER_TESTS=1 MOLE_TEST_NO_AUTH=1 make verify`
  - ShellCheck
  - shell syntax
  - golangci-lint
  - all Go packages
- Fresh temporary-HOME `clean --dry-run --json` and `purge --dry-run --json` probes with fixture-preservation checks
- Fresh read-only `analyze --json --limit 10 --compact` probe against a representative large repository tree
- Negative controls proving JSON preview without dry-run exits nonzero without removing fixtures

Finder-dependent Trash integration tests used their documented skip flag because this spike does not change analyze deletion behavior.

## Feedback areas

I would especially appreciate feedback on:

- whether `--limit` and `--compact` fit the existing `analyze` command;
- whether the shared top-level preview fields are named appropriately;
- whether `clean` and `purge` are the right first two preview surfaces;
- whether this should be narrowed further before promotion; and
- whether any part of this belongs in documentation rather than the CLI.

### DIFF
--- a/bin/clean.sh
+++ b/bin/clean.sh
@@ -10,6 +10,7 @@ export LANG=C
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 source "$SCRIPT_DIR/../lib/core/common.sh"
+source "$SCRIPT_DIR/../lib/core/history.sh"
 
 source "$SCRIPT_DIR/../lib/core/sudo.sh"
 source "$SCRIPT_DIR/../lib/clean/brew.sh"
@@ -26,6 +27,8 @@ SYSTEM_CLEAN=false
 DRY_RUN=false
 PROTECT_FINDER_METADATA=false
 EXTERNAL_VOLUME_TARGET=""
+CLEAN_JSON=false
+CLEAN_JSON_CANDIDATE_FILE=""
 IS_M_SERIES=$([[ "$(uname -m)" == "arm64" ]] && echo "true" || echo "false")
 
 # Whitelist and preview belong to the invoking user even when the whole
@@ -1046,6 +1049,7 @@ safe_clean() {
                     else
                         echo "$display_path  # $size_human" >> "$EXPORT_LIST_FILE"
                     fi
+                    record_clean_json_candidate "$display_path" "$total_size"
                 done
             fi
         else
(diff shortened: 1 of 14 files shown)


tw93:
@philippb Thanks for asking about direction before building this out, and for the detailed write-up. I ran the branch locally against my own machine. I'm going to pass on this one, so here is the reasoning rather than a list of change requests.

The blocking issue is in the `clean` receipt. The paths it emits are not deletion targets, they are the grouped display paths from the human preview. `safe_clean` collapses siblings under their parent when a group has more than one child, and the text list keeps that visible as `, 20 items`. The JSON drops the count and emits the parent alone:

```json
{"path": "/Users/tw93/Library/Caches", "action": "remove", "size_bytes": 1844028416}
```

On my run, 37 of 66 candidate records were grouped parents like that, covering 2 to 20 real children each. `mo clean` never removes `~/Library/Caches` itself. For a human reading the list, the grouping is a readability win; for the consumer this receipt is aimed at, it overstates the deletion scope on more than half the records, and it overstates it most on exactly the paths where being wrong is worst. Candidate order is also nondeterministic, since the awk stage iterates an unordered map. The `purge` half is accurate, so the problem is specific to how `clean` builds that list.

On the direction itself: the CLI deliberately stays narrow, and `--limit` / `--compact` are both a `jq` call away on the consumer side, which puts them under the "prefer quieter defaults over more output modes" rule in `AGENTS.md`. A versioned preview envelope is also a compatibility promise I'd have to keep across releases, and I don't want to make that promise for a surface with no apply path behind it. The agent story I'd rather invest in is the skill in `.claude/skills/mole/`, which already documents `analyze --json`, `history --json` and the dry-run list; what's genuinely missing there is a stable, documented format for the dry-run list, not a second JSON contract.

There is also some layering I'd want undone regardless: `bin/clean.sh` sourcing all of `lib/core/history.sh` for one escaping helper, and `lib/clean/*.sh` probing back into `bin/clean.sh` with `command -v record_clean_json_candidate`, with `project.sh` using a third mechanism for the same job.

Closing as not planned, no reflection on the work, which was carefully done and well documented. If you want to keep pulling on this, the piece I'd actually look at is making the dry-run list itself unambiguous about groups, since that helps the human output and any machine consumer at the same time. Happy to discuss that in an issue first.

philippb:
Thanks for the thoughts and directions. I think a lot of the things you write make sense. The one thing I personally try to do differently on my personal CLI (most of them not as well thought through like mo) is to make it clear for the agent to understand without any skill. 

> The agent story I'd rather invest in is the skill in .claude/skills/mole/, which already documents analyze --json, history --json and the dry-run list; what's genuinely missing there is a stable, documented format for the dry-run list, not a second JSON contract.

I keep this in mind as I use it and next time I think of something. Generally speaking I love what you're doing and happy to share my thinking.

tw93:
That's a fair point, and honestly the stronger version of the idea. "Readable by an agent with no skill loaded" is a good bar to hold a CLI to, and it's a nudge for me to make the plain `--json` and dry-run surfaces self-explanatory rather than leaning on the skill to paper over gaps. The grouped-path ambiguity I ran into is a good example of that gap.

Thanks for taking the direction well, and for sharing how you think about it. Please do bring the next idea, this kind of input is welcome.

PR #1273, opened 2026-07-24
fix(status): collect processes under comma-decimal locales

### LINKED ISSUE #1267: [BUG] Processes card stuck on "Collecting..." under comma-decimal locales (e.g. ru_RU.UTF-8)
## Describe the bug

Under a locale where the decimal separator is a comma (e.g. `ru_RU.UTF-8`), the "Processes" card in `mo status` never populates and stays on "Collecting..." forever.

Root cause: `collectProcesses()` in `cmd/status/metrics_process.go` runs:
```
ps -Aceo pid=,ppid=,pcpu=,pmem=,rss=,comm= -r
```
without forcing `LC_ALL=C`/`LC_NUMERIC=C`. Under `ru_RU.UTF-8`, `ps` formats `%cpu`/`%mem` with a comma (e.g. `8,0` instead of `8.0`). `parseProcessOutput` then calls `strconv.ParseFloat(fields[2], 64)`, which fails on the comma-formatted value, so every line is skipped via `continue`, and the returned process slice is empty, triggering the permanent "Collecting..." placeholder.

## Steps to reproduce

1. Set the macOS system language/region so the locale is `ru_RU.UTF-8` (or any locale that uses a comma as decimal separator)
2. Run `mo status`
3. Processes card shows "Collecting..." indefinitely, no processes ever appear

## Expected behavior

Process list should populate and rank by CPU/memory regardless of system locale.

## Debug logs

```
$ ps -Aceo pid=,ppid=,pcpu=,pmem=,rss=,comm= -r | head -3
  666     1   8,0  0,6  97920 Terminal
73425 73411   7,9  2,1 358288 mimo
  433     1   5,3  0,6  95968 WindowServer
```

```
$ LC_ALL=C ps -Aceo pid=,ppid=,pcpu=,pmem=,rss=,comm= -r | head -3
94517 94500  21.7  2.8 463600 claude bg-spare
73425 73411  12.2  2.1 358336 mimo
  433     1   8.7  0.6  97984 WindowServer
```

## Workaround

Running `LC_ALL=C mo status` fixes it immediately, confirming the locale-dependent `ps` parsing is the cause.

## Environment

```
Mole version 1.47.1
macOS: 27.0
Architecture: arm64
Locale: LANG=ru_RU.UTF-8, LC_NUMERIC=ru_RU.UTF-8
```

## Additional context

Suggested fix: force `LC_ALL=C` (or at least `LC_NUMERIC=C`) in the environment when invoking `ps` inside `collectProcesses()`, so the numeric fields always use a period decimal separator regardless of the user's system locale.

tylergibbs1:
## Summary

- Run both process-collection `ps` paths with `LC_ALL=C` so CPU and memory percentages always use dot decimals.
- Add a regression test for the command environment.
- Fixes #1267

Prepared with Codex assistance and reviewed against the repository's safety rules.

## Safety Review

No. This only changes the environment of read-only `ps` collection. It does not affect cleanup, uninstall, optimize, installer, deletion, path validation, authorization, update, or install behavior.

## Tests

- `go test ./...`
- Manual locale check. `ru_RU.UTF-8` produced `112,8`; the C locale produced `112.8`.

## Safety-related changes

- None.

### DIFF
--- a/cmd/status/metrics.go
+++ b/cmd/status/metrics.go
@@ -3,8 +3,10 @@ package main
 import (
 	"context"
 	"fmt"
+	"os"
 	"os/exec"
 	"slices"
+	"strings"
 	"sync"
 	"time"
 
@@ -574,11 +576,29 @@ func (e snapshotEnrichment) apply(snapshot *MetricsSnapshot, preserveLiveProcess
 
 var runCmd = func(ctx context.Context, name string, args ...string) (string, error) {
 	cmd := exec.CommandContext(ctx, name, args...)
+	cmd.Env = cLocaleEnv()
 	output, err := cmd.Output()
 	if err != nil {
 		return "", err
 	}
 	return string(output), nil
+}
+
+// cLocaleEnv forces the C locale on every metric subprocess. ps and uptime
+// localize their decimal separator, so under ru_RU.UTF-8 they emit "8,0" and
+// every strconv.ParseFloat in the collectors fails (#1267). The shell commands
+// in bin/ already export LC_ALL=C; status-go is exec'd directly and did not.
+func cLocaleEnv() []string {
+	env := os.Environ()
+	filtered := make([]string, 0, len(env)+1)
+	for _, kv := range env {
+		key, _, found := strings.Cut(kv, "=")
+		if found && (key == "LC_ALL" || key == "LANG" || strings.HasPrefix(key, "LC_")) {
+			continue
+		}
+		filtered = append(filtered, kv)
+	}
+	return append(filtered, "LC_ALL=C")
 }
 
 var commandExists = func(name string) bool {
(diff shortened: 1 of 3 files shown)


tw93:
@tylergibbs1 Thanks, the diagnosis was exactly right. I reproduced it here: under `LC_ALL=ru_RU.UTF-8` the process card returned 0 entries on main and 5 with your fix.

I pushed one commit on top of your branch before merging. `collectProcesses` was not the only affected caller: `fallbackCPUUtilization` runs `ps -Aceo pcpu` and `fallbackLoadAvgFromUptime` runs `uptime`, and both parse the same localized decimals, so they silently returned zero under the same locale. Setting `cmd.Env` inside `runCmd` covers all three call sites at once and drops the extra `env` fork. It also lines up with the shell side, where every `bin/*.sh` already exports `LC_ALL=C` and only `status-go` did not, since it is exec'd directly.

I also reshaped the test. Asserting the command name is `env` locks in one implementation, so it now checks the subprocess environment and that `collectProcesses` returns results under a comma-decimal locale. Both fail without the fix.

Merged. The fix is on `main` now and available via `mo update --nightly`; it will be in the next stable release. Thanks for the clean report and the repro.

Issue #843, opened 2026-05-05
[FEATURE] Final Cut Pro Cache


irl231:
## Feature description

Native support within `mo clean` to identify and purge Final Cut Pro (`.fcpbundle`) generated render files and proxy media caches.

## Use case

Final Cut Pro stores heavy cache data (Render Files, Proxy Media, Optimized Media) directly inside its `.fcpbundle` package directories rather than standard macOS `~/Library/Caches/` paths. Standard `mo clean` execution misses these files entirely. FCP background rendering inflates library sizes by hundreds of gigabytes. Users currently have no CLI-native way in Mole to reclaim this storage and must manually clear caches inside the FCP application GUI per library.

## Proposed solution

Integrate a scanner into the `mo clean` routine or add a specific flag (e.g., `mo clean --fcp`).
1. Search user directories (e.g., `~/Movies`) for `.fcpbundle` packages.
2. Target the internal cache structures: `*.fcpbundle/**/Render Files/` and `*.fcpbundle/**/Transcoded Media/`.
3. Report the size of these internal cache directories during `mo analyze`.
4. Allow safe deletion of these folders via `mo clean`, as FCP automatically regenerates them if needed. 

## Alternatives considered

- Executing `find ~/Movies -name "Render Files" -type d -exec rm -rf {} +` manually outside of Mole.
- Paying for standalone GUI applications like Final Cut Library Manager.
- Opening every individual library in Final Cut Pro to trigger `File > Delete Generated Library Files`.

## Additional context

Final Cut Pro cache inflates rapidly during 4K video exports. On restricted hardware like a 14-inch MacBook Pro with an M2 Pro chip, local storage exhaustion occurs frequently due to stale `.fcpbundle` caches. 
Example target paths for deletion:
`~/Movies/Project.fcpbundle/Event/Render Files/High Quality Media/`
`~/Movies/Project.fcpbundle/Event/Transcoded Media/Proxy Media/`
These subdirectories are strictly cache and safe to delete without corrupting the parent FCP library structure.

### NOTE: The above is generated by Claude, but you will know the Idea about this
It can be prevent this, check the video: https://youtube.com/shorts/Prv71j-628g?si=m6Bh6SuZwQ6DWU1P
But, sometimes users forgot or doesn't know this settings, and I want the Final Cut Pro  to be available on `mo clean`

yuzeguitarist:
great idea

yuzeguitarist:
[...]

PR #1281, opened 2026-07-25
feat(clean): JianyingPro (CapCut CN) generated cache cleanup

### LINKED ISSUE #1277: [FEATURE] JianyingPro (CapCut CN) generated cache cleanup — 76 GB reclaimed on a real machine
## Feature description

Native support within `mo clean` to identify and purge JianyingPro's regenerable generated caches under `~/Movies/JianyingPro/User Data/Cache/` — the same problem shape as Final Cut Pro in #843.

## Use case

JianyingPro (剪映专业版, `com.lemon.lvpro`, ByteDance's desktop editor — the dominant editing tool among Chinese content creators, with CapCut as its international sibling) stores heavy generated caches in `~/Movies/JianyingPro/User Data/` rather than standard `~/Library/Caches/` paths. Standard `mo clean` misses these entirely, and `~/Movies` is (rightly) protected by Mole's conservative rules — exactly the situation that motivated #843 for `.fcpbundle` render files.

Field data from a real machine (MacBook Pro Intel, macOS 15.7, JianyingPro 11.1.0, 466GB disk, heavy subtitle-recognition user):

| Path | Size | Content |
|---|---|---|
| `User Data/Cache` (total) | **83 GB** | vs. only 37 GB of actual project drafts |
| `Cache/recognize` | **54 GB** | 881 uncompressed `.pcm`/`.m4a`/`.aac` files — raw audio extracted for auto-subtitle recognition, **never cleaned up after use** (oldest files 10+ months old; PCM ≈ 660 MB/hour) |
| `Cache/AlgorithmCache` | 7.8 GB | 9,361 algorithm intermediate files |
| `Cache/image` | 6.0 GB | 2,121 PNG copies of imported stills |
| `Cache/frameThumbnail` | 3.5 GB | 120,089 timeline thumbnail PNGs |
| `Cache/ILASDKDB` | 2.3 GB | algorithm SDK database |
| `Cache/audioWave` | 2.1 GB | 1,436 waveform files |

## Safety analysis (verified on-machine)

- Plaintext draft configs (`draft_agency_config.json`, `template.json`) reference only `Cache/effect` (8,635 refs), `Cache/onlineMaterial` (13), `Cache/artistEffect` (2). The main `draft_info.json` is encrypted, so a conservative exclusion list is warranted.
- **Suggested regenerable whitelist**: `recognize`, `frameThumbnail`, `audioWave`, `AlgorithmCache`, `ILASDKDB`, `image`, `tmp`, `RemuxCache`, `prerender`, `segmentPrerenderCache`, `MotionBlurCache`, `importcache3`, `ttsTemp`
- **Must-exclude** (draft-referenced or downloaded assets): `effect`, `onlineMaterial`, `artistEffect`, `music`, `AITextTemplate`, `template`, `local_models`, `AigcMaterailCache`, `agencycache` — and never touch `User Data/Projects`
- Real-world verification: deleting the whitelist set freed **76 GB**; projects from 2025-03 opened cleanly afterwards (thumbnails/waveforms regenerate on demand), and the app **automatically recreated** `AlgorithmCache`/`audioWave` on next launch — direct proof of regenerability.
- Suggested guards, mirroring `clean_final_cut_pro_generated_caches()`: only the default location, skip when `com.lemon.lvpro` is running, whitelist-only subdirectory deletion.

CapCut (international version, presumably `~/Movies/CapCut/` with the same layout) is unverified on my side and could be a follow-up once someone confirms the paths.

Happy to submit a PR following the FCP pattern in `lib/clean/app_caches.sh` if this direction sounds good.

YeJe-cpu:
## Summary

- Adds `clean_jianying_pro_generated_caches()` to `mo clean`, covering JianyingPro (剪映专业版 / CapCut CN, `com.lemon.lvpro`), which stores heavy generated caches under `~/Movies/JianyingPro/User Data/Cache/` instead of `~/Library/Caches` — the same problem class as Final Cut Pro (#843).
- Field data (issue #1277): on a real machine this cache reached **83 GB** (vs. 37 GB of actual drafts), dominated by 54 GB of uncompressed subtitle-recognition PCM scratch that the editor never cleans up after use.
- Implements the conservative pattern from `clean_final_cut_pro_generated_caches`, registered in `clean_video_tools()`.

Closes #1277.

## Safety Review

- Affects cleanup behavior (new `mo clean` target). Boundaries, chosen to be strictly narrower than the FCP precedent:
  - **Fixed default root only**: `~/Movies/JianyingPro/User Data/Cache`; no `find` recursion, no globbing outside the root; symlinked root/subdirs are rejected (`-d && ! -L`).
  - **Explicit whitelist of regenerable subdirs only** (recognize / frameThumbnail / audioWave / AlgorithmCache / ILASDKDB / image / prerender & remux temp). Anything not on the list is preserved by construction — including draft-referenced or downloaded assets (`effect` — referenced 8,000+ times by plaintext draft configs — `onlineMaterial`, `artistEffect`, `music`, `AITextTemplate`, `template`, `local_models`, `AigcMaterailCache`, `agencycache`).
  - **Never touches `User Data/Projects`** (user drafts) or any sibling of `Cache`.
  - **Skips while the editor is running.** The running check is narrowed to the main executable path because JianyingPro ships an always-resident menu-bar agent (`VideoFusion-macOSTrayHelper`) inside the app bundle; a broad `-f "/VideoFusion-macOS.app/"` match would read the helper as "editor running" and permanently skip cleanup (caught during on-machine verification, covered by a regression test).
  - All deletions route through `safe_clean` (whitelist honor, protected-path checks, dry-run, op-log).
- On-machine verification (macOS 15.7 Intel, JianyingPro 11.1.0): removing the whitelist set reclaimed ~76 GB; projects from 2025-03 reopened cleanly; the app recreated scratch dirs on next launch. `should_protect_path` / `validate_path_for_deletion` both pass for the new targets.

## Tests

- `MOLE_TEST_NO_AUTH=1 bats tests/clean_app_caches.bats` — 46/46 pass, including 4 new tests:
  - whitelist targets cleaned; excluded assets (`effect`, `music`, `AigcMaterailCache`, `agencycache`) and `Projects` never referenced
  - tray-helper-only process table does not read as "running" (regression for the narrowed pgrep)
  - skips while the editor is running
  - no-op when the cache root is absent
- `MOLE_TEST_NO_AUTH=1 bats tests/clean_core.bats tests/clean_apps.bats` — pass (66/66 on clean_apps).
- `./scripts/check.sh --format` clean; `shellcheck -x lib/clean/app_caches.sh` clean; `bash -n` across `bin lib` clean.
- Manual: sourced the function against the real cache with a print-only `safe_clean` stub — emits exactly the whitelisted existing subdirs, nothing else.

## Safety-related changes

- New cleanup target as described above; strictly whitelist-based, no recursive discovery, no new sudo/osascript/launchctl usage.

### DIFF
--- a/lib/clean/app_caches.sh
+++ b/lib/clean/app_caches.sh
@@ -397,8 +397,15 @@ clean_jianying_pro_generated_caches() {
     #   local_models, AigcMaterailCache, agencycache); plaintext draft configs
     #   reference effect 8000+ times, so anything not on this list is preserved.
     #
+    # image/ and importcache3/ are deliberately excluded: both hold copies of
+    # material the user imported, draft_info.json is encrypted so no plaintext
+    # reference check can prove they are unreferenced, and mo clean deletes
+    # permanently. If the user has since moved or deleted the source file, the
+    # cached copy is the only remaining one. Revisit only with evidence that
+    # the editor re-imports from the original on demand.
+    #
     # Verified on a real machine (macOS 15.7 Intel, JianyingPro 11.1.0):
-    # removing this set reclaimed ~76GB, 2025-era projects reopened cleanly, and
+    # removing this set reclaimed ~70GB, 2025-era projects reopened cleanly, and
     # the app recreated the scratch directories on next launch.
     local -a regenerable_subdirs=(
         recognize
@@ -406,12 +413,10 @@ clean_jianying_pro_generated_caches() {
         audioWave
         AlgorithmCache
         ILASDKDB
-        image
         RemuxCache
         prerender
         segmentPrerenderCache
         MotionBlurCache
-        importcache3
         ttsTemp
         tmp
     )
--- a/lib/clean/app_caches.sh
+++ b/lib/clean/app_caches.sh
@@ -359,6 +359,77 @@ clean_final_cut_pro_generated_caches() {
     safe_clean "${fcp_cache_targets[@]}" "Final Cut Pro generated cache"
 }
 
+jianying_pro_is_running() {
+    command -v pgrep > /dev/null 2>&1 || return 1
+
+    # Match the main editor process only. Narrow the -f pattern to the primary
+    # executable path so the always-resident menu-bar agent
+    # (.../Frameworks/VideoFusion-macOSTrayHelper.app/.../VideoFusion-macOSTrayHelper)
+    # does not read as "editor running" and permanently skip cleanup.
+    pgrep -x "VideoFusion-macOS" > /dev/null 2>&1 && return 0
+    pgrep -f "/VideoFusion-macOS.app/Contents/MacOS/VideoFusion-macOS" > /dev/null 2>&1 && return 0
+    return 1
+}
+
+clean_jianying_pro_generated_caches() {
+    local cache_root="$HOME/Movies/JianyingPro/User Data/Cache"
+    [[ -d "$cache_root" && ! -L "$cache_root" ]] || return 0
+
+    if jianying_pro_is_running; then
+        echo -e "  ${GRAY}${ICON_WARNING}${NC} JianyingPro generated caches · skipped (JianyingPro running)"
+        note_activity
+        return 0
+    fi
+
+    # JianyingPro (剪映专业版 / CapCut CN, com.lemon.lvpro) generated cache
+    # cleanup (issue #1277). Same shape as Final Cut Pro (#843): the editor
+    # keeps heavy generated caches under ~/Movies/JianyingPro/User Data/Cache/
+    # instead of ~/Library/Caches, so standard cleanup never reaches them.
+    #
+    # Safety scope for the first pass:
+    # - only the default cache root under ~/Movies; never User Data/Projects
+    #   (the user's editable drafts) or any sibling of Cache;
+    # - only remove an explicit whitelist of regenerable subdirectories:
+    #   subtitle-recognition PCM scratch, frame thumbnails, audio waveforms,
+    #   algorithm scratch, and prerender/remux temp;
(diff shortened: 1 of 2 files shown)


tw93:
@YeJe-cpu Merged, thanks for the thorough writeup and the on-machine numbers.

I pushed one commit to your branch before merging: dropped `image` and `importcache3` from the whitelist. Both hold copies of imported material rather than derived scratch, `draft_info.json` is encrypted so the plaintext audit cannot clear them, and `mo clean` deletes permanently, so a user who had moved the source file would lose the only remaining copy. The 76 GB in #1277 is fully accounted for by the six measured directories, so those two had no verified evidence behind them either. The remaining set still reclaims about 70 GB.

I also pinned the new tests to `/bin/bash` so they exercise bash 3.2 like the rest of the file.

CapCut international would be a good follow-up if you can confirm the paths.

YeJe-cpu:
Thanks @tw93 — the refinement makes sense, and honestly the reasoning behind it stuck with me more than the two directories themselves.

"Rather keep something than wrongly delete it" feels like the right instinct for irreversible cleanup. My evidence came from one machine where I could delete, reopen the projects, and see what happened. You're shipping to people whose machines you'll never see, and `mo clean` doesn't route through the Trash — under those constraints, "the plaintext audit cannot clear it" should outweigh "it looked derived to me." That's a better default than the one I proposed.

Good catch on `/bin/bash` as well.

On CapCut international: I don't have it installed at the moment, so I have no paths I'd be comfortable vouching for. If I end up using it down the road I'll keep an eye on the layout and report back if it lines up.

Thanks for the quick turnaround on this one.

PR #1291, opened 2026-07-25
Centralize optimize task catalog

M-Hassan-Raza:
Optimize task metadata was duplicated across health JSON, whitelist selection, and execution dispatch. That allowed Legacy Overrides to disappear from the whitelist and made optimize serialize, then parse, its own JSON.

This adds one validated Bash 3.2 catalog for task IDs, handlers, labels, descriptions, ordering, and safety. Health JSON stays unchanged, whitelist entries come from the same catalog, and dispatch resolves exact allow-listed handlers without eval or fallback routing.

Validation:
./scripts/check.sh
MOLE_TEST_NO_AUTH=1 bats tests/optimize_catalog.bats tests/optimize.bats tests/optimize_db.bats tests/manage_whitelist.bats tests/health_json.bats
MOLE_TEST_NO_AUTH=1 MOLE_TEST_JOBS=2 BATS_FORMATTER=tap ./scripts/test.sh
make build
MOLE_TEST_NO_AUTH=1 MOLE_DRY_RUN=1 ./mole optimize


tw93:
@M-Hassan-Raza Merged. This is a meaningful cleanup, not just moving metadata around: the Bash 3.2 catalog fixes the missing Legacy Overrides entry, removes the JSON round-trip, and keeps the existing health output contract stable. The exact handler validation also gives future optimize tasks a much safer path into dispatch.

Thank you for carrying the refactor all the way through the compatibility and test details.

PR #1293, opened 2026-07-26
Track optimize task outcomes explicitly

M-Hassan-Raza:
Optimize counted every catalog task as applied, even when work was unchanged, skipped, unavailable, needed attention, or failed. That made terminal summaries, exit status, and history disagree with what actually happened.

This adds one validated Bash 3.2 outcome contract for all optimize tasks. Each dispatch records exactly one result, partial failures take precedence, terminal and history summaries share the same ledger, and failed runs exit nonzero. Probe and removal failures are reported by phase, slow system checks are bounded, and unknown network state never authorizes maintenance. Cleanup targets and deletion routing are unchanged.

Validation:
./scripts/check.sh
MOLE_TEST_NO_AUTH=1 bats tests/optimize_outcomes.bats tests/optimize_catalog.bats tests/optimize.bats tests/optimize_db.bats tests/optimize_removal_outcomes.bats tests/optimize_probe_outcomes.bats tests/optimize_summary.bats tests/history.bats tests/clean_system_maintenance.bats
MOLE_TEST_NO_AUTH=1 MOLE_TEST_JOBS=2 BATS_FORMATTER=tap ./scripts/test.sh
go test ./...
make build


tw93:
@M-Hassan-Raza, this was a serious piece of work. Thank you for following Optimize's outcome path all the way through instead of patching one misleading summary. Applied, skipped, unavailable, and failed tasks now produce one honest summary, history record, and exit status.

It is on `main` now. Script installs can run `mo update --nightly` and try `mo optimize` on a mix of changed and skipped tasks; the next stable release will include it too.

M-Hassan-Raza:
> @M-Hassan-Raza, this was a serious piece of work. Thank you for following Optimize's outcome path all the way through instead of patching one misleading summary. Applied, skipped, unavailable, and failed tasks now produce one honest summary, history record, and exit status.
> 
> It is on `main` now. Script installs can run `mo update --nightly` and try `mo optimize` on a mix of changed and skipped tasks; the next stable release will include it too.

Great! Would love to try when its on the stable release :D Thanks so much for taking a detailed look at teh PR.

PR #1294, opened 2026-07-26
fix(core): guard empty files_to_clean expansion in app_protection.sh

### LINKED ISSUE #1127: [BUG] `kept_files[@]: unbound variable` in clean/apps.sh during 'mo clean' app leftovers scan
## Describe the bug

Running `mo clean` aborts during the **App leftovers** phase with:

```
➤ App leftovers
  ✓ Found 190 active/installed apps
/opt/homebrew/Cellar/mole/1.43.0/libexec/bin/../lib/clean/apps.sh: line 725: kept_files[@]: unbound variable
```

`set -euo pipefail` is active, so the unbound-variable reference is fatal and the whole `clean` run stops. This is the same class of bug as #790 / #863 (and the same kind of fix as #792 and #867), just in a code path that wasn't guarded yet.

## Root cause

`lib/clean/apps.sh` runs under `#!/bin/bash`, which on macOS is **bash 3.2.57** (Apple never ships newer due to GPLv3). In bash < 4.4, expanding an **empty** array with `"${arr[@]}"` under `set -u` is treated as an unbound-variable reference, even when the array was explicitly declared with `local -a arr=()`.

In the orphaned-service whitelist filter (lines 714–726):

```bash
if [[ $orphaned_count -gt 0 && ${#WHITELIST_PATTERNS[@]} -gt 0 ]]; then
    local -a kept_files=()
    for orphan_file in "${orphaned_files[@]}"; do
        if is_path_whitelisted "$orphan_file"; then
            debug_log "Skipping whitelisted orphan service: $orphan_file"
            continue
        fi
        kept_files+=("$orphan_file")
    done
    orphaned_count=${#kept_files[@]}
    orphaned_files=("${kept_files[@]}")   # <-- line 725
fi
```

The block is entered when orphans were found **and** a whitelist exists. If **every** detected orphan matches the whitelist, `kept_files` stays empty, and line 725 expands an empty array → abort on bash 3.2.

## Steps to reproduce

Requires orphaned system services to be detected while every one of them is covered by a whitelist pattern (so `kept_files` ends up empty). It is therefore intermittent, depending on the machine's current orphan state plus configured whitelist.

A deterministic reproduction of the exact failing line under system bash:

```bash
/bin/bash -c 'set -euo pipefail; declare -a kept_files=(); orphaned_files=("${kept_files[@]}"); echo ok'
# bash: kept_files[@]: unbound variable   (bash 3.2)
# ok                                       (bash >= 4.4)
```

## Expected behavior

`mo clean` completes the App leftovers scan when all detected orphans are whitelisted (i.e. there is nothing left to clean), instead of aborting.

## Proposed fix

Guard the empty-array expansion, matching the idiom already used in #792 / #867:

```diff
         orphaned_count=${#kept_files[@]}
-        orphaned_files=("${kept_files[@]}")
+        # macOS /bin/bash is 3.2, which treats "${empty[@]}" as an unbound
+        # variable under `set -u`. Guard the expansion so an all-whitelisted
+        # scan (kept_files empty) doesn't abort the whole clean run.
+        if (( orphaned_count > 0 )); then
+            orphaned_files=("${kept_files[@]}")
+        else
+            orphaned_files=()
+        fi
     fi
```

## Verification

Tested on macOS bash 3.2.57 by extracting the exact block (lines 714–726) and forcing the all-whitelisted (empty `kept_files`) precondition:

- **before** → `line: kept_files[@]: unbound variable` (exit 1)
- **after**  → completes cleanly, `orphaned_count=0` (exit 0)

Patched file also passes `bash -n` and a full `mo clean --dry-run` runs through App leftovers without error.

## Environment

- Mole: 1.43.0 (latest release; line still present on `main`)
- macOS: Darwin 27 (Apple Silicon)
- Shell: system `/bin/bash` 3.2.57

Lyleton:
## Summary

- Guard `"${files_to_clean[@]}"` expansion with `${#files_to_clean[@]} -gt 0` check in `find_app_files` dedup loop (`lib/core/app_protection.sh` line 1048)
- On macOS built-in Bash 3.2 + `set -u`, expanding an empty array is fatal ("unbound variable"), causing `mo uninstall` to hang during file scan
- Uses the same guard pattern established in #792 / #867

Closes #1127

## Test plan

- [x] `bash -n lib/core/app_protection.sh` passes
- [x] `./scripts/check.sh --format` passes
- [x] `MOLE_TEST_NO_AUTH=1 bats tests/uninstall_safety.bats tests/uninstall_scan_bash32.bats` — 25/25 pass
- [ ] `mo uninstall` with an app that has no leftover files no longer triggers unbound variable error

### DIFF
--- a/lib/core/app_protection.sh
+++ b/lib/core/app_protection.sh
@@ -1045,12 +1045,14 @@ find_app_files() {
             while IFS= read -r -d '' derived_path; do
                 mole_name_has_bundle_id_boundary "$derived_path" "$bundle_id" || continue
                 already_added=false
-                for existing_path in "${files_to_clean[@]}"; do
-                    if [[ "$existing_path" == "$derived_path" ]]; then
-                        already_added=true
-                        break
-                    fi
-                done
+                if [[ ${#files_to_clean[@]} -gt 0 ]]; then
+                    for existing_path in "${files_to_clean[@]}"; do
+                        if [[ "$existing_path" == "$derived_path" ]]; then
+                            already_added=true
+                            break
+                        fi
+                    done
+                fi
                 [[ "$already_added" == "true" ]] || files_to_clean+=("$derived_path")
             done < <(command find "$derived_root" -maxdepth 1 -type d -print0 2> /dev/null)
         done
--- a/tests/uninstall_naming_variants.bats
+++ b/tests/uninstall_naming_variants.bats
@@ -140,6 +140,22 @@ setup() {
     [[ ! "$result" =~ Library/Containers/com.tencent.otherapp.Helper ]]
 }
 
+@test "find_app_files handles the first derived bundle match under bash 3.2 set -u" {
+    mkdir -p "$HOME/Library/Containers/com.example.Widget.Helper"
+
+    run env HOME="$HOME" PROJECT_ROOT="$PROJECT_ROOT" /bin/bash --noprofile --norc <<'EOF'
+set -euo pipefail
+source "$PROJECT_ROOT/lib/core/base.sh"
+source "$PROJECT_ROOT/lib/core/log.sh"
+source "$PROJECT_ROOT/lib/core/app_protection.sh"
+find_app_files "com.example.Widget" "Widget"
+EOF
+
+    [ "$status" -eq 0 ] || return 1
+    [[ "$output" == *"Library/Containers/com.example.Widget.Helper"* ]] || return 1
+    [[ "$output" != *"unbound variable"* ]]
+}
+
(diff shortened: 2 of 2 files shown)


tw93:
@Lyleton, thank you for catching a Bash 3.2 trap that looks harmless in newer shells but can stall the whole uninstall scan on macOS. I added the first derived-bundle regression because this exact empty-array path deserves to stay pinned.

The fix is on `main`. With a script install, run `mo update --nightly` and retry the uninstall scan that used to hang. Stable will include it next.

PR #1302, opened 2026-07-27
feat(purge): discover orphaned Xcode DerivedData from deleted workspaces

yamamel:
## Problem

Xcode stores build outputs in `~/Library/Developer/Xcode/DerivedData/` indexed by a hash suffix, not by project location. Unlike Rust (`target/`) or Node (`node_modules/`) which live inside the project worktree, DerivedData survives worktree deletion as orphaned build cache that `mo purge` could not previously detect.

## Solution

Each DerivedData subdirectory carries an `info.plist` with a `WorkspacePath` field recording the original source tree. When that path no longer exists on disk, the entry is reclaimable.

- **Discovery**: add `discover_orphaned_xcode_derived_data()` — scans DerivedData subdirs, reads WorkspacePath via PlistBuddy, emits orphan paths to stdout
- **Xcode guard**: skip while Xcode process is running (same guard as `clean_xcode_derived_data`)
- **Integration**: append orphans to `all_found_items[]` in `clean_project_artifacts()` so they participate in size calc, activity classification, and the multi-select menu
- **Display**: flag orphan entries with artifact type `DerivedData [orphaned]`, show the deleted workspace path, and render a `deleted_path → DDPath` hover tooltip
- **Safety bypass**: orphans live inside `~/Library/Developer/Xcode/DerivedData/` which is outside `PURGE_SEARCH_PATHS` — skip `is_safe_configured_purge_artifact()` for orphan-flagged entries since safety is established by the WorkspacePath existence check

## Verification

- All 85 purge tests pass (0 failure)
- Manually confirmed 3 orphans on a real dev machine with correct detection

### DIFF
--- a/lib/clean/project.sh
+++ b/lib/clean/project.sh
@@ -422,6 +422,38 @@ is_protected_purge_artifact() {
     return 1
 }
 
+# Discover orphaned Xcode DerivedData entries whose source workspace or project
+# no longer exists on disk. These survive deleted worktrees because Xcode stores
+# build outputs in a global ~/Library/Developer/Xcode/DerivedData/ directory
+# indexed by a hash suffix, not by project location. Each subdirectory carries
+# an info.plist whose WorkspacePath field records the original source tree path;
+# when that path no longer exists, the DerivedData is reclaimable.
+# Outputs one absolute DerivedData subdirectory path per line to stdout.
+discover_orphaned_xcode_derived_data() {
+    local dd_dir="$HOME/Library/Developer/Xcode/DerivedData"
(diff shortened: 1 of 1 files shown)


tw93:
@yamamel, the `WorkspacePath` angle is thoughtful, and I appreciate how completely you carried it through the purge flow. The hard part is that a missing workspace path also describes an external or network volume that is only temporarily offline.

DerivedData cleanup will stay in `mo clean`, where users can inspect candidates with `mo clean --dry-run` and whitelist anything they need to keep. That gives users a review step instead of making `purge` guess whether data is safe to remove, so this stays not planned.

PR #1306, opened 2026-07-27
status: balance the two-column layout without growing the dashboard

gloiiire:
## What

Improve the `mo status` two-column layout when one card, such as CPU with many cores shown, is much taller than its neighbour.

The renderer now compares the stable fixed-pair layout with height-aware stacked rows and uses stacking only when it is strictly shorter. Normal views keep the existing card pairs, while tall CPU cards can use otherwise empty vertical space.

The maintainer follow-up removes the unrelated theme extraction so this PR stays focused and preserves the existing colours.

## Why it fits `status`

- No new metrics, rows, flags, or actions.
- `status` remains read-only and keeps its existing compact dashboard structure.
- Dynamic stacking cannot make the dashboard taller than the current fixed-pair layout.

## Testing

- `go test ./cmd/status -count=1`
- `make verify`
- 120x40 live TUI smoke test
- Regression coverage for the production card order and the tall-card layout


tw93:
@gloiiire thank you for the thoughtful status layout improvement. I merged the PR after a maintainer follow-up that limits dynamic stacking to cases where it makes the dashboard strictly shorter, adds a regression with the production card order, and removes the unrelated theme extraction.

The change is now on `main` and available through Nightly. You can try it with `mo update --nightly`. Please let me know if you notice any remaining layout issue.

PR #1307, opened 2026-07-27
feat(status): add a GPU section (utilization, power, cores)

gloiiire:
## What

Adds a compact GPU card to `mo status`, placed right after CPU:

- **Usage** — whole-device GPU utilization with a trend sparkline, read from **IOKit without root** (on Apple Silicon). When it can't be read, the row says `unavailable` instead of a misleading `0%`.
- **Power** — GPU power draw in watts via IOReport (shown only when available).
- **Cores** — GPU core count.
- A **detail toggle** (`g` key, mirroring the existing `c`/`k` keys): cycles Device → +Renderer → +Tiler utilization rows. Apple exposes no finer per-core breakdown, so these three whole-device figures are the max granularity.

## Why it fits `status`

- **Answers a common question, read-only.** "Why is my Mac hot / fans loud / battery draining under load?" — GPU utilization is a primary signal on Apple Silicon, and there's no built-in way to see it from a terminal without `sudo powermetrics`. This surfaces it at a glance, no privileges.
- **Compact by default — not a dashboard.** The card shows a **single** `Usage` row by default; Render/Tiler are opt-in via `g`. It adds one line, not a wall of metrics.
- **Fails safe.** The card is hidden entirely when there's no real GPU data (placeholder entries are skipped), and unreadable figures render as `unavailable`, never fake zeros.
- **Reuses existing patterns.** Sparkline like the network card, `colorizePercent` coloring, the same card/toggle/prefs plumbing as CPU cores — no new flags or config surface, just one toggle key persisted like the others.

## Notes

- Apple Silicon path uses IOKit / IOReport (no root). Non-Apple / powermetrics paths degrade gracefully (rows marked `unavailable`).
- `-race` test covers concurrent GPU history access (dedicated GPU tick vs. main collection goroutine).

## Testing

- `go test ./cmd/status` passes (regex extraction, detail-toggle rows, `hasGPUCard` gating, concurrent history under `-race`).
- `golangci-lint` + `shellcheck` clean via `./scripts/check.sh --no-format`.

### DIFF
--- a/cmd/status/gpu_power_darwin.go
+++ b/cmd/status/gpu_power_darwin.go
@@ -1,0 +1,107 @@
+//go:build darwin && cgo
+
+package main
+
+/*
+#cgo LDFLAGS: -framework CoreFoundation -lIOReport
+#include <CoreFoundation/CoreFoundation.h>
+#include <string.h>
+#include <time.h>
+#include <stdint.h>
+
+// Private IOReport API — no public header, symbols live in /usr/lib/libIOReport.dylib.
+// This is how tools like macmon read power without root; it is undocumented and
+// may change between macOS releases, which is why the reader fails soft (-1).
+typedef struct __IOReportSubscription* IOReportSubscriptionRef;
+typedef CFDictionaryRef IOReportSampleRef;
+
+extern CFMutableDictionaryRef IOReportCopyChannelsInGroup(CFStringRef group, CFStringRef subgroup, uint64_t a, uint64_t b, uint64_t c);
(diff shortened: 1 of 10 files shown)


tw93:
@gloiiire Thanks for the thoughtful contribution, and sorry we won't merge this one. Mole's Status surface is intentionally staying focused on essential system-health signals; a dedicated GPU section would add another monitoring and maintenance surface without enough day-to-day value, so we're closing this as not planned. For detailed GPU monitoring, macOS's built-in tools remain the better fit.

PR #1312, opened 2026-07-28
fix(status): preserve published battery health

QuakeWang:
## Summary

- Keep the Maximum Capacity published by `system_profiler` authoritative.
- Use the IORegistry capacity ratio only when the published value is unavailable.
- Prefer `AppleRawMaxCapacity` over the buffered `NominalChargeCapacity` for that fallback.
- Never interpret AppleSmartBattery `MaxCapacity` as battery health. On Apple Silicon it is the denominator used with `CurrentCapacity` for the current charge percentage and is normally 100.

## Safety Review

- This only changes read-only status parsing.
- It does not affect cleanup, deletion, path validation, sudo boundaries, or install integrity.
- No new metric, field, command, setting, or data source is introduced.

## Verification

- Red-green regression run against the original PR behavior.
- `go test -count=1 ./...`
- `make build`
- Clean integration with the current `main` passed the same checks.
- Local JSON smoke reported `capacity=100`, `percent=100`, `cycle_count=38`, and `health=Good`.

### DIFF
--- a/cmd/status/metrics_battery.go
+++ b/cmd/status/metrics_battery.go
@@ -230,13 +230,20 @@ func getAppleSmartBatteryHealthData() (cycles int, capacity int) {
 }
 
 func parseAppleSmartBatteryHealth(out string) (cycles int, capacity int) {
-	var design, nominal, rawMax int
+	var normalizedCapacity, design, nominal, rawMax int
 	for line := range strings.Lines(out) {
 		line = strings.TrimSpace(line)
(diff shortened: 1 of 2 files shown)


tw93:
@QuakeWang Thanks for catching this and for the contribution. I adjusted the health-source priority so macOS's published capacity remains authoritative, kept the IORegistry value as a fallback, and merged the PR.

Issue #1299, opened 2026-07-27
[BUG] Unable to uninstall App Store applications on macOS Tahoe (Full Disk Access incorrectly reported)


markxorigin:
## Describe the bug

Mole is unable to uninstall App Store applications installed under `/Applications` on my macOS Tahoe system.

I tested with both **WhatsApp** and **Telegram Lite** (installed from the Mac App Store). Both normal uninstall and `--permanent` fail.

The application can be removed successfully using:

```bash
sudo rm -rf "/Applications/WhatsApp.app"
```

so it does not appear to be a macOS permission issue.

Additionally, Mole reports:

> macOS denied Trash access. Grant App Data or Full Disk Access to your terminal.

However, my terminal already has **Full Disk Access**, so the reported cause appears to be incorrect.

---

## Steps to reproduce

1. Install an App Store application (tested with WhatsApp and Telegram Lite).
2. Run:

```bash
mole uninstall WhatsApp --debug
```

or

```bash
mole uninstall --permanent WhatsApp --debug
```

3. Confirm the uninstall.
4. Mole fails with `Permission denied`.

---

## Expected behavior

The application should either:

- be successfully moved to Trash (or removed when using `--permanent`), or
- report the actual reason why it cannot be removed.

---

## Debug logs

<details>
<summary>Full debug output</summary>

```text
mole uninstall WhatsApp --debug
[DEBUG] Debug logging enabled. Session log: /Users/<user>/Library/Logs/mole/mole_debug_session.log

◎ Matched 1 app(s):
1. ‎WhatsApp  292.3MB  |  Last: Today

Proceed with uninstallation? [y/N] y

Files to be removed:

◎ ‎WhatsApp , 294.2MB
  ✓ /Applications/WhatsApp.app , 292.3MB
  ✓ ~/Library/Containers/net.whatsapp.WhatsApp , 655KB
  ✓ ~/Library/Application Scripts/net.whatsapp.WhatsApp
  ✓ ~/Library/Group Containers/group.net.whatsapp.WhatsApp.shared , 1.1MB
  ✓ ~/Library/Group Containers/group.net.whatsapp.WhatsApp.private , 8KB
  ✓ ~/Library/Application Scripts/net.whatsapp.WhatsApp.WAAppKitBridgeService
  ✓ ~/Library/Application Scripts/group.net.whatsapp.WhatsApp.shared
  ✓ ~/Library/Application Scripts/group.net.whatsapp.WhatsApp.private
[...]

PR #1322, opened 2026-07-29
fix(uninstall): honor sudo_local in removal diagnostics

### LINKED ISSUE #1321: [BUG] Uninstall failure suggests 'mole touchid' even when Touch ID is configured — diagnose_removal_failure checks /etc/pam.d/sudo, not sudo_local (Tahoe)
## Summary

On macOS Tahoe, when an app uninstall fails, Mole tells the user to run `mole touchid` even when Touch ID for `sudo` is already correctly configured. The suggestion is generated by `diagnose_removal_failure()`, which detects Touch ID by grepping `/etc/pam.d/sudo` only — but on Sonoma (14)+ and Tahoe (26) `pam_tid.so` lives in `/etc/pam.d/sudo_local`, which is exactly where Mole's own `mole touchid` writes it. So Mole's failure diagnostic disagrees with Mole's own Touch ID installer, and points the user at a non-fix.

This is a **diagnostic/messaging bug**, separate from the underlying uninstall failure in #1299 (App Management gate on root-owned `/Applications/*.app`). #1299 covers *why the removal fails* and its FDA misattribution; this issue covers *why the on-screen remediation says `mole touchid`*.

## Environment

- Mole 1.48.1 (Homebrew bottle)
- macOS 26.6, build 25G72 (Tahoe), Apple Silicon
- Terminal: Ghostty, granted **both** Full Disk Access and App Management, terminal fully relaunched afterwards
- Target: `/Applications/AppCleaner.app`, owned by `root`, in `/Applications` (`drwxrwxr-x root:admin`)

## Symptom

```
Uninstall incomplete
• Failed: AppCleaner permission denied
☞ Try 'mole touchid' or check with 'ls -l'
```

## Root cause 1 — wrong PAM file (confirmed on this machine)

`lib/core/file_ops.sh`:

```sh
1622    local touchid_file="/etc/pam.d/sudo"
...
1649    if [[ -f "$touchid_file" ]] && grep -q "pam_tid.so" "$touchid_file" 2> /dev/null; then
1650        suggestion="Try running again or check file ownership"
1651    else
1652        suggestion="Try 'mole touchid' or check with 'ls -l'"
1653    fi
```

On this machine Touch ID for sudo **is** configured, via the modern drop-in:

```
$ grep -n pam_tid /etc/pam.d/sudo
# (no match)
$ grep -n 'include\|sudo_local' /etc/pam.d/sudo
2:auth       include        sudo_local
$ grep -n pam_tid /etc/pam.d/sudo_local
2:auth       sufficient     pam_tid.so
```

`grep pam_tid.so /etc/pam.d/sudo` returns false, so `diagnose_removal_failure()` always concludes Touch ID is absent and prints `Try 'mole touchid'` — even immediately after `mole touchid` succeeded.

This is internally inconsistent with Mole's own `touchid.sh`, which is already Sonoma-aware and checks `sudo_local` first, then `sudo`:

```sh
# lib/bin/touchid.sh
19  PAM_SUDO_FILE="${MOLE_PAM_SUDO_FILE:-/etc/pam.d/sudo}"
20  PAM_SUDO_LOCAL_FILE="${MOLE_PAM_SUDO_LOCAL_FILE:-$(dirname "$PAM_SUDO_FILE")/sudo_local}"
...
32      # Check sudo_local first
34          grep -q "pam_tid.so" "$PAM_SUDO_LOCAL_FILE" 2> /dev/null && return 0
...
41      grep -q "pam_tid.so" "$PAM_SUDO_FILE" 2> /dev/null
```

`diagnose_removal_failure()` (used at both `MOLE_ERR_AUTH_FAILED` and the default branch) never checks `sudo_local`. This is also aggravated by an OS upgrade: Tahoe ships `/etc/pam.d/sudo` as a stock file containing only `auth include sudo_local`, so the Touch ID line is *never* in `sudo`.

## Root cause 2 — privacy denial is flattened into a generic "permission denied"

Two things make a TCC / App Management denial indistinguishable from an ordinary permission error at the summary layer:

1. `lib/uninstall/batch.sh` collapses the reason with a glob that also matches the privacy reason:

```sh
1424    permission*denied*) reason_summary="permission denied" ;;
```

`diagnose_removal_failure()` returns `macOS privacy permission denied` for `MOLE_ERR_PRIVACY_DENIED`, but this glob rewrites it to plain `permission denied`, dropping the "privacy/macOS" qualifier.

2. More importantly, the privacy exit code (`MOLE_ERR_PRIVACY_DENIED=14`) does not reach `diagnose_removal_failure()` on every failing path. The `14` is only produced on the unprivileged `mv` branch (`file_ops.sh:1200-1204`); when the sudo-staging path fails instead it returns a plain `1`, and `diagnose_removal_failure 1` falls through to the default branch — which is the `mole touchid` text. So the same underlying TCC/App-Management denial can surface either as the correct FDA hint (#1299) *or* as the misleading `mole touchid` hint, depending on which sub-step failed.

The net effect for the user: an App Management denial is reported as a generic "permission denied" with a Touch ID suggestion that has nothing to do with the actual cause.

## Suggested direction

- In `diagnose_removal_failure()`, reuse the `sudo_local`-aware check already present in `touchid.sh` (check `/etc/pam.d/sudo_local` first, then `/etc/pam.d/sudo`) so Touch ID is never reported as missing when it is configured. Better yet, call the existing detector rather than re-implementing the grep.
- Do not suggest `mole touchid` for a TCC/App Management denial. For `MOLE_ERR_PRIVACY_DENIED`, guide toward App Management / Full Disk Access (as #1299 discusses), and ensure that exit code is preserved end-to-end so the summary does not fall back to the generic branch.
- Consider not flattening `macOS privacy permission denied` into `permission denied` at `batch.sh:1424`, so the summary keeps the distinction.

Happy to run any diagnostic on this machine (macOS 26.6, Apple Silicon) to help narrow it down.

r266-tech:
## Summary

- Reuse the shared `check_touchid_support` detector in uninstall failure diagnostics, so Sonoma+ and Tahoe configurations in `/etc/pam.d/sudo_local` are recognized.
- Keep the existing credential and ownership guidance when Touch ID is configured, while preserving `mole touchid` guidance when it is not.
- Resolve the detector lazily as a shell function, so unrelated failures and standalone `file_ops.sh` consumers do not invoke it or a same-named program from `PATH`.

Fixes #1321.

## Safety Review

- This changes uninstall failure messaging only. It does not change deletion behavior, error classification, authentication, path validation, protected directories, symlink handling, or sudo boundaries.
- The existing shared detector is called only for authentication and generic permission diagnostics. Its output is suppressed, and a missing function preserves the previous fallback guidance.

## Tests

- `./scripts/check.sh`
- `MOLE_TEST_NO_AUTH=1 bats tests/file_ops_mole_delete.bats tests/uninstall.bats` (104 tests)
- `MOLE_TEST_NO_AUTH=1 bats tests/file_ops_size.bats tests/file_ops_safe_remove_symlink.bats tests/user_file_ops.bats tests/core_safe_functions.bats` (90 tests)

## Safety-related changes

- Uninstall diagnostics now share the modern-macOS Touch ID configuration check already used by the sudo manager; destructive and authorization boundaries are unchanged.

### DIFF
--- a/lib/core/file_ops.sh
+++ b/lib/core/file_ops.sh
@@ -1619,7 +1619,6 @@ diagnose_removal_failure() {
 
     local reason=""
     local suggestion=""
-    local touchid_file="/etc/pam.d/sudo"
 
     case "$exit_code" in
         "$MOLE_ERR_SIP_PROTECTED")
@@ -1627,7 +1626,7 @@ diagnose_removal_failure() {
             ;;
         "$MOLE_ERR_AUTH_FAILED")
             reason="authentication failed"
-            if [[ -f "$touchid_file" ]] && grep -q "pam_tid.so" "$touchid_file" 2> /dev/null; then
+            if declare -F check_touchid_support > /dev/null 2>&1 && check_touchid_support > /dev/null 2>&1; then
                 suggestion="Check your credentials or restart Terminal"
             else
                 suggestion="Try 'mole touchid' to enable fingerprint auth"
@@ -1646,7 +1645,7 @@ diagnose_removal_failure() {
             ;;
         *)
             reason="permission denied"
-            if [[ -f "$touchid_file" ]] && grep -q "pam_tid.so" "$touchid_file" 2> /dev/null; then
+            if declare -F check_touchid_support > /dev/null 2>&1 && check_touchid_support > /dev/null 2>&1; then
                 suggestion="Try running again or check file ownership"
             else
                 suggestion="Try 'mole touchid' or check with 'ls -l'"
--- a/lib/core/sudo.sh
+++ b/lib/core/sudo.sh
@@ -9,15 +9,17 @@ set -euo pipefail
 # ============================================================================
 
 check_touchid_support() {
+    local pam_sudo_file="${MOLE_PAM_SUDO_FILE:-/etc/pam.d/sudo}"
+    local pam_sudo_local_file="${MOLE_PAM_SUDO_LOCAL_FILE:-$(dirname "$pam_sudo_file")/sudo_local}"
+
     # Check sudo_local first (Sonoma+)
-    if [[ -f /etc/pam.d/sudo_local ]]; then
-        grep -q "pam_tid.so" /etc/pam.d/sudo_local 2> /dev/null
-        return $?
+    if [[ -f "$pam_sudo_local_file" ]] && grep -q "pam_tid.so" "$pam_sudo_local_file" 2> /dev/null; then
+        return 0
     fi
 
     # Fallback to checking sudo directly
-    if [[ -f /etc/pam.d/sudo ]]; then
-        grep -q "pam_tid.so" /etc/pam.d/sudo 2> /dev/null
+    if [[ -f "$pam_sudo_file" ]]; then
+        grep -q "pam_tid.so" "$pam_sudo_file" 2> /dev/null
         return $?
     fi
     return 1
--- a/tests/file_ops_mole_delete.bats
+++ b/tests/file_ops_mole_delete.bats
@@ -497,6 +497,106 @@ EOF
     [[ "$output" != *"Touch ID"* ]]
 }
 
+@test "unrelated removal diagnostics do not probe Touch ID" {
+    run /bin/bash --noprofile --norc <<EOF
+$(prelude)
+touchid_checks=0
+check_touchid_support() {
+    touchid_checks=\$((touchid_checks + 1))
+    printf 'detector noise\n'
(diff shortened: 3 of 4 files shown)


tw93:
@r266-tech Thanks. Removal diagnostics now use the shared Touch ID detector, with regression coverage for `sudo_local` and the legacy `sudo` fallback. This is merged into `main`.

Available now via `mo update --nightly`. The separate App Management removal failure remains tracked in #1299.

PR #1324, opened 2026-07-30
feat(purge): clean stale Terraform providers

avegao:
## Summary

  - Add Terraform provider cleanup to `mo purge`.
  - Detect unselected provider versions under `.terraform/providers` using `.terraform.lock.hcl`.
  - Keep the selected provider version and all its platform variants.
  - Report unreferenced versions in the global Terraform plugin cache without deleting them.

  ## Safety Review

    - Yes. It extends `purge` cleanup behavior.
    - Yes. Provider candidates must match the expected `.terraform/providers/<host>/<namespace>/<type>/<version>` layout, remain inside configured purge
    roots, have a valid lockfile, and pass the existing safe deletion funnel.
    - Projects without a valid lockfile fail closed.
    - Global plugin caches are analysis-only and are never added to the deletion menu.
    - Scanning is bounded by `MOLE_TIMEOUT_TERRAFORM_SCAN_SEC`.

  ## Tests

  - `git diff --check`
  - `shellcheck lib/clean/project.sh`
  - `bash -n lib/clean/project.sh bin/purge.sh lib/core/timeouts.sh`
  - Manual dry-run with active and stale Terraform provider versions.
  - Manual validation that paths outside configured roots are rejected.
  - Manual validation that global plugin cache entries are reported but not deleted.

  Automated Bats tests and formatting checks could not be executed because `bats` and `shfmt` are not installed in the local environment.

  ## Safety-related changes

  - Added regression tests for lockfile selection, missing lockfiles, root validation, dry-run behavior, and analysis-only global cache handling.
  - Reuses the existing purge confirmation, dry-run, operation logging, activity recheck, and safe removal flow.


tw93:
@avegao Thanks for the contribution. I'm going to close this as not planned. Mole keeps purge focused on large, clearly disposable project artifacts; provider packages can be active, shared, or needed offline, and the current lock and path checks are not strong enough for a permanent deletion path. Users who need to reclaim this space can review a project's `.terraform` directory directly and rebuild it with `terraform init` when appropriate.

PR #1329, opened 2026-07-31
fix(uninstall): recover App Management-blocked app bundles through Finder

r266-tech:
## Summary

- Preserve whether a sudo-required uninstall was downgraded because an ancestor is mutable.
- When the unprivileged direct Trash move is permission-denied for an exact top-level `/Applications/*.app` bundle, retry through the existing Finder/AppleScript Trash path.
- Return a distinct App Management/Finder diagnostic if that retry fails, while leaving sandbox container paths on their existing App Data / Full Disk Access path.

Addresses #1299.

## Safety Review

- This changes the uninstall deletion sink, so the fallback is gated on all four observed conditions: the caller originally required sudo, the mutable-ancestor guard downgraded it, the target is an immediate `/Applications/*.app` child, and the direct `mv` returned a permission/TCC denial.
- The change does not bypass `_mole_privileged_path_has_mutable_ancestor` and never retries with sudo. Finder receives the path as argv, not interpolated AppleScript.
- Finder success is accepted only when the original bundle path is gone. Finder failure returns a dedicated error and does not fall through to permanent deletion or privileged staging.
- `~/Library/Containers`, Group Containers, and Application Scripts keep their existing direct-move and privacy-diagnostic behavior.
- `--permanent` remains intentionally unchanged and fail-closed below a mutable `/Applications` ancestor; this patch restores the default recoverable Trash path only.

## Tests

- `MOLE_TEST_NO_AUTH=1 bats tests/file_ops_mole_delete.bats tests/user_file_ops.bats` (69 passed)
- `MOLE_TEST_NO_AUTH=1 bats tests/uninstall.bats tests/uninstall_remove_file_list.bats` (77 passed)
- `MOLE_TEST_NO_AUTH=1 bats tests/clean_user_core.bats` (46 passed)
- `MOLE_TEST_NO_AUTH=1 ./scripts/test.sh` (1240 passed, 6 environment-dependent skips)
- `./scripts/check.sh --no-format` (go vet, ShellCheck, and Bash syntax passed)
- All authorization-capable paths were mocked; verification did not invoke real sudo, Finder, or AppleScript authorization.

## Safety-related changes

- Adds a narrowly scoped, recoverable Finder fallback for App Management-blocked top-level app bundles after the existing privilege downgrade.

### DIFF
--- a/lib/core/file_ops.sh
+++ b/lib/core/file_ops.sh
@@ -16,6 +16,7 @@ readonly MOLE_ERR_AUTH_FAILED=11
 readonly MOLE_ERR_READONLY_FS=12
 readonly MOLE_ERR_PROTECTED_PATH=13
 readonly MOLE_ERR_PRIVACY_DENIED=14
+readonly MOLE_ERR_APP_MANAGEMENT_DENIED=15
 
 # Ensure dependencies are loaded
 _MOLE_CORE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
@@ -750,6 +751,7 @@ safe_sudo_remove() {
 mole_delete() {
     local path="$1"
     local needs_sudo="${2:-false}"
+    local sudo_was_downgraded=false
     local mode="${MOLE_DELETE_MODE:-permanent}"
 
     [[ -z "$path" ]] && return 1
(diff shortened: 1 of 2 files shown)


tw93:
@r266-tech, thank you for tracing the Tahoe/App Management failure and for adding focused tests. I am sorry that the uninstall path changed on `main` while this PR was open.

After reviewing the deletion sink, we cannot safely add the Finder retry here. The app path is below a user-mutable `/Applications`, leaving a path-replacement window between validation and Finder resolving it; if Finder then fails, the current ordering can also leave the bundle in place after launch services, login items, and the process have already been torn down. `main` now resolves #1299 by stopping before any teardown and directing the user to Finder, without adding an elevated or path-based deletion bypass. That stricter fix supersedes this approach, so I am closing this PR unmerged.

Issue #1054, opened 2026-06-04
[FEATURE]  Mo update --nightly only install a new version if there is a new nightly avialble


iammike2:
> Just look at the current state of events
> 
> mike@MACMINI /Users/mike % `mole --version`
> 
> Mole version 1.41.0
> Channel: Nightly (e31d46f)
> macOS: 15.7.7
> Architecture: arm64
> Kernel: 24.6.0
> SIP: Enabled
> Disk Free: 428.02GB
> Install: Manual
> Shell: /bin/zsh
> 
> mike@MACMINI /Users/mike % `mole update --nightly`
> 
> | Fetching Mole source, main...
> ✓ Installed mole to /usr/local/bin
> ✓ Installed mo alias
> ✓ Installed modules
> ✓ Installed libraries
> Downloading analyze...
> ✓ Downloaded analyze binary
> Downloading status...
> ✓ Downloaded status binary
> ◎ Mole installed successfully, version 1.41.0
> Usage:
>   mo                           # Interactive menu
>   mo clean                     # Deep cleanup
>   mo uninstall                 # Remove apps + leftovers
>   mo optimize                  # Check and maintain system
>   mo analyze                   # Explore disk usage
>   mo status                    # Monitor system health
>   mo touchid                   # Configure Touch ID for sudo
>   mo update                    # Update to latest version
>   mo --help                    # Show all commands
> 
> ✓ Updated to nightly build (main), 1.41.0
> 
> mike@MACMINI /Users/mike % `mole --version       `
> 
> Mole version 1.41.0
> Channel: Nightly (e31d46f)
> macOS: 15.7.7
> Architecture: arm64
> Kernel: 24.6.0
> SIP: Enabled
> Disk Free: 428.01GB
> Install: Manual
> Shell: /bin/zsh

As you can see it does all that 

> | Fetching Mole source, main...
> ✓ Installed mole to /usr/local/bin
> ✓ Installed mo alias
> ✓ Installed modules
> ✓ Installed libraries
> Downloading analyze...
> ✓ Downloaded analyze binary
> Downloading status...
> ✓ Downloaded status binary
> ◎ Mole installed successfully, version 1.41.0

and the nightly version is exactly the same. A bit overkill if you ask me hahahaha

tw93:
[...]

Issue #960, opened 2026-05-24
[FEATURE] Offer a whitelist option for detaching images in mo optimize


iammike2:
## Feature description

currently when running `mo optimize --dry-run` it gives me this in debug

```
/Volumes/EXT3/Mail/TB.dmg ← /Volumes/EXT3/Mail/TB.dmg
    /Volumes/mail ← /Volumes/EXT3/Mail/TB.dmg
  → Would offer detach for 2 mounted image(s) 

```

1- They are the same image, so why offer detach for 2 mounted image
2- How can I whitelist this? This is my mail folder and I don't want that detached.

Mole version 1.39.1
Channel: Nightly (969a0fa)
macOS: 15.7.7
Architecture: arm64
Kernel: 24.6.0
SIP: Enabled
Disk Free: 402Gi
Install: Manual
Shell: /bin/zsh

tw93:
@iammike2 thanks for the report. Both parts addressed.

1. Duplicate detach offer: the `hdiutil info` parser was treating any line with an absolute path as a mount, including `image-alias` / `icon-path` / `shadow-path`. Tightened to only read `/dev/disk*` lines, which are the only ones that actually carry a mount column. Fixed in `aafaee38`. Your `TB.dmg ← TB.dmg` row should be gone.

2. Whitelist: already supported, just not visible enough. `mo optimize --whitelist` accepts both mount paths (`/Volumes/mail`) and image paths (`/Volumes/EXT3/Mail/TB.dmg`); whichever you add will skip the detach offer. Added a hint line under the candidate list in `74442ba3` so it's discoverable from the prompt.

Please update and re-run:

```bash
mo update --nightly
mo optimize --dry-run
```

Leaving this open until you confirm.

iammike2:
thanks @tw93 

I can't see a way to add that path via  `mo optimize --whitelist`

Adding it manually in `/Users/<user>/.config/mole/whitelist_optimize` like this

# Mole Optimization Whitelist - These checks will be skipped during optimization

> system_maintenance
> cache_refresh
> saved_state_cleanup
> fix_broken_configs
> network_optimization
> sqlite_vacuum
> launch_services_rebuild
> font_cache_rebuild
> dock_refresh
> prevent_network_dsstore
> periodic_maintenance
> /Volumes/EXT3/Mail/TB.dmg
> /Volumes/mail

Still gives me this

>   • Mounted image detach candidates:
[...]

PR #1337, opened 2026-08-02
fix(update): stop hashless nightly reinstall loops

### LINKED ISSUE #1336: [BUG] For some reason mo update --nightly keeps reinstalling all the time
Mole version 1.48.1
Channel: Nightly
macOS: 15.7.8
Architecture: arm64
Kernel: 24.6.0
SIP: Enabled
Disk Free: 436.96GB
Install: Manual
Shell: /bin/zsh

Running `mo update --nightly` keeps doing this

> | Fetching Mole source, main...
> ✓ Installed mole to /usr/local/bin
> ✓ Installed mo alias
> ✓ Installed modules
> ✓ Installed libraries
> Downloading analyze...
> ✓ Downloaded analyze binary
> Downloading status...
> ✓ Downloaded status binary
> ◎ Mole installed successfully, version 1.48.1
> Usage:
>   mo                           # Interactive menu
>   mo clean                     # Deep cleanup
>   mo uninstall                 # Remove apps + leftovers
>   mo optimize                  # Check and maintain system
>   mo analyze                   # Explore disk usage
>   mo status                    # Monitor system health
>   mo touchid                   # Configure Touch ID for sudo
>   mo update                    # Update to latest version
>   mo --help                    # Show all commands
> 
> ✓ Updated to nightly build (main), 1.48.1

The only thing I find strange that the commit version is missing from the `mo --version`

r266-tech:
Fixes #1336.

## What changed

- Fall back from the unauthenticated commit API to a bounded, non-interactive `git ls-remote` lookup for `main`.
- Accept the fallback SHA only after a successful probe, with ambient Git config, credential helpers, URL rewrites, and non-HTTPS transports disabled.
- If neither lookup produces a trustworthy 40-character SHA, stop an ordinary nightly update before downloading or installing anything. An explicit `--force` keeps the existing repair behavior.

This prevents an unknown HEAD from being treated as evidence that a reinstall is needed. A successful update continues to pass the resolved SHA to the installer, so `COMMIT_HASH` is written and the next nightly check is idempotent.

## Verification

- `MOLE_TEST_NO_AUTH=1 bats tests/update.bats` — 27/27 passed
- `./scripts/check.sh` — formatting, `go vet`, ShellCheck, syntax, and diagnostic guidance passed
- Full runner reached 1,459 tests. Five unrelated `scan_installed_apps` cache tests fail identically on untouched `origin/main` in this environment; the focused updater suite is green on the current base.

### DIFF
--- a/lib/manage/update.sh
+++ b/lib/manage/update.sh
@@ -632,7 +632,48 @@ get_latest_commit_from_github() {
     fi
     sha=$(printf '%s\n' "$response" |
         grep '"sha"[[:space:]]*:[[:space:]]*"[0-9a-f]\{40\}"' | head -1 | sed -E 's/.*"sha"[[:space:]]*:[[:space:]]*"([^"]+)".*/\1/') || sha=""
-    echo "$sha"
+    if [[ "$sha" =~ ^[0-9a-f]{40}$ ]]; then
+        printf '%s\n' "$sha"
+        return 0
+    fi
+
+    # The unauthenticated API can be rate-limited even while ordinary GitHub
+    # access works. Keep the fallback bounded and accept only main's exact SHA.
+    if command -v git > /dev/null 2>&1; then
+        # Ignore ambient credential helpers and URL rewrites: this public,
+        # read-only probe must resolve the literal GitHub remote or fail.
+        if response=$(run_with_timeout "$MOLE_TIMEOUT_MEDIUM_PROBE_SEC" \
+            /usr/bin/env \
+            -u GIT_CONFIG_PARAMETERS \
+            -u GIT_EXEC_PATH \
+            -u GIT_DIR \
+            -u GIT_WORK_TREE \
+            -u GIT_COMMON_DIR \
+            -u GIT_CONFIG \
+            -u GIT_PROXY_COMMAND \
+            -u GIT_SSH \
+            -u GIT_SSH_COMMAND \
+            -u GIT_SSL_NO_VERIFY \
+            GIT_TERMINAL_PROMPT=0 \
+            GIT_ASKPASS=/usr/bin/false \
+            SSH_ASKPASS=/usr/bin/false \
+            GIT_CONFIG_NOSYSTEM=1 \
+            GIT_CONFIG_GLOBAL=/dev/null \
+            GIT_CONFIG_COUNT=0 \
+            LC_ALL=C \
+            git -c credential.helper= -c core.askPass=/usr/bin/false \
+            -c protocol.allow=never -c protocol.https.allow=always \
+            -c http.sslVerify=true -C / \
+            ls-remote https://github.com/tw93/mole.git refs/heads/main \
+            2> /dev/null); then
+            sha=$(printf '%s\n' "$response" |
+                awk '$2 == "refs/heads/main" { print $1; exit }')
+        else
+            sha=""
+        fi
+    fi
+    [[ "$sha" =~ ^[0-9a-f]{40}$ ]] || sha=""
+    printf '%s\n' "$sha"
 }
 
 mole_update_message_cache_is_current() {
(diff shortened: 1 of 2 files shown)


iammike2:
> This prevents an unknown HEAD from being treated as evidence that a reinstall is needed.

How is that possible that an unkown HEAD meshes this up?

r266-tech:
“Unknown HEAD” here means that Mole could not resolve the SHA of upstream `main`; it does not mean the installed checkout lost its Git `HEAD`.

Before this patch, the only nightly no-op path required both values to be valid and their first seven characters to match:

```bash
installed_commit=<the saved COMMIT_HASH>
latest_commit=<GitHub API result>

if installed_commit is valid && latest_commit is valid && their 7-char prefixes match; then
    exit as already current
fi
```

If the unauthenticated API was rate-limited or returned an unusable response, `latest_commit` was empty. That made the comparison guard false, so execution fell through into the download/install path, with the same empty value passed as `MOLE_INSTALL_COMMIT`. In other words, the installer received no resolved commit identity from the updater; if upstream resolution failed again, the next ordinary nightly check could repeat the same path.

This change first tries a bounded `git ls-remote` fallback. If neither source yields an exact SHA, an ordinary nightly update now stops before download; `--force` remains the explicit way to reinstall without that proof.

iammike2:
@r266-tech 

this started this morning as before I had no issues as I am on nightly since ages

tw93:
@r266-tech Thanks, this is a well-scoped fix and it shipped in V1.49.1.

Merged as is, with one thing added on top. Your installer stub wrote only the argument and commit logs, and nightly verification on current `main` also requires a per-attempt install receipt plus the resolved commit in `install_channel`. Verification rejected the stub install, self-heal ran, and `mo update --nightly forwards the git-resolved commit to the installer` failed. Your branch base was two commits behind `main`, which is why it passed for you and not here. The stub now installs a working entry script and writes `CHANNEL` / `COMMIT_HASH` / `INSTALL_RECEIPT` like the other nightly stubs, and the four new tests are green on `main`.

One follow-up I deliberately did not fold in. The `git ls-remote` fallback also runs on the background update notice (`check_for_updates`), so a nightly user whose API lookups keep failing now gets an extra `ls-remote` of up to `MOLE_TIMEOUT_MEDIUM_PROBE_SEC` on every `mo` invocation. It is backgrounded with stdin closed and output discarded, so it blocks nothing, but gating the fallback to the explicit update path would be tidier. Happy to take that as a follow-up PR if you are interested.

Worth a direct check on your side:

```bash
mo update
mo update --nightly
```

The second `--nightly` run should settle on `Already on latest nightly, <short sha>` rather than reinstalling.

iammike2:
@tw93 and @r266-tech 

Please look at the latest comment in https://github.com/tw93/Mole/issues/1336

Something was removed from the codebase that caused this 

I found an example in

https://github.com/tw93/Mole/issues/960

Mole version 1.39.1
**Channel: Nightly (969a0fa)**  <----- This was in the version always before YESTERDAY but got somehow removed

iammike2:
and look at this

https://github.com/tw93/Mole/issues/1054

So something was defintely removed from the codebase

r266-tech:
You are right that the older reports show the nightly SHA was present before; that is the expected healthy state. They do not point to the SHA display being removed: the current version command still prints it whenever the install receipt has `COMMIT_HASH`.

The failure was that the old update path could install after upstream `main` could not be resolved. In that case it passed no resolved commit to the source installer, and the source archive has no `.git` directory from which to recover one. The install receipt could therefore contain `CHANNEL=nightly` without `COMMIT_HASH`, which both explains the blank SHA and prevents the next ordinary nightly check from proving it is already current.

#1337 changes that failure path: an ordinary `mo update --nightly` now stops before installing unless it has an exact upstream SHA, and it passes a resolved SHA through to the installer. We still do not know why the first API lookup failed; rate limiting was only one possible trigger.

After installing a release that contains #1337, please run `mo update --nightly` once and then `mo --version`, and keep both outputs. If the update says it could not resolve the latest commit, it correctly installed nothing and the prior blank SHA can remain until a later successful update. A blank SHA after a successful nightly installation, or another reinstall despite matching valid SHAs, would indicate a separate problem worth reopening with those outputs.

iammike2:
@r266-tech 

That rate limiting was then not from my side. Because my check for updates happens only on boot of my MAC once. Homebrew first, Mole later. 

Afterwards homebrew would still update while mole kept failing and failing.

and FYI, that same script *just simple commands like brew update etc etc* is already on my PC since June 2025.

PR #1338, opened 2026-08-02
fix(update): limit the git fallback to explicit nightly updates

r266-tech:
## Summary

- keep background nightly notices on the commit API without starting a Git probe when that request fails
- preserve the bounded, isolated `git ls-remote` fallback for explicit nightly updates and self-heal paths
- add a regression covering both lookup scopes and the background caller

Follow-up to #1337, as requested by @tw93 after that fix merged.

## Test plan

- `MOLE_TEST_NO_AUTH=1 bats tests/update.bats` (31 tests)

### DIFF
--- a/lib/manage/update.sh
+++ b/lib/manage/update.sh
@@ -685,6 +685,7 @@ get_install_receipt() {
 }
 
 get_latest_commit_from_github() {
+    local lookup_scope="${1:-allow-git-fallback}"
     local response=""
     local sha=""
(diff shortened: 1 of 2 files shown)


youxi798:
@r266-tech Merged as-is, thanks.

No changes on my side. Scoping it with an argument instead of a setting was the right call, and keeping a network-bound Git process out of the path that runs behind every `mo` invocation is worth having on its own.

Issue #1339, opened 2026-08-02
[BUG] `mo uninstall` aborts for every app when any installed bundle lacks a readable Info.plist (iOS apps on Apple Silicon)


brun04maral:
## Describe the bug

`mo uninstall` aborts during the preview phase for **every** app, regardless of which app is selected. It exits silently: no error is printed, the shell prompt returns, and `mo history --json` records the session as properly ended (`ended_at` populated) with zero operations.

The root cause is unrelated to the selected app. Before removing anything, the uninstaller scans **all** installed apps to check whether another live install shares the same bundle id. If **any single app** in the scan roots has a missing or unreadable `Info.plist`, the scan returns "incomplete/unknown" and the whole uninstall is aborted.

On Apple Silicon this is easy to trigger, because iOS/iPadOS apps installed from the App Store do not have `Contents/Info.plist` at all.

## Steps to reproduce

1. On an Apple Silicon Mac, install any iPhone/iPad app from the App Store (it lands in `/Applications` with a `Wrapper/` directory instead of `Contents/`).
2. Run `mo uninstall`.
3. Select any app and confirm.
4. The "Selected N apps" list appears, then `Scanning files...`, then the command exits silently without reaching the final confirmation.

With `--debug`, the only clue is:

```
[DEBUG] Could not complete the live same-bundle scan for <selected app>
```

Note the message names the *selected* app, which is misleading: the unreadable bundle is a different app entirely.

## Expected behavior

Either:

- the scan tolerates bundles it cannot read and proceeds, or
- the abort is surfaced to the user with the path of the bundle that could not be read.

Silently returning success is the worst of both, because the user has no way to know what happened or which app is responsible.

## Root cause

`lib/uninstall/batch.sh`, in `_uninstall_collect_live_sibling_candidate` (line numbers from `main` as of 2026-08-02):

```bash
# line 857
[[ "$missing_info_is_unknown" == true ]] && return 2
...
# lines 871-874
if [[ $plist_rc -ne 0 || -z "$app_bundle" ]]; then
    [[ $plist_rc -eq 124 || $plist_rc -ge 128 ]] && return "$plist_rc"
    return 2
fi
```

The `2` propagates up through `uninstall_live_bundle_has_other_install` into the `else` branch of the caller:

```bash
# batch.sh:1295-1306
uninstall_live_bundle_has_other_install "$original_bundle_id" "$app_path" || live_sibling_rc=$?
[...]

PR #1341, opened 2026-08-02
Add app-scan skip list to unblock App leftovers when a bundle is unreadable (#1339)

gcosta74:
## Summary

Fixes the underlying cause of #1339 for `mo clean` as well (the original issue only covers `mo uninstall`): a single `.app` bundle with no readable `Info.plist` currently aborts the *entire* installed-app scan, so `mo clean`'s "App leftovers" section gets skipped for every app on the machine, not just the offending one.

Reproduced with a real Mac Catalyst / wrapped-iOS bundle on Apple Silicon (SecuritySpy Viewer, bundle ID `com.bensoftware.SecuritySpyViewer`, and separately JW Library) — both ship `Info.plist` under `Wrapper/*.app` instead of the usual `Contents/Info.plist`, which the scanner in `lib/clean/apps.sh` doesn't know how to read.

## What this adds

- **`mo clean --skip-app "SomeApp.app"`** — lets the user explicitly vouch for a bundle they know is fine, so the scanner skips just that one app instead of giving up on scanning every other app. Backed by a plain-text config at `~/.config/mole/app_scan_skip` (one bundle basename per line, `#` comments allowed) — same conventions as the existing whitelist config.
- The scan otherwise keeps its current "fail closed" behavior for anything *not* on the list — we still can't tell "unreadable due to an unusual bundle layout" apart from "might actually be an uninstalled orphan," so nothing gets more permissive by default.
- `MOLE_APP_SCAN_DIRS_OVERRIDE` — lets tests point `scan_installed_apps` at a fixture directory instead of the real `/Applications` etc. (same "overridable from tests" convention already used by `_MOLE_BUNDLE_RESOLVER_APP_ROOTS` in `lib/core/bundle_resolver.sh`). Without this, `tests/clean_apps.bats` is unintentionally sensitive to whatever's actually installed on the machine running the suite — I only noticed because a real bundle on my own Mac reproduced this exact bug and made several *pre-existing* tests fail non-deterministically depending on what's in `/Applications`. Confirmed those same failures reproduce on a clean checkout of `main`, unrelated to this change.

## Testing

- `./scripts/check.sh` — formatting, ShellCheck, syntax all pass
- `bats tests/clean_apps.bats` — 2 new tests pass; no regressions vs. `main`
- Manually verified against a real installed app with the Wrapper/-bundle layout: `mo clean --dry-run` went from `Skipped: Unable to scan installed applications (...)` to a completed "App leftovers" scan (`✓ Nothing to clean`) after `mo clean --skip-app`.

## Notes for maintainers

Happy to also fold the fallback `Wrapper/*.app/Info.plist` path-resolution approach suggested in #1339 into this PR if you'd rather fix bundle detection at the source instead of (or in addition to) an explicit skip list — wanted to keep this PR's scope to one approach first. The skip list has the advantage of working for *any* reason a bundle might be unreadable (not just the Wrapper/ case), and gives the user an explicit, auditable escape hatch rather than silently trusting more bundle shapes.

### DIFF
--- a/bin/clean.sh
+++ b/bin/clean.sh
@@ -1973,6 +1973,20 @@ main() {
                 manage_whitelist "clean"
                 exit 0
                 ;;
+            "--skip-app")
+                shift
+                if [[ $# -eq 0 ]]; then
+                    echo "Missing app name for --skip-app (e.g. \"SecuritySpy.app\")" >&2
+                    exit 1
+                fi
+                if ! add_app_scan_skip "$1"; then
+                    echo "Failed to update $APP_SCAN_SKIP_CONFIG" >&2
+                    exit 1
+                fi
+                echo "Added \"$1\" to the app-scan skip list ($APP_SCAN_SKIP_CONFIG)."
+                echo "It will be excluded from installed-app scanning on the next 'mo clean' run."
+                exit 0
+                ;;
             "--select" | "--categories" | "--exclude")
                 echo "mo clean $1 was removed in this release." >&2
                 echo "Use 'mo clean --dry-run' to preview cleanup and 'mo clean --whitelist' to protect paths." >&2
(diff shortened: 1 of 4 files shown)


tw93:
@gcosta74 Thanks for this, and sorry for the slow reply.

The path you found is real and I had missed it: clean's installed-app scan made the same `Contents/Info.plist` assumption, and one unreadable bundle took the whole App leftovers section with it.

I've fixed it a different way, so I'm going to close this. Those bundles aren't really unreadable: iOS apps on Apple Silicon keep their plist under `Wrapper/<name>.app`, and the vendor launchers just have no `CFBundleIdentifier`, which is an answer rather than a failure. Reading both makes them scan normally, with nothing for you to declare. Mole tries hard not to add a setting when a default fix exists.

Closing, but this is the reason I knew to look at that path at all.

Issue #1342, opened 2026-08-02
[BUG] `mo clean` silently crashes with exit status `124`


nickdos:
## Describe the bug

Running `mo clean` runs for 30 seconds and finishes without showing any disk space saved. Inspecting the exit status, shows a value of 124.

## Steps to reproduce

1. Run command: `mo clean`
2. Type sudo password
3. See output

```bash
% mo clean    
Clean Your Mac

◎ Use --dry-run to preview, --whitelist to manage protected paths
✓ Admin access granted

⚙ Apple Silicon | Free space: 72.24GB
✓ Whitelist: 21 core patterns active

➤ System
  ☞ Time Machine local snapshots · 24 (review: tmutil listlocalsnapshots /)

➤ User essentials
% echo $? 
124
```

## Expected behavior

I expect to see [output in README](https://github.com/tw93/Mole#deep-system-cleanup) for `mo clean`.

## Debug logs

Please run the command with `--debug` flag and paste the output here:

```bash
mo clean --debug
```

<details>
<summary>Debug output</summary>

```text
GH errors with `Body can not be longer than 65536 characters` - attached 2 files: shell output and generated log file.
```

</details>

## Environment

Please run `mo update` to ensure you are on the latest version, then paste the output of `mo --version` below:

```text
Mole version 1.49.1
macOS: 26.5.2
Architecture: arm64
Kernel: 25.5.0
SIP: Enabled
Disk Free: 70.76GB
Install: Manual
Shell: /bin/zsh
```

## Additional context

Running with --debug a second time took a lot longer and produced a lot more output so seems odd I got different results. 

[20260803 Mole clean debug output.txt](https://github.com/user-attachments/files/30640326/20260803.Mole.clean.debug.output.txt)

[mole_debug_session.log.txt](https://github.com/user-attachments/files/30640338/mole_debug_session.log.txt)

nickdos:
I suspect a Time Machine backup was running at the time and this may have caused the issue. Re-running just regular clean option now proceeds past the point where it failed the first time. I now see:

```bash
======================================================================
Cleanup complete
Tracked cleanup: 340.6MB | Items cleaned: 49 | Categories: 15
Free space: 68.20GB (-210.4MB)
Skipped while active: Chrome
======================================================================
```

Yuxin-Qiao:
Root cause found and fix submitted in #1345.

[...]

PR #1345, opened 2026-08-03
fix(clean): print final summary when a cleanup step times out (#1342)

Yuxin-Qiao:
## Summary

When a `mo clean` scan or size check hits its internal safety timeout (`MOLE_TIMEOUT_DISK_VERIFY_SEC`, default 30s), the whole run aborts with exit status `124` and **no summary at all** - the output just stops mid-run. This was reported in #1342 (exit 124 after ~30s, output stopping at "User essentials", no "Cleanup complete" block).

The sizing timeout is intentional cancellation semantics (see `get_path_size_kb` in `lib/core/file_ops.sh`), so this PR keeps the cancellation and the exit code, but makes the outcome comprehensible:

- Cleanup sections now run inside a `run_clean_sections()` wrapper, so a cancelled step (timeout `124` or signal `128+`) stops the remaining sections but falls through to the final summary.
- The summary heading becomes `Cleanup cancelled` / `Cleanup interrupted` (or `Dry run ...` variants), with an explicit reason line, e.g.:
  `◎ Cancelled: a scan or size check timed out (exit 124). Remaining cleanup was skipped.`
- Normal runs are unchanged: `Cleanup complete`, exit 0.

## Reproduction

Issue #1342: `mo clean` runs ~30s, exits `124`, no summary. Root cause: with a Time Machine backup (or any disk I/O stall) running, `du`/`stat` sizing of `~/Library/Caches/*` exceeds the 30s `run_with_timeout` deadline; `_safe_clean_impl` returns 124, which propagated through `clean_user_essentials` -> `perform_cleanup` and exited before the summary block.

Deterministic repro used during development (fake HOME + slow `du` in PATH, no real files touched):

```bash
mkdir -p /tmp/mr/Library/Caches /tmp/mrbin
for i in $(seq -w 1 30); do mkdir /tmp/mr/Library/Caches/cache$i; done
printf '#!/bin/bash\nsleep 40\n' > /tmp/mrbin/du && chmod +x /tmp/mrbin/du
PATH=/tmp/mrbin:$PATH HOME=/tmp/mr MOLE_TEST_NO_AUTH=1 ./mo clean; echo $?   # 124 after ~30s
```

Before: output ends at `➤ User essentials`, exit `124`, no summary.
After: same exit `124`, but now:

```
Cleanup cancelled
System was already clean; no additional space freed.
Free space: 22.77GB (+11.3MB)
```

## Tests

- New `tests/clean_summary_cancel.bats`: end-to-end-ish regression for #1342 (cancel/success headings and exit codes via stubbed sections), 3 cases.
- `bats tests/clean_core.bats tests/clean_user_core.bats tests/file_ops_size.bats tests/regression.bats tests/core_common.bats tests/clean_summary_cancel.bats` - all 187 tests pass.
- `shellcheck bin/clean.sh`, `bash -n`, `shfmt -d`, `scripts/audit_function_duplication.py` - all clean.

### DIFF
--- a/bin/clean.sh
+++ b/bin/clean.sh
@@ -1675,116 +1675,123 @@ perform_cleanup() {
         return 0
     }
 
-    if [[ -n "$EXTERNAL_VOLUME_TARGET" ]]; then
-        start_section "External volume"
-        _run_cleanup_step clean_external_volume_target "$EXTERNAL_VOLUME_TARGET" || return $?
-        end_section
-    else
-        # ===== 1. System =====
-        if [[ "$SYSTEM_CLEAN" == "true" ]]; then
-            start_section "System"
-            _run_cleanup_step clean_deep_system || return $?
-            _run_cleanup_step clean_local_snapshots || return $?
+    local cleanup_cancel_rc=0
+    # Sections run inside a function so a cancelled step (timeout/signal,
+    # exit 124+) stops the remaining sections but still falls through to the
+    # final summary instead of returning from perform_cleanup with no output.
+    run_clean_sections() {
+        if [[ -n "$EXTERNAL_VOLUME_TARGET" ]]; then
+            start_section "External volume"
+            _run_cleanup_step clean_external_volume_target "$EXTERNAL_VOLUME_TARGET" || return $?
+            end_section
+        else
+            # ===== 1. System =====
+            if [[ "$SYSTEM_CLEAN" == "true" ]]; then
+                start_section "System"
+                _run_cleanup_step clean_deep_system || return $?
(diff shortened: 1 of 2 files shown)


youxi798:
@Yuxin-Qiao Merged, thanks.

I changed nothing in the code. I did note in the merge message that this makes the stop visible rather than stopping it from happening: one slow directory can still end the run, and narrowing that to skip the single target is a separate change, so the issue stays open.

One thing for next time: most of the diff was the re-indent from wrapping the sections in a function, which buried the dozen lines that actually changed. Splitting "extract function" and "change behaviour" into two commits makes that much easier to read.

PR #1347, opened 2026-08-03
fix(install): derive source temp dir from TMPDIR so cleanup accepts it

### LINKED ISSUE #1343: [BUG] install.sh exits 1 after a successful install: safe_rm/mktemp disagree about the temp root when TMPDIR is unset (regression in V1.49.1)
### What happens

Since V1.49.1, `install.sh` exits `1` **after a fully successful install**. The last thing it prints is:

```
◎ Mole installed successfully, version 1.49.1

☻ safe_rm: refusing to remove non-temp path: /var/folders/df/djsxfhc17x95674wsm_g8s980000gn/T/tmp.gNQE2Op3RR
```

Everything installs correctly — the non-zero exit comes purely from cleanup. Any caller that checks the exit status (package managers, CI, `set -e` wrappers) treats the install as failed.

### Root cause

Two pieces that disagree about where the temp dir lives:

`fetch_source` creates it with a bare `mktemp -d`:

```bash
tmp="$(mktemp -d)"
INSTALL_SOURCE_TMP="$tmp"
```

`safe_rm` then gates removal on `TMPDIR`:

```bash
tmp_root="${TMPDIR:-/tmp}"
...
case "$target" in
    "$tmp_root"/* | /tmp/*) ;;
    *)
        log_error "safe_rm: refusing to remove non-temp path: $target"
        return 1
```

On macOS a bare `mktemp -d` (no template, no `-t`) **ignores `TMPDIR`** and always uses `confstr(_CS_DARWIN_USER_TEMP_DIR)`, i.e. `/var/folders/…/T/`:

```console
$ TMPDIR=/tmp mktemp -d
/var/folders/8y/54_37j71057dsjqh56rkxrz40000gn/T/tmp.sUoPUlRMhK
```

So whenever `TMPDIR` is unset or points anywhere other than the Darwin per-user temp dir, `tmp_root` is `/tmp`, the target is `/var/folders/…`, and `safe_rm` refuses and returns `1`. Since `cleanup_installer` is the `EXIT` trap and `safe_rm` is its last command, that `1` becomes the script's exit status.

### Why this is new in V1.49.1

V1.48.1 had the same `safe_rm` logic, but `cleanup_tmp` closed over a function-local `tmp`. By the time the `EXIT` trap fired, `${tmp:-}` was empty and cleanup returned early, so `safe_rm` never ran. V1.49.1 moved the path to a global `INSTALL_SOURCE_TMP` with a top-level `trap 'cleanup_installer' EXIT`, which made the cleanup actually execute and exposed the latent mismatch. (The V1.48.1 behaviour also means temp dirs were being leaked rather than cleaned.)

### Reproduction

```console
$ env -u TMPDIR bash install.sh --prefix "$HOME/mole-repro-test" --config "$HOME/mole-repro-cfg"
...
◎ Mole installed successfully, version 1.49.1
☻ safe_rm: refusing to remove non-temp path: /var/folders/.../T/tmp.XXXXXXXX
$ echo $?
1
```

Use a prefix under `$HOME`. A prefix under `/tmp` does **not** reproduce this — `install_lock_has_unsafe_ancestor` walks up to `/tmp`, which is a symlink to `private/tmp` on macOS, and rejects the install with `Refusing a privileged install through /tmp` before it ever gets far enough to install. `/private/tmp` fails too, being world-writable (`1777`).

Confirmed on macOS (Apple Silicon and Intel):

| Version | `TMPDIR` | Exit |
| --- | --- | --- |
| V1.48.1 | unset | 0 |
| V1.49.1 | unset | 1 |
| V1.49.1 | set to `$(getconf DARWIN_USER_TEMP_DIR)` | 0 |

This is easy to hit from package managers: Homebrew filters `TMPDIR` out of the environment it passes to install scripts, so `TMPDIR` is always unset there and the failure is deterministic.

### Suggested fix

Make the two agree. Either have `mktemp` honour `TMPDIR`:

```bash
tmp="$(mktemp -d "${TMPDIR:-/tmp}/mole.XXXXXX")"
```

or have `safe_rm` accept the real Darwin temp root as well, e.g. by also allowing `$(getconf DARWIN_USER_TEMP_DIR)` alongside `$TMPDIR` and `/tmp`.

The first is probably cleaner — it keeps a single notion of "temp root" that both sides derive from.

It may also be worth not letting cleanup failures decide the exit status, so a successful install can't be reported as a failure by its own teardown.

Yuxin-Qiao:
## Summary

Fixes #1343. On macOS, `install.sh` exits `1` after a fully successful install whenever `TMPDIR` is unset. `resolve_source_dir` creates the source temp dir with a bare `mktemp -d`, which on macOS ignores `TMPDIR` and returns the Darwin per-user temp root (`/var/folders/.../T/`). `safe_rm` only accepts paths under `${TMPDIR:-/tmp}`, so the EXIT-trap cleanup refuses to remove the source dir and its non-zero status becomes the script's exit status. This is a regression in V1.49.1 (the V1.48.1 cleanup early-returned and leaked the dir instead).

## Fix

Create the temp dir with an explicit template derived from the same root `safe_rm` uses, keeping a single notion of "temp root" for both sides:

```bash
tmp="$(mktemp -d "${TMPDIR:-/tmp}/mole.XXXXXX")"
```

This matches the existing checksums-file call (`mktemp "${TMPDIR:-/tmp}/mole-checksums.XXXXXX"`) and is independent of macOS version: with the template, `mktemp` honours the explicit path whether `TMPDIR` is unset, `/tmp`, or the Darwin user temp dir. The `update` path uses the same `resolve_source_dir`, so one change covers both.

## Tests

- Behavioral regression: evaluates `install.sh`'s own source-download `mktemp` lines and `cleanup_installer` with `TMPDIR` unset, then asserts the dir is removed and `safe_rm` did not refuse. Before this fix on macOS it fails with the exact `safe_rm: refusing to remove non-temp path` error.
- Source guard: pins the `${TMPDIR:-/tmp}` template in `install.sh`.
- `bats tests/install_checksum.bats` passes (21 tests); `./scripts/check.sh` passes.

## Reproduction

```console
$ env -u TMPDIR bash install.sh --prefix "$HOME/mole-repro-test" --config "$HOME/mole-repro-cfg"
◎ Mole installed successfully, version 1.49.1
☻ safe_rm: refusing to remove non-temp path: /var/folders/.../T/tmp.89JgkkmD95
$ echo $?
1
```

With the fix the same run exits `0` and the source temp dir is removed.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -622,7 +622,7 @@ resolve_source_dir() {
     fi
 
     local tmp
-    tmp="$(mktemp -d)"
+    tmp="$(mktemp -d "${TMPDIR:-/tmp}/mole.XXXXXX")"
     INSTALL_SOURCE_TMP="$tmp"
 
     local branch="${MOLE_VERSION:-}"
--- a/tests/install_checksum.bats
+++ b/tests/install_checksum.bats
@@ -710,6 +710,44 @@ EOF
 	[[ "$output" != *"safe_rm: refusing to remove non-temp path"* ]] || return 1
 }
 
+@test "installer source temp stays removable by safe_rm when TMPDIR is unset" {
+	run env HOME="$HOME" PROJECT_ROOT="$PROJECT_ROOT" /bin/bash --noprofile --norc << 'EOF'
+set -euo pipefail
+unset TMPDIR
+log_error() { printf 'ERROR:%s\n' "$*"; }
+stop_line_spinner() { :; }
+release_install_lock() { :; }
+
+eval "$(sed -n '/^safe_rm() {/,/^}/p' "$PROJECT_ROOT/install.sh")"
+eval "$(sed -n '/^cleanup_installer() {/,/^}/p' "$PROJECT_ROOT/install.sh")"
+
+# Evaluate install.sh's own source-download mktemp lines with TMPDIR unset,
+# then run the same EXIT-trap cleanup path against the created directory.
+tmp_lines="$(sed -n '/mktemp -d/{p;n;p;q;}' "$PROJECT_ROOT/install.sh" | sed 's/^[[:space:]]*//')"
+eval "$tmp_lines"
+source_tmp="$INSTALL_SOURCE_TMP"
+printf 'downloaded source\n' > "$source_tmp/payload"
+cleanup_installer
+
+[[ -z "$INSTALL_SOURCE_TMP" ]] || exit 1
+[[ ! -e "$source_tmp" ]] || exit 1
+EOF
+
+	[ "$status" -eq 0 ] || {
+		echo "$output"
+		return 1
+	}
+	[[ "$output" != *"safe_rm: refusing to remove non-temp path"* ]] || return 1
+}
+
+@test "source download mktemp template derives from TMPDIR" {
+	run grep -qE 'mktemp -d "\$\{TMPDIR:-/tmp\}/mole\.XXXXXX"' "$PROJECT_ROOT/install.sh"
+	[ "$status" -eq 0 ] || {
+		echo "$output"
+		return 1
+	}
+}
+
 @test "standalone installer serializes writers with the stable install lock" {
 	run env HOME="$HOME" PROJECT_ROOT="$PROJECT_ROOT" /bin/bash --noprofile --norc << 'EOF'
 set -euo pipefail


youxi798:
@Yuxin-Qiao Merged, thanks. Two things I changed, both worth knowing about.

The same one-line fix had already landed here independently before I got to this, so the merge carries no new code. I merged it anyway rather than closing it, because the diagnosis and the reproduction arrived first and the tests are worth keeping.

I did edit one of your tests. It scraped install.sh with `sed -n '/mktemp -d/{p;n;p;q;}'`, and once a comment above the line also mentioned `mktemp -d`, it matched the comment instead of the assignment, eval'd nothing, and failed. I anchored it on the assignment and made it error out loudly if it matches nothing. Worth avoiding generally: pinning a phrase that can also appear in prose turns a scraping miss into what looks like a real failure.

PR #1361, opened 2026-08-03
AudioCopy.app triggers edge-case fail closed

joemcmahon:
AudioCopy.app has a strange internal structure (a broken WrappedBundle symlink) that causes Mole to fail closed while looking for orphaned app data. This fix skips over apps for which we simply can't find a plist path at all.

## Summary

- Checks to see if we can actually find Info.plist at all. Skips if not.

## Safety Review

Will allow cleanup behavior to proceed that otherwise would be blocked, but does not add any ADDITIONAL cleanup.

Affects symlink handling (detects invalid symlinks during lookup and skips affected apps altogether).

## Tests

Added new test that creates a fake version of the buggy app that breaks the scan, and a good app. Verifies that the scan does not break if the buggy app is present.

Ran a full cleanup; nothing extra was cleaned up, but the scan continued successfully through the `scan_installed_apps` step without skipping it.

## Safety-related changes

- None.

### DIFF
--- a/lib/clean/apps.sh
+++ b/lib/clean/apps.sh
@@ -245,10 +245,14 @@ scan_installed_apps() {
                     # Vendor uninstallers and Steam launchers ship those, and a
                     # bundle with no id owns no bundle-id-named leftovers, so
                     # leaving it out of the list cannot invent an orphan.
+                    # No plist at all (absent Contents/ and no Wrapper/ found,
+                    # e.g. an iOS app whose WrappedBundle symlink is dangling)
+                    # is the same: the bundle cannot own any bundle-id-named
+                    # leftovers and cannot be mistaken for one.
                     # Anything that will not parse still fails the scan closed,
                     # because there the id may exist and simply be unreadable,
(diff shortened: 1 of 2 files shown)


tw93:
@joemcmahon Thanks for the sharp diagnosis, the dangling `WrappedBundle` symlink was exactly it, and one such app disabling the whole App-leftovers scan is the bug worth fixing.

I landed the fix on `main` with a slightly narrower condition than the PR. Treating every app with no plist as skippable would relax a fail-closed boundary that an existing test pins (an app whose identity is merely unreadable must still fail the scan closed, or a live app's data could look like an orphan). So the skip now requires the specific evidence you found: a dangling `WrappedBundle` symlink paired with no readable plist anywhere. That covers AudioCopy.app while keeping the fail-closed contract for the unreadable case, with a regression test for each side.

Closing this in favor of that commit, with credit to you in the message. `mo update --nightly` picks it up now. Reopen if any app still fails the scan closed for you.

PR #1369, opened 2026-08-04
fix(optimize): resolve the notification database on macOS 15+ (#1368)

### LINKED ISSUE #1368: [BUG] Notification cleanup uses the pre-Sequoia database path
## Describe the bug

On macOS 15 and later, `mo optimize` reports:

```text
➤ Notifications
  ✓ Notification Center database not found
```

Mole currently checks:

```text
$(getconf DARWIN_USER_DIR)/com.apple.notificationcenter/db2/db
```

The active notification database moved in macOS 15 to:

```text
$HOME/Library/Group Containers/group.com.apple.usernoted/db2/db
```

On the affected system, the old directory contains only stale `db-shm` and empty `db-wal` files from 2024. `lsof` confirms that `usernoted` has the group-container database open.

As a result, the Notifications task returns `unchanged` without inspecting the active database. The success icon makes the message look like a harmless absence rather than an unsupported path.

## Steps to reproduce

1. Use macOS 15 or later with an active notification database.
2. Confirm the current database exists:

   ```bash
   ls -l "$HOME/Library/Group Containers/group.com.apple.usernoted/db2/db"
   lsof "$HOME/Library/Group Containers/group.com.apple.usernoted/db2/db"
   ```

3. Run:

   ```bash
   mo optimize --dry-run --debug
   ```

4. Observe `Notification Center database not found`.
5. Confirm the path hard-coded by Mole does not contain a main `db` file:

   ```bash
   ls -l "$(getconf DARWIN_USER_DIR)/com.apple.notificationcenter/db2/db"
   ```

## Expected behavior

Mole should resolve the notification database by macOS version and inspect the active database:

- Prefer `$HOME/Library/Group Containers/group.com.apple.usernoted/db2/db` on macOS 15 and later.
- Retain the Darwin-user-directory path only as a fallback for older systems.
- Validate the expected SQLite schema before deleting records.
- Preserve the existing size threshold and skip safely if the database is busy or locked.
- Report the selected path and size in debug output.
- If no supported database exists, report the attempted paths as unavailable rather than successful/unchanged.

On the affected system, the active database is only 339,968 bytes, so the corrected result should be `Notification Center database is healthy`, not `database not found`.

## Debug logs

```bash
mo optimize --dry-run --debug
```

<details>
<summary>Relevant output</summary>

```text
➤ Notifications
  ✓ Notification Center database not found
```

</details>

## Environment

```text
Mole version 1.49.2
macOS: 27.0
Architecture: arm64
Kernel: 27.0.0
SIP: Enabled
Disk Free: 58.39GB
Install: Homebrew
Shell: /bin/zsh
```

## Additional context

- The path change was identified during macOS 15 development in July 2024: <https://github.com/ydkhatri/mac_apt/pull/119>
- A macOS forensics parser uses the group-container path for macOS 15 and later: <https://github.com/ydkhatri/mac_apt/blob/6b1cd4e9301aab0da922d0a6b3f6900e945b7acb/plugins/notifications.py#L30-L43>
- Mole added this task in April 2026 with the older path: <https://github.com/tw93/Mole/commit/02f3a4d43212ade8ba6880cea0f548931d3a69cf>
- Current upstream code still uses the old location: <https://github.com/tw93/Mole/blob/7b29d15bca17257aa1bda939e139f5f3cc1d71f8/lib/optimize/tasks.sh#L1420-L1443>
- The same message appears incidentally in debug logs from unrelated issues #718 and #1300, suggesting the mismatch is not host-specific:
  - <https://github.com/tw93/Mole/issues/718>
  - <https://github.com/tw93/Mole/issues/1300>

Yuxin-Qiao:
## Summary

Fix #1368: resolve the active NotificationCenter database by macOS version instead of always using the pre-Sequoia Darwin-user-directory path.

- Prefer `$HOME/<local-path>` on macOS 15 and later.
- Keep `$(getconf DARWIN_USER_DIR)/com.apple.notificationcenter/db2/db` as the fallback for older systems.
- Verify the `record` table exists before running `DELETE` + `VACUUM`; a probe failure is reported as busy/locked, a missing table as an unexpected schema.
- Report the selected database path and size in `--debug` output.
- When no supported database exists, report the attempted paths and mark the task `unavailable` instead of a green "not found" `unchanged`.

The size threshold (50MB) and the 30-day delete window are unchanged.

## Real-system reproduction

macOS 27 (arm64). Before the fix, the old path has no main `db` file and `mo optimize` reports `Notification Center database not found`:

```console
$ ls -l "$(getconf DARWIN_USER_DIR)/com.apple.notificationcenter/db2"
ls: /com.apple.notificationcenter/db2: No such file or directory

$ ls -l "$HOME/<local-path>"
-rw-r--r--  1 <user>  staff  1429504 Aug  4 20:10 .../<local-path>

$ sqlite3 "$HOME/<local-path>" \
    "SELECT name FROM sqlite_master WHERE type='table' AND name='record';"
record
```

After the fix, the same machine resolves the group-container database and reports it healthy (below the 50MB threshold), with the path and size in debug output:

```console
$ MO_DEBUG=1 MOLE_DRY_RUN=1 bash -c 'source lib/core/common.sh; source lib/optimize/tasks.sh; execute_optimization notification_cleanup'
[DEBUG] Database: <local-path>
[DEBUG] Size: 1.4MB
  → Notification Center database is healthy (1.4MB)
```

## Tests

- `bats tests/optimize_db.bats` (12/12, including 4 new notification cases: macOS 15+ group-container path, pre-15 legacy fallback, no-database unavailable with attempted paths, missing `record` table)
- `bats tests/optimize.bats tests/optimize_probe_outcomes.bats tests/optimize_catalog.bats tests/optimize_outcomes.bats tests/optimize_summary.bats tests/optimize_removal_outcomes.bats`
- `shellcheck mole install.sh bin/*.sh lib/*/*.sh scripts/*.sh`
- `bash -n` on changed files

### DIFF
--- a/lib/optimize/tasks.sh
+++ b/lib/optimize/tasks.sh
@@ -1419,14 +1419,53 @@ opt_shared_file_list_repair() {
 
 # Clean old delivered notifications from NotificationCenter database.
 opt_notification_cleanup() {
-    local nc_db_dir
-    nc_db_dir="$(getconf DARWIN_USER_DIR 2> /dev/null || true)/com.apple.notificationcenter/db2"
-    local nc_db="$nc_db_dir/db"
+    if [[ "${MO_DEBUG:-}" == "1" ]]; then
+        debug_operation_start "Notification Center Cleanup" "Remove delivered notifications older than 30 days from the active NotificationCenter database"
+        debug_operation_detail "Method" "DELETE + VACUUM on the active SQLite database (group-container path on macOS 15+)"
+        debug_risk_level "LOW" "Only delivered notifications older than 30 days are removed"
+    fi
 
-    if [[ ! -f "$nc_db" ]]; then
-        opt_msg "Notification Center database not found"
-        optimize_task_result "$MOLE_OPTIMIZE_OUTCOME_UNCHANGED"
+    # The notification database moved in macOS 15 to the usernoted group
+    # container; the Darwin-user-directory path only holds stale files there.
+    local os_major=""
+    os_major=$(sw_vers -productVersion 2> /dev/null | cut -d. -f1 || true)
+
+    local group_db="$HOME/Library/Group Containers/group.com.apple.usernoted/db2/db"
+    local legacy_dir=""
+    local darwin_user_dir=""
+    darwin_user_dir=$(getconf DARWIN_USER_DIR 2> /dev/null || true)
+    if [[ -n "$darwin_user_dir" ]]; then
+        legacy_dir="$darwin_user_dir/com.apple.notificationcenter/db2/db"
+    fi
+
+    local -a db_candidates=()
+    if [[ "$os_major" =~ ^[0-9]+$ && "$os_major" -ge 15 ]]; then
+        db_candidates=("$group_db")
+        [[ -n "$legacy_dir" ]] && db_candidates+=("$legacy_dir")
+    else
+        [[ -n "$legacy_dir" ]] && db_candidates+=("$legacy_dir")
+        db_candidates+=("$group_db")
+    fi
+
+    local nc_db=""
+    local -a attempted_paths=()
+    local candidate=""
+    for candidate in "${db_candidates[@]}"; do
+        attempted_paths+=("$candidate")
+        if [[ -f "$candidate" ]]; then
+            nc_db="$candidate"
+            break
+        fi
+    done
+
+    if [[ -z "$nc_db" ]]; then
+        echo -e "  ${YELLOW}${ICON_WARNING}${NC} Notification Center database not found (tried: ${attempted_paths[*]})"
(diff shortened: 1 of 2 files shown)


tw93:
@Yuxin-Qiao Closing this without merging, and thank you for the careful PR.

The path bug #1368 is already fixed on `main`: Mole prefers the usernoted group-container database when it exists, falls back to the Darwin-user-directory layout, and reports unavailable instead of a success-style "not found".

```bash
mo update --nightly
```

The branch conflicts with that landing, and the existence-based resolver already covers the live macOS 15+ case. Your extra schema probe on the `record` table is a good hardening idea; if I add it as a follow-up I will credit this PR.

tw93:
Superseded by the #1368 fix already on main. See comment above.

PR #1350, opened 2026-08-03
fix(clean): skip a Mail Downloads dir whose sizing times out (#1344)

Yuxin-Qiao:
## Summary

Fixes #1344. When `du` stalls on a Mail Downloads directory (the reporter's `~/Library/Containers/com.apple.mail/.../Mail Downloads` returns `Operation not permitted` on macOS 26, even to root), the 30s sizing probe times out and `_clean_mail_downloads` propagates exit `124` up through `clean_user_essentials`, ending the whole `mo clean` run at "User essentials". On V1.49.1 that exit is silent, which is the reported "silently crashes" behavior (same family as #1342).

This is the follow-up the #1345 merge note asked for: skip the single slow target instead of letting it end the run.

## Change

- In `_clean_mail_downloads`, a directory sizing timeout (`124`) now stops the spinner, prints a warning (`Mail Downloads · skipped (sizing timed out)`), and moves on to the next directory instead of recording a cancellation and returning the status.
- A per-attachment sizing timeout now skips that file (`debug_log` only) instead of aborting the whole mail cleanup.
- Signal-class statuses (`>=128`, e.g. Ctrl-C `130`) keep their existing cancellation semantics: `_mole_record_clean_cancellation` + `return`.

## Reproduction (mechanism, macOS 27 x86 host via a stalled `du` shim)

```console
$ env HOME=<fake-home-with-unreadable-Mail-Downloads> PATH=<shim>$PATH MOLE_TIMEOUT_DISK_VERIFY_SEC=2 bash bin/clean.sh --debug
```

Before, V1.49.1:

```text
➤ User essentials
...
$ echo $?
124
```

(no summary, silent exit - matches the issue)

Before, current main: same exit `124` but with the `Cleanup cancelled` summary from #1345; the run still stops at User essentials.

After this PR:

```text
➤ User essentials
  ◎ Mail Downloads · skipped (sizing timed out)
➤ App caches
...
Cleanup complete
$ echo $?
0
```

The fast-fail variant in the issue (du exits `1` with no output, the `(returned: )` debug line) already resolves to a zero-size skip and is covered by existing behavior; the timeout variant is what ends the run and is fixed here.

## Tests

New `tests/clean_mail_downloads.bats` (5 tests):
- dir sizing timeout skips the target and keeps the run going
- one dir timing out does not stop cleanup of the other dir
- attachment sizing timeout skips the file without aborting
- signal (130) still cancels the run
[...]

PR #1377, opened 2026-08-05
fix(clean): keep cleaning when a cache size check times out (#1374)

### LINKED ISSUE #1374: clean: a slow ~/Library/Caches size check (exit 124) aborts the whole run
Follow-up to #1345 (summary on timeout) and #1350 (skip a Mail Downloads dir whose sizing times out).

On a large `~/Library/Caches` (mine: ~12 GB, ~173k files), `mo clean` cancels the entire run the first time a single item takes longer than the 30s size-check budget:

```
➤ User essentials
Cleanup cancelled
```

#1350 fixed this for Mail Downloads, but the main cache path (`safe_clean ~/Library/Caches/*` → `_safe_clean_impl`) still treats a sizing timeout as fatal.

The size is only used for the "freed" total; the delete set is chosen separately. So a stalled `du` shouldn't stop the run — same reasoning as #1350.

**Fix I'd like to PR:** in `_safe_clean_impl`, treat a `124` sizing timeout as "size unknown, keep cleaning" (signals `>=128` and real `rm` timeouts stay fatal, unchanged), and add one summary line so it isn't silent:

> Some items exceeded the 30s size-check budget and were counted as 0, so the total is under-reported. Raise `MOLE_TIMEOUT_DISK_VERIFY_SEC` to measure them.

Branch is ready with bats tests. One thing to check first: the project-cache sweep (`process_project_cache_matches`) has the same shape but intentionally stops on timeout, so I left it alone — happy to include it if you'd prefer.

mole 1.49.2, macOS 15.5, Apple Silicon.

Yuxin-Qiao:
## Summary

Fixes #1374: a single `~/Library/Caches/*` item whose size check exceeds `MOLE_TIMEOUT_DISK_VERIFY_SEC` (exit 124) cancelled the whole `mo clean` run. The size is only used for the "freed" total; the delete set is chosen separately, so a stalled sizing scan should not stop cleanup (same reasoning as #1350).

Changes in `_safe_clean_impl` (bin/clean.sh):
- All three sizing paths (bulk `stat`, parallel per-item workers, small-batch) now treat a sizing `124` as "size unknown": the item is still cleaned and counted as 0 KB.
- Signals (`>=128`) and real `rm`/guard timeouts stay fatal, unchanged.
- A run that under-counted sizes now prints one summary line:
  `Some items exceeded the Ns size-check budget and were counted as 0, so the total is under-reported. Raise MOLE_TIMEOUT_DISK_VERIFY_SEC to measure them.`
- `process_project_cache_matches` intentionally keeps its stop-on-timeout behavior, unchanged.

## Reproduction

Harness: throwaway `HOME` + a PATH `du` stub that stalls one cache dir past the budget, `MOLE_TEST_NO_AUTH=1`, no real files touched (script: `repro-scripts/repro-1374-cache-size-timeout.sh` in the mole-fix workspace, not part of this PR).

| case | version | exit | result |
|---|---|---|---|
| control, no slow item | main | 0 | `Cleanup complete` |
| one slow cache item | main (before) | 124 | `Cleanup cancelled` + all remaining sections skipped |
| one slow cache item | main (after) | 0 | `Cleanup complete`, all 9 items cleaned, note printed |
| one slow cache item, default 30s budget | main (after) | 0 | same |
| one slow cache item | V1.49.2 (released) | 124 | unchanged bug |

## Tests

- `tests/clean_core.bats`: updated the old "size timeout cancels deletion" test to the new semantics; added a parallel-worker timeout test (one of five workers times out, all five items are removed, `MOLE_CLEAN_CANCEL_STATUS` stays 0).
- `tests/clean_summary_cancel.bats`: added a full-`perform_cleanup` test asserting the run completes and prints the under-reported note.
- `make check` passes (go vet, shellcheck, syntax, duplication audit).
- Bats: clean_core 53/53, clean_summary_cancel 4/4, clean_mail_downloads 6/6, plus 14 related clean/size/safe-function files all green. Full `make test`: 1550 tests, 1 failure, 15 skipped; the failure is a pre-existing `core_timeout.bats` tty/perl-fallback test that also fails on unmodified `main` in this environment.

### DIFF
--- a/bin/clean.sh
+++ b/bin/clean.sh
@@ -1071,8 +1071,12 @@ _safe_clean_impl() {
             run_with_timeout "$MOLE_TIMEOUT_DISK_VERIFY_SEC" \
                 stat -f%z "${existing_paths[@]}" < /dev/null \
                 > "$bulk_stat_file" 2> /dev/null || bulk_stat_rc=$?
-            if [[ $bulk_stat_rc -eq 124 || $bulk_stat_rc -ge 128 ]]; then
+            if [[ $bulk_stat_rc -ge 128 ]]; then
                 cleanup_interrupt_rc=$bulk_stat_rc
+            elif [[ $bulk_stat_rc -eq 124 ]]; then
+                # The size is only used for the freed total; a stalled stat
+                # must not cancel the delete set. Sizes are already 0 here.
+                MOLE_CLEAN_SIZING_TIMEOUTS=$((${MOLE_CLEAN_SIZING_TIMEOUTS:-0} + 1))
             fi
             while IFS= read -r _bytes; do
                 [[ "$_bytes" =~ ^[0-9]+$ ]] || _bytes=0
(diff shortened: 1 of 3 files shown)


tw93:
@Yuxin-Qiao The sizing-timeout half is included in #1385, now merged, so I am closing this PR as superseded rather than landing the same change twice. Thanks for both patches.

PR #1385, opened 2026-08-06
fix(clean): keep cleaning when a cache removal times out (#1384)

### LINKED ISSUE #1384: [BUG] 一台老iMac在清理的时候总是cancel
先运行 --dry-run：
```
===========================
Dry run complete - no changes made
Potential space: At least 14.61GB | Items: 2524 | Categories: 8
Skipped while active: Chrome
Detailed file list: /Users/xxxx/.config/mole/clean-list.txt
Use mo clean --whitelist to add protection rules
========================================
```

再运行 mo clean：

```
Clean Your Mac

◎ Use --dry-run to preview, --whitelist to manage protected paths
✓ Admin access granted

⚙ Intel | Free space: 42.99GB
✓ Whitelist: 21 core patterns active

➤ System
  ✓ System diagnostic logs

➤ User essentials


===================================
Cleanup cancelled
System was already clean; no additional space freed.
Free space: 43.44GB (+448.9MB)
===================================
```
试过几次都这样，每次到这里都卡住了并取消了

系统是13.6.4 ，17还是18年的那个iMac 27.

Yuxin-Qiao:
## Summary

Fixes #1384: on slow machines (2017/2018 iMac, Fusion drive, macOS 13), `mo clean` consistently cancelled the whole run with `Cancelled: a scan or size check timed out (exit 124)` right after the `User essentials` section started, even though `--dry-run` completed fine. One slow `~/Library/Caches` item was enough to end every remaining section.

Two commits:

1. `fix(clean): keep cleaning when a cache size check times out (#1374)` - a sizing timeout (the per-item 30s `du` budget) now means "size unknown, keep cleaning" instead of cancelling. The affected item is still deleted, only the freed total is under-reported.
2. `fix(clean): keep cleaning when a cache removal times out (#1384)` - a removal timeout (the per-item 30s `rm -rf` budget) now means "this item was not removed, skip it and continue" instead of cancelling. `safe_remove`/`safe_sudo_remove` still return 124 to dedicated callers but no longer set the sticky `MOLE_CLEAN_CANCEL_STATUS`; only signals (`>=128`) remain fatal. Sizing probes inside the removal sinks (oplog size probe, sudo size probe, dry-run preview measurement) get the same non-fatal treatment as the batch sizing from #1374.

The summary now reports both cases instead of cancelling:

```text
Cleanup complete
Tracked cleanup: 25KB | Items cleaned: 6 | Categories: 1
◎ 1 item(s) exceeded the 30s removal budget and were left in place. Raise MOLE_TIMEOUT_DISK_VERIFY_SEC for slower disks.
```

Operators on very slow disks can raise `MOLE_TIMEOUT_DISK_VERIFY_SEC` to keep the per-item budgets while still getting a complete run either way.

## Reproduction

`repro-scripts/repro-1384-removal-timeout.sh` (throwaway HOME in /tmp, PATH stubs, no real files touched) seeds seven cache dirs and makes one of them slow via a 40s `du` or `rm` stub against the 30s budget:

```text
v1.49.1-slow-cache-du  exit 124  <- released version, exact reporter symptom
main-slow-cache-du     exit   0  <- sizing timeout is size-unknown (#1374)
main-slow-cache-rm     exit   0  <- removal timeout skips the item (#1384)
```

Before this PR, current main still exited 124 in the `rm` case with `Removal timed out` followed by `Cleanup cancelled`. Evidence logs: `evidence/clean-1384/` (before) and `evidence/clean-1384-fixed/` (after).

## Tests

- `tests/clean_core.bats`: removal timeout keeps cleaning (small batch and parallel-result loop), interrupted removal (`>=128`) still cancels.
- `tests/clean_summary_cancel.bats`: a run with removal timeouts completes and prints the under-report note.
- Existing sizing-timeout tests from #1374 continue to pass.
- `shellcheck` and `bash -n` pass for `bin/clean.sh` and `lib/core/file_ops.sh`.

### DIFF
--- a/bin/clean.sh
+++ b/bin/clean.sh
@@ -1071,8 +1071,12 @@ _safe_clean_impl() {
             run_with_timeout "$MOLE_TIMEOUT_DISK_VERIFY_SEC" \
                 stat -f%z "${existing_paths[@]}" < /dev/null \
                 > "$bulk_stat_file" 2> /dev/null || bulk_stat_rc=$?
-            if [[ $bulk_stat_rc -eq 124 || $bulk_stat_rc -ge 128 ]]; then
+            if [[ $bulk_stat_rc -ge 128 ]]; then
                 cleanup_interrupt_rc=$bulk_stat_rc
+            elif [[ $bulk_stat_rc -eq 124 ]]; then
+                # The size is only used for the freed total; a stalled stat
+                # must not cancel the delete set. Sizes are already 0 here.
+                MOLE_CLEAN_SIZING_TIMEOUTS=$((${MOLE_CLEAN_SIZING_TIMEOUTS:-0} + 1))
             fi
             while IFS= read -r _bytes; do
                 [[ "$_bytes" =~ ^[0-9]+$ ]] || _bytes=0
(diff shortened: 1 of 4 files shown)


tw93:
@Yuxin-Qiao Rebased onto main and merged. The split between sizing timeouts and removal timeouts is exactly what the slow-disk reports needed, thank you.

PR #1391, opened 2026-08-06
fix(clean): never unlink a live SQLite cache database (#1390)

### LINKED ISSUE #1390: [BUG] mo clean removes the live SQLite cache of a running app process, sending Fusion 360's AcCoreConsole into a disk-filling error loop
## Describe the bug

`mo clean` deleted `~/Library/Caches/com.autodesk.AcCoreConsole/Cache.db` together with its `-wal` and `-shm` companions while the owning process (Autodesk Fusion's `AcCoreConsole` helper) was running with that database open.

The helper entered an error loop: it began writing continuously to unlinked temp files (open file handles with no directory entry, under `$TMPDIR/ogs/<id><pid>/wip` and `/private/var/tmp/accc<pid>*`) plus Autodesk CER crash logging. It wrote ~20GB within minutes and drove the volume to 0 bytes free. Because the files were unlinked, `du`/`find` could not locate the space - only Activity Monitor's Bytes Written and `lsof -p <pid>` revealed the writer. The process ignored SIGTERM; SIGKILL released all held space immediately (free space went 0 -> 20GB within a minute).

The disk-full state was severe enough that new processes could not start and shell startup files failed to write.

Filed publicly rather than via SECURITY.md after some deliberation: there is no path-validation bypass, privilege issue, or attacker leverage here - the deleted path is a legitimate cleanup target and the gap is a missing process-state guard. Happy to move this to the private channel if you see it differently.

## Steps to reproduce

1. Have Autodesk Fusion (360) running - or its helpers still alive after the main app exits (`AcCoreConsole`, `ADPClientService`); the helpers can outlive the parent window
2. Run `mo clean` (the sweep reports `Autodesk cache · 9 items`, including `com.autodesk.AcCoreConsole/Cache.db`, `Cache.db-wal`, `Cache.db-shm`)
3. Watch free space fall continuously; in Activity Monitor (Disk tab, sort by Bytes Written), `AcCoreConsole` climbs by GBs

Recovery: `kill -9 <AcCoreConsole pid>` - space returns instantly as the unlinked files are released.

## Expected behavior

Per-app cache sweeps should skip caches whose owning process is currently running, the way the codebase already handles neighbouring cases:

- Simulator caches are skipped with `skipped (process state unknown)`
- Incomplete-download cleanup checks `lsof` before touching files
- `lib/core/file_ops.sh` refuses to delete Apple's active powerlog SQLite database, with the comment "Size and mtime are diagnostic signals, never deletion authority for a database that a KeepAlive system service can reopen" - the same reasoning applies to third-party helper processes
- AGENTS.md's own product boundary: "Do not remove ... active databases"

Possible fixes, cheapest first: (a) before deleting under `~/Library/Caches/<bundle-id>`, check for a running process matching the bundle id (`pgrep`/`lsappinfo`) and skip with a `skipped (app running)` outcome; (b) treat an SQLite `-shm` file (or `lsof` hit on the `.db`) as a liveness signal and skip the database family; (c) both, since helpers like AcCoreConsole don't always match the bundle name of the parent app.

While the reproduction above is Fusion 360, the fix is general: any app or helper daemon holding an open `Cache.db` could plausibly fail the same way.

## Debug logs

The incident run was a normal (non `--debug`) run, so no debug capture exists. Evidence from the incident:

<details>
<summary>Evidence</summary>

```text
# clean output (relevant line)
✓ Autodesk cache · 9 items, 10.9MB

# deleted paths (from the clean preview list)
~/Library/Caches/com.autodesk.AcCoreConsole/Cache.db      # 49KB
~/Library/Caches/com.autodesk.AcCoreConsole/Cache.db-shm  # 33KB
~/Library/Caches/com.autodesk.AcCoreConsole/Cache.db-wal  # 0B
~/Library/Caches/com.autodesk.AcCoreConsole/fsCachedData  # 0B

# where the looping process was writing (lsof -p <pid>, files open for write)
/private/var/folders/.../T/ogs/<id><pid>/wip
/private/var/tmp/accc<pid>YStlyd
~/Library/Application Support/Autodesk/CER/<hash>/cer.log

# free space: 8.1GB before clean -> 0B within ~10 minutes of clean finishing
# after kill -9: 12.3GB free -> 20.1GB free 30 seconds apart (df, 1M blocks)
/dev/disk3s5  233752 195366 12330 95% /System/Volumes/Data
/dev/disk3s5  233752 187547 20149 91% /System/Volumes/Data
```

</details>

## Environment

```text
Mole version 1.49.2
macOS: 26.5.2
Architecture: arm64
Kernel: 25.5.0
SIP: Enabled
Install: Homebrew
Shell: /bin/zsh
```

## Additional context

Workaround applied locally: `~/Library/Caches/com.autodesk.*` added to `~/.config/mole/whitelist`. Autodesk CER crash data from the incident is retained and available if useful.

Yuxin-Qiao:
## Summary

Fixes #1390: `mo clean` unlinked `~/Library/Caches/com.autodesk.AcCoreConsole/Cache.db` together with its `-wal` / `-shm` companions while the owning helper process had the database open. The owner then kept writing to the orphaned inodes, filling the volume with space `du`/`find` cannot see.

Two layers, so both the reported app and the general mechanism are covered:

1. **General SQLite liveness guard at the deletion boundary** (`lib/core/file_ops.sh`): `validate_path_for_deletion` now refuses any SQLite family member (`*.db` / `-wal` / `-shm` / `-journal`) whose `-shm` companion exists (WAL mode means at least one connection is open) or whose main file `lsof` reports open. A missing or unresponsive `lsof` denies rather than allows, matching the existing tri-state rule that "could not tell" is never permission to delete. The same rule is applied in `record_dry_run_cleanup_target` so the dry-run preview agrees with the real run. This protects every app and helper database a cache sweep can reach, including helpers whose process name does not match the parent app.
2. **Autodesk Fusion-family process guard** (`lib/clean/app_caches.sh`): the Autodesk cache sweep is now guarded like Xcode / Simulator / Final Cut Pro. `Autodesk Fusion`, `Fusion 360`, `AcCoreConsole`, and `ADPClientService` count as running and defer the family to the end-of-run "Skipped while active" list; an unknown process state stops the family with the existing `stopped (process state unknown)` outcome.

This follows the codebase's own boundary: the powerlog guard in `file_ops.sh` and AGENTS.md both say size/mtime are never deletion authority for a database a live process can keep using.

## Reproduction

`repro-scripts/repro-1390-live-sqlite.sh` builds a throwaway HOME (in `/tmp`, no real files touched) with `com.autodesk.AcCoreConsole/Cache.db` + `-wal` + `-shm` + `fsCachedData`, holds the database open with a real `sqlite3` WAL connection (the AcCoreConsole shape), and runs a real `mo clean` for each build:

```text
before (V1.49.2):  ✓ Autodesk cache · 4 items, 46KB
                   files_after:   <empty>
                   link count:    missing   <- unlinked while holder had it open

after (this PR):   ◎ Autodesk cache · stopped (process state unknown)
                   files_after:   Cache.db Cache.db-shm Cache.db-wal fsCachedData
                   link count:    1
```

Evidence logs: `evidence/clean-1390/` (before) and `evidence/clean-1390-fixed/` (after). With the process probe answering "not running", the same after-build keeps the family via the SQLite liveness guard (verified locally with a `pgrep` shim: `✓ Autodesk cache · 0B` and `Cache.db` family intact).

## Tests

- `tests/core_safe_functions.bats`: `validate_path_for_deletion` refuses a live `-shm` family, refuses a database `lsof` reports open, and allows an idle `.db`.
- `tests/clean_app_caches.bats`: `clean_3d_tools` defers the Autodesk family while a Fusion helper runs and cleans it when not running.
- `tests/clean_misc.bats`: `clean_3d_tools` still reaches Blender / Cinema 4D / Autodesk / SketchUp.
- `tests/clean_core.bats`: full `mo clean --dry-run` on a private HOME never previews a live SQLite family.
- `./scripts/check.sh`: go vet, ShellCheck, syntax, function-duplication audit all pass.

### DIFF
--- a/bin/clean.sh
+++ b/bin/clean.sh
@@ -338,6 +338,20 @@ record_dry_run_cleanup_target() {
     if declare -f holds_compiled_model_cache > /dev/null 2>&1 && holds_compiled_model_cache "$path" 2> /dev/null; then
         return 1
     fi
+    # Keep preview eligibility identical to real cleanup (#1390 / PR #1391).
+    if declare -f _mole_should_refuse_live_user_cache_path > /dev/null 2>&1 &&
+        _mole_should_refuse_live_user_cache_path "$path"; then
+        return 1
+    fi
+    if declare -f _mole_is_sqlite_database_path > /dev/null 2>&1 &&
+        _mole_is_sqlite_database_path "$path" &&
+        declare -f _mole_sqlite_database_in_use > /dev/null 2>&1; then
+        local sqlite_state=0
+        _mole_sqlite_database_in_use "$path" || sqlite_state=$?
+        if [[ $sqlite_state -eq 0 || $sqlite_state -eq 2 ]]; then
+            return 1
+        fi
+    fi
 
     if [[ -z "${CLEAN_PREVIEW_LEDGER_FILE:-}" || ! -f "$CLEAN_PREVIEW_LEDGER_FILE" ]]; then
         register_dry_run_cleanup_target "$path" || return 1
--- a/lib/clean/app_caches.sh
+++ b/lib/clean/app_caches.sh
@@ -62,6 +62,21 @@ _final_cut_pro_delete_guard_allows() {
     mole_clean_process_guard final_cut_pro_is_running "Final Cut Pro started"
 }
 
+_autodesk_cleanup_process_state() {
+    # Fusion 360 was renamed "Autodesk Fusion"; the app process and the helpers
+    # that keep the cleaned cache databases open (AcCoreConsole, ADPClientService),
+    # which can outlive the main window, all count as running (#1390).
+    mole_pgrep_any \
+        -x "Autodesk Fusion" \
+        -x "Fusion 360" \
+        -x "AcCoreConsole" \
+        -x "ADPClientService"
+}
+
+_autodesk_app_cache_delete_guard_allows() {
+    mole_clean_process_guard _autodesk_cleanup_process_state "Autodesk Fusion running"
+}
+
 _defer_app_cache_guard_family() {
     case "$1" in
         _simulator_app_cache_delete_guard_allows) mole_defer_cleanup_family "Simulator" ;;
--- a/lib/core/file_ops.sh
+++ b/lib/core/file_ops.sh
@@ -183,17 +183,34 @@ _mole_user_cache_owner_component() {
     return 0
 }
 
-_mole_is_user_cache_sqlite_family_path() {
-    local base
-    base=$(basename "$1")
+# SQLite main file or -wal / -shm / -journal companion. Case-insensitive so a
+# cache sweep cannot delete a database through a case variant of its name
+# (contributor PR #1391 + main reverse-DNS gate).
+_mole_is_sqlite_database_path() {
+    local restore_nocasematch=false
+    local result=1
+    local path="$1"
+    local base="${path%-wal}"
+    base="${base%-shm}"
+    base="${base%-journal}"
+
+    if ! shopt -q nocasematch; then
+        shopt -s nocasematch
+        restore_nocasematch=true
+    fi
+
     case "$base" in
-        *.db | *.db-wal | *.db-shm | \
-            *.sqlite | *.sqlite-wal | *.sqlite-shm | \
-            *.sqlite3 | *.sqlite3-wal | *.sqlite3-shm)
-            return 0
+        *.db | *.sqlite | *.sqlite3)
+            result=0
             ;;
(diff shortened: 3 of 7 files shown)


tw93:
Merged after rebasing onto current `main` and combining both approaches: your SQLite `-shm` / `validate_path` liveness gate plus the reverse-DNS process memoization and Autodesk deferral already on main. Thank you for the careful repro and the two-layer design for #1390.

PR #1399, opened 2026-08-07
fix: align battery Health percentage format with Level

hammorockstar:
## Summary

- `Health` used `%5d%%` (integer, no decimal place) while `Level` used `%5.1f%%` (one decimal place)
- This caused two extra leading spaces before the Health percentage value in the status TUI
- Fix: use `%5.1f%%` with `float64(b.Capacity)` for consistent formatting

**Before:**
```
Level  █████████████░░░   83.0%
Health █████████████░░░     86%
```

**After:**
```
Level  █████████████░░░   83.0%
Health █████████████░░░   86.0%
```

## Test plan

- [ ] Run `mo status` and verify Level and Health percentages are visually aligned

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/cmd/status/view.go
+++ b/cmd/status/view.go
@@ -752,7 +752,7 @@ func renderBatteryCard(batts []BatteryStatus, thermal ThermalStatus) cardData {
 
 		// Add capacity line if available.
 		if b.Capacity > 0 {
-			capacityText := fmt.Sprintf("%5d%%", b.Capacity)
+			capacityText := fmt.Sprintf("%5.1f%%", float64(b.Capacity))
 			if b.Capacity < 70 {
 				capacityText = dangerStyle.Render(capacityText)
 			} else if b.Capacity < 85 {


tw93:
@david-dacs Thanks for the patch, and sorry to turn this one down.

I rendered `renderBatteryCard` directly to check the claim, and the two rows already line up. `%5d%%` and `%5.1f%%` both produce exactly six columns for any capacity value, so the `%` signs land in the same terminal column:

```
Level  ███████░░░░░░░░░   45.0%
Health ████████████████    100%

Level  █████████████░░░   83.0%
Health █████████████░░░     86%

Level  █░░░░░░░░░░░░░░░    7.5%
Health █░░░░░░░░░░░░░░░      9%
```

The leading spaces you noticed are the right-alignment padding, which is what keeps those columns flush. What the patch actually moves is the decimal point, and `Capacity` is an `int` parsed from `sppower_battery_health_maximum_capacity`, so the `.0` it would print can never be anything but `.0`. I would rather not put a fixed decimal on a whole-number field.

Closing for that reason. If the card looks different on your Mac, a screenshot would be worth seeing and I will reopen.

tw93:
@david-dacs A fairer word than my earlier reply gave you: what you noticed is real. The decimal points in that card genuinely do not line up, and reading a column of numbers where one has a decimal and the other does not is the kind of thing most people scroll past. You went and found the two format strings, which is more than a report would have taken.

Where I landed differently is only on the remedy. `Capacity` comes off `sppower_battery_health_maximum_capacity` as a whole number, so printing a decimal there states a precision the value does not have, and I would rather live with the ragged decimal than with a digit that is always zero.

Please do send the next one. A patch that turns out to rest on a misread still cost you the work of reading the code, and I would rather have contributors who look.

PR #1407, opened 2026-08-09
fix(analyze): cancel live scans end to end

M-Hassan-Raza:
Live analyze scans now carry one cancellation context through recursive walks, semaphore waits, du, and Spotlight.

Cancellation also owns event and cache publication, so abandoned scans cannot emit stale completion events or commit cache state after navigation or refresh.

Tests cover blocked and nested du probes, worker shutdown, event capacity, and cache publication ordering. The full Go suite, race detector, repository checks, and local build pass.


tw93:
@M-Hassan-Raza Thanks for the thorough cancellation work. Live Analyze scans now carry cancellation through traversal, Spotlight, caching, and teardown. Merged.

PR #1409, opened 2026-08-10
[FEATURE] Allow opt-in optimization of SQLite databases larger than 100 MiB

r266-tech:
Adds a bounded, unit-bearing database size option, probe-before-policy classification, scaled timeout limits, temporary-space checks, factual skip reporting, and operation logging.

The patch was applied to the frozen base.

Focused optimize and SQLite regression tests completed with no failures beyond the frozen-base baseline.

`make check` passed.

### DIFF
--- a/bin/optimize.sh
+++ b/bin/optimize.sh
@@ -193,15 +193,25 @@ main() {
     # Set current command for operation logging
     export MOLE_CURRENT_COMMAND="optimize"
 
-    local health_json
-    for arg in "$@"; do
+    local health_json database_max_size_seen=0
+    while [[ $# -gt 0 ]]; do
+        local arg="$1"
         case "$arg" in
             "--help" | "-h")
                 show_optimize_help
+                echo "  --database-max-size SIZE  Allow curated SQLite databases up to SIZE (maximum 1GiB)"
                 exit 0
                 ;;
             "--debug")
                 export MO_DEBUG=1
+                ;;
+            "--database-max-size")
+                shift
+                if [[ "$database_max_size_seen" -eq 1 || $# -eq 0 ]] || ! optimize_set_database_max_size "$1"; then
+                    echo "Use 'mo optimize --help' for supported options." >&2
+                    exit 1
+                fi
+                database_max_size_seen=1
                 ;;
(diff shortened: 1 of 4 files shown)


r266-tech:
Updates the bounded, unit-bearing --database-max-size implementation by replacing the ShellCheck-incompatible suffix expansion with Bash-compatible literal suffix branches, and corrects the CLI propagation test mock to read the production run_with_timeout argument position.

Verified: ./scripts/check.sh --format, the four filtered mo optimize tests, all 18 optimize_db tests, and make check passed. The complete shell suite exceeded its runtime budget and was not completed.

r266-tech:
Adds a deterministic `sqlite3` fixture to the existing CLI propagation regression in `tests/optimize.bats`, so the valid bounded `--database-max-size` path reaches the curated dry-run optimization consistently.

Verified: `./scripts/check.sh --format`, the filtered CLI propagation test, `bats tests/optimize_db.bats`, and `make check` passed.

r266-tech:
SQLite databases above 100 MiB were skipped before freelist inspection, with no bounded opt-in for larger curated databases.

This change adds `--database-max-size` with explicit positive units and a 1 GiB upper bound, while retaining the 100 MiB default and limiting the override to the existing curated SQLite targets. It classifies compact databases before applying the size policy and preserves process, integrity, timeout, temporary-space, dry-run, logging, and factual outcome safeguards.

The CLI propagation regression now uses deterministic, side-effect-free SQLite responses and an explicit no-color presentation mode, so the dry-run summary assertion is independent of ambient terminal settings.

Verified:
- `bats --filter "mo optimize" tests/optimize.bats` (4 tests)
- `bats tests/optimize_db.bats` (18 tests)
- `make check`

All listed commands passed. The local `make check` run skipped ShellCheck because it was not installed and completed its remaining configured checks.

tw93:
@r266-tech Thanks for the thorough implementation and tests. I’m closing this because the reporting problem is already fixed on `main`, while the 100 MiB ceiling remains intentional; adding a size override broadens a live-database `VACUUM` path and introduces a safety setting Mole does not want to own.

PR #1413, opened 2026-08-10
fix(uninstall): guard empty array expansion in cask search for bash 3.2 (#1411)

### LINKED ISSUE #1411: [BUG] uninstall: `uniq[@]: unbound variable` in brew.sh breaks all Homebrew cask detection on macOS bash 3.2
## Describe the bug

`_detect_cask_via_caskroom_search()` in `lib/uninstall/brew.sh` iterates an array that is empty by construction:

```bash
# lib/uninstall/brew.sh:158-162
local -a uniq=()
local candidate existing seen
for candidate in "${tokens[@]}"; do
    seen=false
    for existing in "${uniq[@]}"; do   # <-- line 162, aborts on the first iteration
```

macOS ships Bash 3.2.57 as `/bin/bash`, and Bash before 4.4 treats an empty array expansion as an unset variable. Mole's shell libraries run under `set -euo pipefail`, so the first iteration dies with:

```
lib/uninstall/brew.sh: line 160: uniq[@]: unbound variable
```

(Bash 3.2 reports the enclosing `for candidate` line, 160; the offending expansion is on 162.)

`uniq` is always empty when the loop is entered, so this fires for every app that reaches stage 2 of `get_brew_cask_name()` — which is the normal layout for a cask-installed app: Homebrew moves the bundle to `/Applications/Foo.app` and leaves `Caskroom/<token>/<version>/Foo.app` as a symlink pointing back, so stage 1 (`_detect_cask_via_resolved_path`) fails and stage 2 finds the Caskroom match.

Two user-visible symptoms:

**1. `mo uninstall --list` aborts partway with exit 1 and truncated output.**

`bin/uninstall.sh:1437` and `:1481` call it as `cask=$(get_brew_cask_name "$app_path" 2> /dev/null || true)`. The `|| true` is *inside* the command substitution, so it cannot catch a `set -u` shell exit — the subshell is terminated before `|| true` runs, the assignment inherits status 1, and `set -e` in the parent kills the whole listing loop. On this machine the JSON array stops mid-stream without its closing `]`.

**2. Interactive `mo uninstall` silently stops treating cask apps as casks.**

`lib/uninstall/batch.sh:1487` captures the status (`|| cask_detect_rc=$?`), sees 1, and records `is_brew_cask=false`. The app is then removed by file deletion instead of `brew uninstall --cask`, so Homebrew's Caskroom metadata survives: `brew list --cask` still reports the app as installed, the cask's `zap` stanza never runs, and a later `brew upgrade` can reinstall it.

Introduced by d884a268 (`fix(uninstall): Harden brew uninstall`), first released in V1.22.1. Still present on `main` at e83f44f8 (VERSION 1.50.0).

## Steps to reproduce

Direct, on a clean checkout of `main` (substitute any cask-installed app):

```bash
git clone https://github.com/tw93/Mole /tmp/mole-repro && cd /tmp/mole-repro
/bin/bash --noprofile --norc -c '
set -euo pipefail
source lib/core/common.sh
source lib/uninstall/brew.sh
rc=0; out=$(get_brew_cask_name /Applications/Ghostty.app) || rc=$?
echo "rc=$rc out=[$out]"'
```

```
lib/uninstall/brew.sh: line 160: uniq[@]: unbound variable
rc=1 out=[]
```

Through the command surface:

```bash
MOLE_TEST_NO_AUTH=1 ./mole uninstall --list > /tmp/list.json; echo "rc=$?"
wc -l /tmp/list.json
grep -c '"source": "Homebrew"' /tmp/list.json
```

```
rc=1
       8          # 69 apps installed; output stops mid-array, no closing ]
       0
```

Note this only reproduces under macOS's `/bin/bash` 3.2. Under Homebrew's bash 5 the empty expansion is legal and nothing goes wrong, which is probably why it has gone unnoticed.

## Expected behavior

`get_brew_cask_name` returns the cask token for a cask-installed app; `mo uninstall --list` completes with exit 0 and labels those apps `Homebrew`; interactive uninstall routes them through `brew uninstall --cask`.

## Debug logs

Same machine, same commands, with only the one-line fix below applied:

```
rc=0 out=[ghostty]
```

```
rc=0
      69
      10
```

## Environment

```text
macOS 26 (Darwin 25.5.0), arm64
/bin/bash --version -> GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)
Mole: origin/main @ e83f44f8, VERSION="1.50.0"
```

## Additional context

One-line fix, using the guarded-expansion idiom already used elsewhere in the repo (e.g. `bin/uninstall.sh:1437` uses `"${apps_data[@]+"${apps_data[@]}"}"`):

```diff
-        for existing in "${uniq[@]}"; do
+        # Bash 3.2 treats an empty array as unset under `set -u`, and this
+        # loop starts empty by construction.
+        for existing in "${uniq[@]+"${uniq[@]}"}"; do
```

I checked the other arrays in `lib/uninstall/brew.sh` and `lib/uninstall/batch.sh` for the same shape: `tokens` (brew.sh:160) is guarded by the `((${#tokens[@]} > 0))` check on line 155, `uninstall_args` is never empty, `_mole_brew_probe`'s `"$@"` is fine (Bash special-cases empty `$@` even in 3.2), and every array loop in `batch.sh` is preceded by an explicit `-gt 0` / `-eq 0` guard. `uniq` is the only unguarded site that reproduces.

Happy to send a PR with the fix plus a regression test if that's useful — just say the word.

Yuxin-Qiao:
### Summary
Fixes `uniq[@]: unbound variable` failure in `_detect_cask_via_caskroom_search` when running under macOS system `/bin/bash` (version 3.2.57).

### Root Cause
In Bash < 4.4, expanding an empty array (`"${uniq[@]}"`) under `set -u` is treated as an unbound variable. Because `uniq` is initialized empty (`local -a uniq=()`), the deduplication loop in `_detect_cask_via_caskroom_search` throws `unbound variable` and aborts on its very first iteration, breaking Cask detection for installed applications.

### Solution
Guarded the empty array expansion using `"${uniq[@]+"${uniq[@]}"}"`, which is Bash 3.2 compatible and safe under `set -u`.

### Tests
- Added a Bats unit test in `tests/brew_uninstall.bats` verifying `_detect_cask_via_caskroom_search` executes cleanly under `/bin/bash` with `set -euo pipefail`.
- Verified all tests pass.

Closes #1411

### DIFF
--- a/lib/uninstall/brew.sh
+++ b/lib/uninstall/brew.sh
@@ -159,7 +159,7 @@ _detect_cask_via_caskroom_search() {
     local candidate existing seen
     for candidate in "${tokens[@]}"; do
         seen=false
-        for existing in "${uniq[@]}"; do
+        for existing in "${uniq[@]+"${uniq[@]}"}"; do
             if [[ "$candidate" == "$existing" ]]; then
                 seen=true
                 break
--- a/tests/brew_uninstall.bats
+++ b/tests/brew_uninstall.bats
@@ -717,4 +717,3 @@ EOF
     [ "$status" -eq 0 ]
     [[ "$output" == "test-cask-app" ]]
 }
-
--- a/tests/brew_uninstall.bats
+++ b/tests/brew_uninstall.bats
@@ -690,3 +690,31 @@ EOF
 
     [ "$status" -eq 0 ]
 }
+
+@test "_detect_cask_via_caskroom_search handles empty uniq array expansion under set -u" {
+    mkdir -p "$BATS_TEST_TMPDIR/TestCaskApp.app"
+
+    run env HOME="$HOME" PROJECT_ROOT="$PROJECT_ROOT" TEST_APP_PATH="$BATS_TEST_TMPDIR/TestCaskApp.app" /bin/bash --noprofile --norc << 'EOF'
+set -euo pipefail
+source "$PROJECT_ROOT/lib/core/common.sh"
+source "$PROJECT_ROOT/lib/uninstall/brew.sh"
+
+find() {
+    echo "/opt/homebrew/Caskroom/test-cask-app/1.0.0/TestCaskApp.app"
+}
+run_with_timeout() {
+    shift
+    "$@"
+}
+_mole_brew_probe() {
+    echo "test-cask-app"
+    return 0
+}
+
+_detect_cask_via_caskroom_search "$TEST_APP_PATH"
+EOF
+
+    [ "$status" -eq 0 ]
+    [[ "$output" == "test-cask-app" ]]
+}
+


tw93:
@Yuxin-Qiao Thanks for turning the Bash 3.2 report into a focused regression. I removed the stray EOF blank line, and the cask-detection fix is merged.

Issue #1370, opened 2026-08-04
[FEATURE] Prune stores for each locally installed pnpm major


TyceHerrman:
## Feature description

Extend Mole's pnpm cleanup to prune every distinct store generation that can be reached through an already-installed matching pnpm major version.

Current `main` resolves the store path from the active `pnpm` binary and runs `pnpm store prune` once. When pnpm 10 and pnpm 11 coexist, the active pnpm 11 binary selects `~/Library/pnpm/store/v11`; an unreferenced `v10` store remains untouched even when a usable pnpm 10 binary is already installed.

This is a narrow follow-up to #1295. The maintainer explicitly identified multi-version pnpm pruning as potentially useful after measuring actual reclaimed space:

https://github.com/tw93/Mole/issues/1295#issuecomment-5148544734

## Use case

Measured on macOS 27 arm64 with Mole 1.47.1. Current `main` was rechecked on 2026-08-04 and still uses the same one-active-binary behavior:

https://github.com/tw93/Mole/blob/main/lib/clean/dev.sh

At 2026-07-26 10:20:30 EDT, before pruning:

| Store | Matching installed binary | Allocated size |
| --- | --- | ---: |
| `~/Library/pnpm/store/v10` | pnpm 10.34.5 | 634 MiB |
| `~/Library/pnpm/store/v11` | pnpm 11.17.0 | 212 MiB |
| **Total** | | **846 MiB** |

The combined store contained 43,930 files, 787 directories, and 6 symlinks. The active `pnpm store path` resolved to `v11`. Direct checks showed:

```text
pnpm 10.34.5 -> ~/Library/pnpm/store/v10
pnpm 11.17.0 -> ~/Library/pnpm/store/v11
```

At 12:40:31 EDT, after confirming that no pnpm process was running, both already-installed binaries ran their native owner command:

```zsh
if pgrep -f '(^|/)(pnpm)( |$)' >/dev/null; then
  echo "Stopped: pnpm running."
else
  for pnpm_bin in \
    "$HOME/.local/share/mise/installs/pnpm/10.34.5/pnpm" \
    "$HOME/.local/share/mise/installs/pnpm/11.17.0/pnpm"
  do
    [[ -x "$pnpm_bin" ]] &&
      COREPACK_ENABLE_DOWNLOAD_PROMPT=0 "$pnpm_bin" store prune
  done
fi
```

[...]

PR #1419, opened 2026-08-12
feat(clean): report unmatched pnpm store generations

tenngoxars:
## Why

`pnpm store prune` only prunes the generation selected by the binary that runs it. #1370 covers every store reachable through the supported installed-binary inventory (PATH plus mise), but generation-shaped siblings can still remain under `~/Library/pnpm/store/`.

An unmatched directory is not enough evidence for automatic deletion: a usable pnpm binary may live outside the supported inventory, and a discovered binary may temporarily fail or time out while resolving its store. This PR makes those generations observable for manual review without treating an incomplete owner inventory as proof that they are orphaned.

## Change

- `collect_pnpm_store_snapshot` resolves the supported pnpm candidates once and stores deduplicated path/binary pairs in one snapshot.
- Resolved stores still run one plain owner-native `pnpm store prune`, with no `--force` and no downloads.
- A complete snapshot reports unmatched default-root generations through debug logging for manual review; it never sends them to a filesystem deletion sink.
- Generation candidates must be direct children whose basename matches `^v[0-9]+$`; names such as `v10-backup`, `v1.2`, and `vendor` are ignored.
- Any unusable, unresolved, timed-out, empty, relative, or unsafe candidate marks the snapshot incomplete, suppressing the unmatched-generation report.
- The no-pnpm path is quiet under macOS Bash 3.2 with `set -u`.

## Safety

- Unmatched generations are never deleted. PATH plus known version-manager roots cannot prove that no pnpm binary exists elsewhere.
- Native pruning remains bound to the same binary that resolved each store path.
- The whole flow skips when a pnpm process is running or process state is unknown.
- Corepack download prompts remain disabled; version and store-path probes remain timeout-bounded.
- Non-targets include every in-use generation, non-default store roots, non-generation sibling names, incomplete snapshots, and the entire default root when no supported pnpm resolves there.

## Validation

- `./scripts/check.sh` — passed (format, ShellCheck, Bash syntax, function duplication, diagnostic guidance).
- ShellCheck over every `tests/**/*.bats` and `tests/**/*.sh` file — passed.
- `MOLE_TEST_NO_AUTH=1 bats tests/clean_dev_caches.bats tests/dev_extended.bats` — 169/169 passed.
- New regressions cover strict generation names, incomplete snapshots, root containment, no raw deletion, and the no-pnpm Bash 3.2 nounset path.
- The full local suite was also attempted; all pnpm/developer-cleanup cases passed. The run was stopped after 1,384/1,581 tests because unrelated environment-sensitive suites had already failed locally.

## Scope

This revision deliberately does not automate whole-generation deletion. Doing that safely would require either owner-complete evidence or an explicit product decision to wipe unmatched stores as rebuildable caches rather than calling them proven orphans.

### DIFF
--- a/lib/clean/dev.sh
+++ b/lib/clean/dev.sh
@@ -194,12 +194,19 @@ list_installed_pnpm_binaries() {
     fi
 }
 
-# Resolve the store path of every usable installed pnpm binary, deduplicated,
-# one pair per line: "<store_path><TAB><pnpm_bin>". Never downloads; unusable
-# binaries are skipped. The binary is the first one that resolved the store,
-# which is the one that owns the prune call for that generation.
+# Resolve the store path of every pnpm candidate discovered by
+# list_installed_pnpm_binaries. Results stay in parallel global arrays so the
+# prune and read-only unmatched-generation report consume one snapshot. A
+# failed candidate marks the snapshot incomplete: successful stores may still
+# use their owner prune, but absence from that snapshot never authorizes a
+# filesystem delete or an unmatched-generation report.
 # shellcheck disable=SC2329
-list_in_use_pnpm_store_paths() {
+collect_pnpm_store_snapshot() {
+    _PNPM_STORE_SNAPSHOT_PATHS=()
+    _PNPM_STORE_SNAPSHOT_BINS=()
+    _PNPM_STORE_SNAPSHOT_CANDIDATES=0
+    _PNPM_STORE_SNAPSHOT_COMPLETE=true
+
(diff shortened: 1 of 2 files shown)


tw93:
@tenngoxars Thanks for keeping this report-only and fail-closed. I’m closing it as not planned: unmatched pnpm generations cannot be proven orphaned, and a debug-only inventory adds substantial cleanup complexity without a normal user outcome; Mole will keep owner-native `pnpm store prune` as the supported path.

PR #1420, opened 2026-08-12
Group purge artifacts by project

M-Hassan-Raza:
Purge currently mixes build artifacts from unrelated projects in one size-sorted list, making it hard to keep an active project without inspecting every row.

This resolves each artifact to one exact physical project root, keeps projects together, and orders groups by aggregate reclaimable size. The selector adds visible group connectors, current-project size and selection context, and X Skip Project. Same-named projects and separate worktrees remain independent.

The supporting reliability changes discard incomplete scan output, report partial root failures, bound concurrent root scans, and preserve non-fatal purge outcomes. No deletion targets, matching rules, flags, settings, caches, or name-based selectors were added. Dry-run, artifact-level selection, full-path confirmation, activity rechecks, and the deletion funnel remain intact.

Project groups in the purge selector
Skipping one exact project

Verified with 110 focused purge/config tests, the exact 195-test high-risk CI command, ShellCheck, syntax checks, Go lint, formatting, and the duplication audit.

### DIFF
--- a/bin/purge.sh
+++ b/bin/purge.sh
@@ -224,17 +224,13 @@ perform_purge() {
     fi
 
     clean_project_artifacts
-    local exit_code=$?
+    local purge_outcome="${PURGE_RUN_OUTCOME:-scan_failed}"
 
     # Clean up
     trap - INT TERM
     cleanup_monitor
 
-    # Exit codes:
-    # 0 = success, show summary
-    # 1 = user cancelled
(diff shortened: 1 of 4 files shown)


tw93:
@M-Hassan-Raza Thanks for reworking purge around project groups. I replaced the quadratic grouping pass with one sort; the grouped selector, skip-project action, and fail-closed scan handling are merged.

PR #1422, opened 2026-08-13
feat(clean): remove leftover cask staging dirs under var/homebrew/tmp

ggdayup:
## Summary

Homebrew stages cask installs under `$(brew --prefix)/var/homebrew/tmp/.caskroom`. When a cask upgrade is interrupted or fails — e.g. the app is currently running, or the installer is a `.pkg` requiring sudo — Homebrew leaves both the `<version>` directory and its `<version>.staged` symlink behind. These leftovers are **not** touched by `brew cleanup` (which only handles `~/Library/Caches/Homebrew`), so they can accumulate to several GB over time. On a real machine this was observed at **12 GB across 13 casks** (Docker Desktop alone: 3 versions, 6.3 GB).

## Changes

- `lib/clean/brew.sh`: new `clean_homebrew_tmp_staged()`:
  - Resolves `$(brew --prefix)/var/homebrew/tmp/.caskroom`, iterates each cask dir, removes every `<version>.staged` symlink **and** the directory it points to
  - Safety: only ever deletes paths inside the staging root (symlink target is re-validated against it), skips entirely when `brew` is running (never races an in-progress install), and respects the user whitelist and `--dry-run`
  - Uses the project's `safe_remove` / `safe_remove_symlink` wrappers (never raw `rm`)
- `lib/clean/dev.sh`: wired the step into `clean_developer_tools`, **before** `clean_homebrew` so it runs on every clean, independent of the 7-day throttle
- Tests: 2 bats tests in `clean_system_maintenance.bats` (removal + no-op), plus stub updates in `clean_dev_caches.bats` / `dev_extended.bats`

## Why not just brew cleanup?

`brew cleanup` only prunes the download cache and old formula versions; `var/homebrew/tmp/.caskroom` staged leftovers are never visited. The existing `mo installer` also only scans `~/Library/Caches/Homebrew` for `.dmg`/`.pkg` files.

## Verification

- `scripts/check.sh --no-format` passes (go vet, ShellCheck, syntax, function duplication)
- bats tests for the new function pass (locally; the full-file suite hits a local sandbox bulk-delete restriction unrelated to these changes)

### DIFF
--- a/lib/clean/brew.sh
+++ b/lib/clean/brew.sh
@@ -198,6 +198,71 @@ restore_homebrew_active_links() {
     fi
 }
 
+# Clean leftover cask staging files under var/homebrew/tmp/.caskroom.
+# Homebrew stages cask installs here; an interrupted or failed upgrade (e.g.
+# the app is running, or a .pkg installer needs sudo) leaves the <version>
+# dir and its <version>.staged symlink behind, which can grow to several GB.
+# Runs independently of the 7-day clean_homebrew throttle so leftovers are
+# removed on every clean.
+clean_homebrew_tmp_staged() {
+    command -v brew > /dev/null 2>&1 || return 0
+    # Never race an in-progress cask install.
+    if pgrep -x brew > /dev/null 2>&1; then
+        debug_log "Homebrew cask staging cleanup skipped: brew is running"
+        return 0
+    fi
+
+    local prefix=""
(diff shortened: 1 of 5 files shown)


tw93:
@ggdayup Thanks for measuring the real 12 GB impact. I’m closing this as not planned: `.staged` is Homebrew transaction state, and neither the marker nor a one-time process check proves the target is abandoned at deletion time; the current containment check can also be escaped through a symlinked ancestor. Mole should leave this to Homebrew or explicit manual review instead of deleting it automatically.

PR #1428, opened 2026-08-14
fix(purge): honor user whitelist entries

### LINKED ISSUE #1427: [BUG] mo purge deletes whitelisted paths — bin/purge.sh never loads ~/.config/mole/whitelist
### Summary
`safe_remove` honors `is_path_whitelisted` (`lib/core/file_ops.sh:871-878`), but only `bin/clean.sh` (lines 70-126) parses `~/.config/mole/whitelist` into `WHITELIST_PATTERNS`. `bin/purge.sh` sources `common.sh` / `log.sh` / `clean/project.sh` — none of which load the whitelist — so during `mo purge` the array is empty and `is_path_whitelisted` returns false for every path. Whitelisted paths are deleted.

### Repro (Mole 1.50.0, macOS 26 / Apple Silicon)
1. Add a project's `node_modules` to `~/.config/mole/whitelist` (e.g. `$HOME/myproject/node_modules`)
2. Run `mo purge`, select that entry, confirm
3. It is deleted. Expected: skipped, the way `mo clean` skips whitelisted paths.

### Real impact
`mo purge` deleted a whitelisted service's `node_modules`, silently breaking a production send path until deps were reinstalled.

### Suggested fix
Hoist the whitelist-loading block from `bin/clean.sh:70-126` into a shared lib function (e.g. `lib/core/base.sh`) and call it from `bin/purge.sh` before `clean_project_artifacts`.

Verified locally: with the same loading block added to `purge.sh`, `safe_remove` correctly skips — a sandboxed purge over two canary projects deleted the unprotected one and skipped the whitelisted one (`Items: 1`).

Yuxin-Qiao:
## Summary

- centralize validated whitelist loading in the shared core policy
- initialize the invoking user's whitelist before `mo purge` scans and reuse the same loader from `mo clean`
- add a regression test proving a whitelisted project artifact survives purge while an unwhitelisted artifact is removed

## Root cause

`bin/purge.sh` sourced the project cleanup helpers but never populated `WHITELIST_PATTERNS`. The downstream `safe_remove` guard was correct, but it received an empty whitelist.

Fixes #1427

## Validation

- `bash -n lib/core/base.sh bin/clean.sh bin/purge.sh lib/clean/project.sh tests/purge.bats`
- `shellcheck lib/core/base.sh bin/clean.sh bin/purge.sh`
- `MOLE_TEST_NO_AUTH=1 bats tests/purge.bats tests/manage_whitelist.bats` (108/108)
- full suite attempted; remaining failures are environment-bound system probes, permissions, TTY, and network/cache setup rather than this change

### DIFF
--- a/bin/clean.sh
+++ b/bin/clean.sh
@@ -37,6 +37,8 @@ IS_M_SERIES=$([[ "$(uname -m)" == "arm64" ]] && echo "true" || echo "false")
 MOLE_USER_HOME="$(get_invoking_home)"
 [[ -n "$MOLE_USER_HOME" ]] || MOLE_USER_HOME="$HOME"
 
+load_mole_whitelist "$MOLE_USER_HOME"
+
 CLEAN_PREVIEW_FINAL_FILE="$MOLE_USER_HOME/.config/mole/clean-list.txt"
 CLEAN_PREVIEW_STAGING_FILE=""
 CLEAN_PREVIEW_LEDGER_FILE=""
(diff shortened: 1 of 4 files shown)


tw93:
@Yuxin-Qiao Thanks for catching the policy gap. Purge now loads the same user whitelist before evaluating artifacts, with a regression for the command surface. Merged.

Issue #910, opened 2026-05-16
[FEATURE] Clean obsolete VS Code-compatible extension directories


TyceHerrman:
## Summary

clean installed VS Code-compatible extension directories marked obsolete

These editors write `.obsolete` marker files under their extension roots:

```text
~/.cursor/extensions/.obsolete
~/.vscode/extensions/.obsolete
~/.vscode-insiders/extensions/.obsolete
```

The file is JSON, and its keys are obsolete extension directory names.

Example:

```text
obsolete entry: anthropic.claude-code-2.1.142-darwin-arm64
active manifest: ~/.cursor/extensions/anthropic.claude-code-2.1.143-darwin-arm64
```

## Requested behavior

During `mo clean`, for each supported extension root:

1. Read `.obsolete`, if present.
2. Treat each key as a direct child directory name under that extension root.
3. Delete only existing direct child directories named by `.obsolete`.
4. Leave `.obsolete` itself intact.
5. Skip or warn if the matching editor appears to be running.

Suggested roots:

```text
~/.cursor/extensions
~/.vscode/extensions
~/.vscode-insiders/extensions
~/.vscode-oss/extensions
~/.vscodium/extensions
~/.vscodium-insiders/extensions
```

tw93:
@TyceHerrman thanks for the detailed pattern. The Mole Mac app delegates `clean` rules to the CLI side, so this is a CLI feature request. I will track it for the next CLI clean rules update; the .obsolete JSON parse + per-root iteration list above is a clean spec, no code changes needed on the Mac side.

tw93:
@TyceHerrman Thanks for the suggestion. `mo clean` now reads each editor's `extensions/.obsolete` marker (Cursor, VS Code, VS Code Insiders), and offers the stale extension directories it lists for removal. Keys are validated as plain direct-child names so a crafted `.obsolete` cannot escape the extensions root. Fixed on `main`.

This will be in the next nightly. Upgrade with `mo update --nightly`.

PR #1429, opened 2026-08-15
feat: clean guarded PyInstaller and Clang caches

jiji262:
## Summary

Adds four cleanup targets, all rebuildable caches that accumulate quietly and are not covered today. Sizes below were measured on a daily-driver Apple Silicon Mac (macOS 15.6); several of them regrew within hours of being cleaned manually, so they keep paying off:

| Target | Location | Measured |
|---|---|---|
| pytest temp dirs | `$TMPDIR/pytest-of-*` | 371 MB (regrew to 306 MB the same day) |
| Node compile cache / Jest temp | `$TMPDIR/node-compile-cache`, `$TMPDIR/jest_*` | 23 MB + 23 MB (regrew to 79 MB) |
| Clang module cache (CLI builds) | `$(getconf DARWIN_USER_CACHE_DIR)clang` | 219 MB |
| PyInstaller bincache | `~/Library/Application Support/pyinstaller` (`bincache00py311…`, `bincache00py312…`) | 277 MB |
| electron-updater staged downloads | `~/Library/Application Support/Caches/*-updater/pending` | 138 MB (Tabby) |

`clean_dev_python`, `clean_dev_frontend`, and `clean_dev_other_langs` in `lib/clean/dev.sh` pick up the first four; a new `clean_electron_updater_caches` in `lib/clean/app_caches.sh` (registered in `clean_user_gui_applications`) handles the last one.

Explicit non-targets:

- Omaha-style updaters (e.g. Comet's `CometUpdater`) are untouched: safe removal there needs an installed-version comparison like `clean_edge_updater_old_versions`, not a blanket delete.
- Inside `*-updater` only `pending/` payloads are removed; updater metadata stays. An interrupted update is simply re-downloaded on demand.
- In `$TMPDIR` only the three named tool patterns are matched; sibling entries can be live app state.
- In `DARWIN_USER_CACHE_DIR` only `clang/` is touched; Xcode's module cache lives in DerivedData and stays with the Xcode cleanup.

## Safety Review

- Affects cleanup behavior: yes, five new `safe_clean` globs. All user-scope, no new sudo, no new deletion primitives; whitelist filtering, protected-path checks, and dry-run preview apply as usual via `safe_clean`.
- No changes to path validation, symlink handling, or install/update paths.

## Tests

- `./scripts/check.sh` passes.
- `MOLE_TEST_NO_AUTH=1 ./scripts/test.sh` (full Bats suite): 1616 tests, 8 failures, 15 skipped — no new failures. All 8 reproduce identically on a clean `main` checkout of this machine (`clean_editor_obsolete_extensions` ×2 (#910), project-root discovery ×4, plus two environment-flaky cases that pass on re-run), unrelated to this change.
- Manual: created fixtures for every new target, ran `MOLE_TEST_NO_AUTH=1 MOLE_DRY_RUN=1 ./mole clean`, and confirmed all six labels appear with correct sizes and no unintended matches (verified each glob's expansion with `du` before/after).

## Safety-related changes

- None beyond the new cleanup globs described above.

### DIFF
--- a/lib/clean/app_caches.sh
+++ b/lib/clean/app_caches.sh
@@ -1029,13 +1029,6 @@ clean_remote_desktop() {
     safe_clean ~/Library/Caches/com.sunlogin.*/* "Sunlogin cache"
 }
 # Main entry for GUI app cleanup.
-# Shared electron-updater download staging (~/Library/Application
-# Support/Caches/<app>-updater). Only the pending payloads are removed:
-# an interrupted update is re-downloaded by the app's updater on demand.
-clean_electron_updater_caches() {
-    safe_clean ~/Library/Application\ Support/Caches/*-updater/pending/* "Electron updater downloads"
-}
-
 clean_user_gui_applications() {
     stop_section_spinner
     clean_communication_apps
@@ -1058,5 +1051,4 @@ clean_user_gui_applications() {
     clean_note_apps
     clean_launcher_apps
     clean_remote_desktop
-    clean_electron_updater_caches
 }
--- a/lib/clean/app_caches.sh
+++ b/lib/clean/app_caches.sh
@@ -1029,6 +1029,13 @@ clean_remote_desktop() {
     safe_clean ~/Library/Caches/com.sunlogin.*/* "Sunlogin cache"
 }
 # Main entry for GUI app cleanup.
+# Shared electron-updater download staging (~/Library/Application
+# Support/Caches/<app>-updater). Only the pending payloads are removed:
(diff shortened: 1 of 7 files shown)


tw93:
@jiji262, thank you for measuring these caches and documenting the non-targets. I kept the measured PyInstaller bincache and Clang module cache, added active-process, path-identity, symlink, and whitelist guards, and merged the PR. I left out the shared temporary and electron-updater targets because they can still be active owner state.

Issue #1084, opened 2026-06-08
[BUG] mo clean prompts for sudo password multiple times mid-run, impossible to complete


kazani-351:
## Describe the bug

`mo clean` prompts for sudo password repeatedly mid-run (5–7+ times) while the spinner is animating. The password prompt appears behind the spinner output, making it impossible to see or respond to cleanly. The command never completes — it hangs indefinitely waiting for input that gets swallowed by the spinner.

## Steps to reproduce

1. Run `sudo -v && mo clean --debug`
2. Enter password when prompted initially
3. Command reaches "Scanning accessible rebuildable GPU caches" and begins prompting for sudo repeatedly mid-spinner
4. Command hangs — never completes

## Expected behavior

Sudo credentials entered once at the start are reused for the full run. All privileged operations should either be batched into a single sudo call, or sudo should be prompted explicitly before the spinner starts so input isn't swallowed by the animation.

## Debug logs

```bash
mo clean --debug
```

<details>
<summary>Debug output</summary>

```text
Cleaning rebuildable system service caches...Path validation: protected path skipped: /Library/Caches/com.apple.iconservices.store
  - Cleaning rebuildable system service caches...[DEBUG] Skipped sudo remove for protected path: /Library/Caches/com.apple.iconservices.store
  \ Scanning accessible rebuildable GPU caches...Password:
  / Scanning accessible rebuildable GPU caches...Password:
  \ Scanning accessible rebuildable GPU caches...Password:
  - Scanning accessible rebuildable GPU caches...Password:
  | Scanning accessible rebuildable GPU caches...Password:
  \ Scanning accessible rebuildable GPU caches...Password:
  / Scanning accessible rebuildable GPU caches...Password:
```

</details>

## Environment

```text
Mole version 1.42.0
macOS: 12.7.6 (Monterey)
Architecture: x86_64
Kernel: 21.6.0
SIP: Disabled
Disk Free: 84.71GB
Install: Manual
Shell: /bin/zsh

<img width="584" height="338" alt="Image" src="https://github.com/user-attachments/assets/87b021f2-f0d9-4b9b-841c-c6e163f0e745" />

```

## Additional context

Workarounds attempted — all failed:
- `sudo -v && mo clean` — credential cache not inherited by subshells
- Background sudo refresh: `sudo -v && (while true; do sudo -v; sleep 50; done &) && mo clean` — same issue
- Adding `/Library/Caches` to whitelist — prompts still appear

[...]

Issue #723, opened 2026-04-13
Safety: route destructive deletes through macOS Trash by default (with --permanent opt-out)


oysi2025:
## Summary

Mole is a powerful destructive tool — `mo clean`, `mo uninstall`, `mo purge` all call `rm -rf` on user data. A single bug in a path-building expression, a typo in a whitelist entry, or a mis-parsed glob can silently nuke irreplaceable files. Even with `--dry-run` and whitelists, there is no recovery path once a delete happens.

macOS has a built-in safety net: the Trash. This proposal is to route all user-facing deletes through Trash by default, and keep true `rm` behind an explicit `--permanent` flag (or reserve it for caches that are genuinely disposable).

## Motivation

- Mole's own `SECURITY_AUDIT.md` acknowledges the blast radius of destructive commands.
- Users running `mo clean` on an unfamiliar machine have no undo.
- Competing tools (CleanMyMac, AppCleaner) default to Trash for app removal.
- Cost: ~zero — macOS empties Trash on schedule or on user demand.

## Proposed design

1. Add a single helper `mole_delete <path>` in `lib/core/file_ops.sh` that:
   - Default: moves to Trash via `osascript -e 'tell app "Finder" to delete POSIX file ...'` or the lighter `trash` CLI (fall back to AppleScript if missing).
   - `--permanent` flag or `MOLE_DELETE_MODE=permanent` env: uses `rm -rf` as today.
   - Logs every deletion (source path, mode, timestamp, byte size) to `~/Library/Logs/mole/deletions.log` for forensic review.

2. Replace direct `rm -rf` calls across `lib/clean/*`, `lib/uninstall/*`, `lib/manage/purge_paths.sh` with `mole_delete`.

3. Exceptions that stay on `rm`:
   - Caches under `~/Library/Caches/*` that are re-created automatically (configurable list).
   - Files already inside `.Trash`.

4. New top-level flag `--permanent` for `clean`, `uninstall`, `purge`; documented in `--help`.

## Migration plan

- **Phase 1**: Add helper + deletion log, keep `rm` everywhere, but log through the helper. Ships observability with zero behavior change.
- **Phase 2**: Flip default to Trash for `uninstall` only (highest user value, lowest regression risk).
- **Phase 3**: Flip default for `clean` and `purge`, with a release-note heads-up.

## Test plan

- `bats` suite that deletes fixtures and asserts files land in `~/.Trash`, not gone.
- Dry-run assertions unchanged.
[...]

Issue #1373, opened 2026-08-04
[FEATURE] Add opt-in --trash mode to clean, purge, and installer


Nebow:
## Feature request

Following the discussion in #723, add an explicit `--trash` option to the following commands:

```bash
mo clean --trash
mo purge --trash
mo installer --trash
```

When provided, eligible deletion targets should be moved to the macOS Trash instead of being permanently deleted. The user would then need to empty the Trash manually to reclaim the disk space.

Without `--trash`, existing behavior should remain unchanged.

### Motivation

This would provide a recovery path for accidental cleanup while avoiding any change to Mole’s current defaults. Since the user must explicitly choose `--trash`, they understand that:

* disk space will not be reclaimed immediately;
* files remain recoverable until the Trash is emptied;
* large cleanup operations may temporarily increase Trash usage.

This would also make these commands more consistent with:

* `mo uninstall`, which already uses Trash by default;
* `mo analyze`, which already moves user-selected items to Trash.

### Expected behavior

* Preserve the existing confirmation and dry-run flows.
* Clearly report when files are being moved to Trash rather than permanently deleted.
* Record each successful deletion using the existing operation log.
* Reuse the existing `mole_delete` Trash-routing logic.
* Fail safely when a target cannot be moved to Trash because of permissions or an unsupported location.

### Implementation notes

* `mo installer` already uses `mole_delete`, so this should require relatively little wiring.
* `mo purge` should route its selected artifact removals through the shared deletion helper.
* `mo clean` has several deletion paths and safe-removal wrappers that would need to honor the selected deletion mode.
* Non-file operations and unrelated commands such as `optimize`, `update`, and `remove` would remain outside this change.

This keeps the current defaults intact while giving users an explicit, reversible option for all three user-facing cleanup commands.

tw93:
@Nebow Thank you for the proposal. I am not adding a `--trash` mode for clean, purge, or installer: those stay permanent-delete by design, and recoverable Trash stays on uninstall and analyze. Closing as not planned.

PR #1442, opened 2026-08-20
feat(clean): add --trash and confirm system caches on interactive runs

eryueniao7350:
## Why

`mo clean` is the only Mole command that deletes permanently — `mo uninstall`
has routed through the Trash since #723 — and on an interactive run it enabled
the privileged sweep from a cached sudo timestamp without asking.

This PR closes both gaps. They are separable; happy to split if you prefer.

## `mo clean --trash`

Opt-in recoverable cleanup. The switch sits at the shared sink (`safe_remove`
and `safe_sudo_remove`), after path validation, the final identity re-check and
the sink guard have all run, so only the last action differs — no call site in
`lib/clean` changes and no future one can forget to honour it.

- **Never downgrades.** A failed Trash move leaves the item in place rather
  than falling back to `rm -rf`; the run summary reports how many were left
  behind, so silence cannot read as "cleaned".
- **Covers the sudo sections** via the same privileged mover `mole_delete`
  already uses for root-owned app bundles.
- `safe_sudo_find_delete` drops back to its single-file path in trash mode —
  the privileged `xargs rm` batch has no Trash equivalent.
- Removals log as `TRASHED`, which `mo history` already counts and renders.

## System caches confirmed on interactive runs

`start_cleanup` called `adopt_sudo_session` and, on success, set
`SYSTEM_CLEAN=true` with no prompt. A timestamp the user established for
something else authenticates them; it does not say this run was meant to delete
outside `$HOME`.

The session is still adopted first, so the confirmation costs a keypress rather
than a password, and every later privileged call still stays out of the spinner.
The prompt says "admin access already active" when it is.

**#1084 is unaffected.** That issue asked for credentials entered once to be
reused for the whole run instead of prompting mid-spinner, and that contract is
untouched — including the non-interactive path, which has no terminal to ask.
`tests/clean_core.bats` "mo clean adopts cached sudo before system cleanup
(#1084)" passes unmodified.

## Testing

- `tests/clean_trash_mode.bats` (7) — routing, the no-fallback guarantee, the
  `TRASHED` log line, dry-run, and the same two through `safe_clean`.
- `tests/clean_system_confirm.bats` (5) — interactive with and without a cached
  session, the prompt wording, and both non-interactive paths.
- `tests/completion.bats` snapshot updated for the new flag.
- `shellcheck` and `shfmt -i 4 -ci -sr` clean; full `./scripts/test.sh`
  compared against a pristine `main` checkout.

No Go code touched.

### DIFF
--- a/bin/clean.sh
+++ b/bin/clean.sh
@@ -442,10 +442,24 @@ read_clean_sudo_password_remainder() {
     printf -v "$__remainder_var" '%s' "$remainder"
 }
 
+# Seam for the interactive check so the authorization policy can be covered
+# without allocating a pty.
+clean_stdin_is_tty() {
+    [[ -t 0 ]]
+}
+
 prompt_for_system_clean() {
+    local sudo_cached="${1:-false}"
     local prompt_attempt=0
+    local prompt_line="System caches need sudo."
+    if [[ "$sudo_cached" == "true" ]]; then
+        # A cached timestamp settles who the user is, not whether this run was
+        # meant to sweep system caches. Keep asking; only the password step is
+        # already done.
+        prompt_line="System caches, admin access already active."
+    fi
     while true; do
-        echo -ne "${PURPLE}${ICON_ARROW}${NC} System caches need sudo. ${GREEN}Enter${NC} continue, ${GRAY}Space${NC} skip: "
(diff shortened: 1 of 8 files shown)


tw93:
@eryueniao7350 thank you for the thoughtful work. I’m keeping clean, purge, and installer as permanent-delete commands by design, with recoverable Trash behavior limited to uninstall and analyze; this is the same product decision recorded in #1373, so I’m closing this as not planned.

Issue #545, opened 2026-03-06
[FEATURE] Selective clean (run only specific sections)


spider-yamet:
## Feature description

Allow users to run **only selected sections** of `mo clean` (e.g. only caches, only browsers, only developer tools) via a flag like `--only <category>` or `--sections`. No new cleanup logic—only filtering which of the **existing** sections run.

## Use case

Sometimes users want to clean one area (e.g. browser caches) without running the full clean. Today they must run the whole flow or skip interactively. A focused run (e.g. `mo clean --only browsers` or `mo clean --only caches`) is faster, easier to script, and keeps Mole’s behavior predictable and transparent.

## Proposed solution

- Add a flag, e.g. `mo clean --only <section>` or `mo clean --sections "caches,browsers"`.
- Sections could map to current groups, e.g.:
  - `caches` – user app cache, sandboxed app caches, macOS system caches
  - `browsers` – browser caches
  - `dev` – developer tools
  - `system` – deep system (if sudo available)
  - `trash` – trash only
  - etc.
- `--dry-run` and `--whitelist` behave as today; no new deletion rules.
- Help text lists available sections (e.g. `mo clean --only help` or in `mo clean --help`).

## Alternatives considered

- **Separate subcommands per category:** Would duplicate logic and bloat the CLI; a single `clean` with `--only` keeps one code path.
- **Config file to disable sections:** Possible future extension; `--only` is simpler and doesn’t require config file changes.

## Additional context

- Stays “lightweight and focused”: same cleanup rules, only a filter on which sections run.
- No time-based or new risky logic; all existing safeguards apply.

tw93:
@spider-yamet Thanks for the suggestion. I understand the use case, but for Mole I want to keep mo clean as simple and direct as possible, and avoid adding more grouping options or extra configuration. If there are items users do not want cleaned, the current whitelist support is the better fit for now.

PR #1443, opened 2026-08-20
feat(clean): add --only and --skip section selection

eryueniao7350:
## Why

`--select` / `--categories` / `--exclude` were removed in v1.51.0, which leaves
`--whitelist` — one path at a time — as the only way to hold something back. A
developer who wants everything except the toolchain caches, or who wants to
re-run just the browser sweep, has no way to say so.

If removing category selection was a deliberate simplification rather than a
cleanup of a confusing flag set, say so and I will close this — the safety
changes in the companion PR stand on their own.

## What

Section-level selection with stable keys:

```
mo clean --skip dev,projects
mo clean --only browsers,app-caches
mo clean --list-sections
```

Fifteen keys, printed by `--list-sections`. `--only` and `--skip` are mutually
exclusive, and both are rejected alongside `--external`, which is a
single-purpose sweep with its own section.

An unknown key is an error, not a no-op: silently ignoring a typo would run a
different cleanup than the one that was asked for.

## Implementation

Selection is resolved in `start_section`, the one point every section name
passes through, so **the fifteen section blocks in `perform_cleanup` are
unchanged** — no re-indentation, no per-block conditionals.

- `_run_cleanup_step` skips the steps of a deselected section.
- `_safe_clean_impl` carries a backstop for a step that opens its own section
  after the selection was resolved (`clean_apple_silicon_caches` today).
- A deselected section prints no header at all.
- Unknown labels fail open: a nested section is only reachable through a step
  whose own section was already selected.
- A run that deselects `system` never asks for admin access.

## Testing

`tests/clean_sections.bats` (11) covers the parser, the conflict rules, both
selection directions, fail-open on an unknown label, that a deselected section
deletes nothing and prints nothing, that a selected one still deletes, and that
the flag resets between sections.

`shellcheck` and `shfmt -i 4 -ci -sr` clean; full `./scripts/test.sh` compared
against a pristine `main` checkout. Manual dry runs: default lists 14 sections,
`--skip dev,browsers` drops exactly those two, `--only app-caches` leaves one.

No Go code touched.


tw93:
@eryueniao7350 thank you for checking first. Section selection was removed deliberately to keep mo clean as one simple flow; whitelist remains the supported way to protect paths, matching the decision in #545, so I’m closing this as not planned.

PR #1445, opened 2026-08-20
feat(clean): remove old Autodesk Fusion app bundles

### LINKED ISSUE #1438: [FEATURE] Remove unused Autodesk Fusion app bundles
## Feature description

Autodesk Fusion has an in app auto update mechanism. It ends up keeping around months worth of old versions of the app.

It keeps the app bundles in `~/Library/Application Support/Autodesk/webdeploy/production`

In there it has an alias named `Autodesk Fusion.app` pointing at the current version.

Each version of the app bundle is in its own folder, an example name is `5b508d94493ed3344e38945ffca1b03b713ee401`

The actual app is launched from a stub app bundle in `~/Applications/Autodesk Fusion.app`

## Use case

This would free up a fair bit of wasted space, I had 60GB of unused app bundles after using Fusion for a couple of months.

## Proposed solution

Delete the folders containing the old app bundles.

Identify which is the current version using the alias, don't delete that.

The `~/Library/Application Support/Autodesk/webdeploy/production` also includes new versions of the app bundle that will be installed on next run. So ideally mole would inspect the .app files before deleting them an not delete any dated newer that the one pointed to by the alias.

buiducnhat:
## Problem

Autodesk Fusion's in-app updater can leave previous app bundles under `~/Library/Application Support/Autodesk/webdeploy/production`. Each release lives in a hash-named folder and the `Autodesk Fusion.app` alias identifies the current version. The reporter accumulated about 60 GB of old bundles (#1438).

## Solution

Add `clean_autodesk_fusion_old_bundles`, called from `clean_3d_tools`, with deletion evidence bound at every stage:

1. Resolve and stabilize the current Fusion alias inside the real production root; fail closed when it is missing, changes, or escapes containment.
2. Consider only direct lowercase 40-hex version directories containing a verified Fusion bundle, exact bundle ID, numeric bundle version, and executable.
3. Keep the current version, equal versions, every newer staged version, symlinks, and anything unrecognized. Remove only strictly older verified versions.
4. Recheck Autodesk/updater process state, current-alias identity, candidate metadata, parent identity, and target identity at the final deletion sink.
5. Preserve timeout/signal cancellation and report partial removal failures accurately.

## Testing

- Full Bats suite: 1,647 passed, 6 existing environment-dependent skips, 0 failures.
- Relevant cleanup and file-operation suites on the latest-main merge tree: 338/338 passed.
- Go, integration, installation, formatting, ShellCheck, syntax, and duplication checks passed.

Closes #1438

### DIFF
--- a/lib/clean/app_caches.sh
+++ b/lib/clean/app_caches.sh
@@ -716,6 +716,172 @@ _autodesk_cache_delete_guard_allows() {
     mole_clean_process_guard autodesk_cache_process_state "Autodesk running"
 }
 
+# Autodesk Fusion's in-app updater keeps every previous app bundle under
+# ~/Library/Application Support/Autodesk/webdeploy/production. An alias named
+# "Autodesk Fusion.app" points at the currently installed version. Bundles
+# newer than that alias are staged for the next launch and must be kept (#1438).
+clean_autodesk_fusion_old_bundles() {
+    local production_dir="$HOME/Library/Application Support/Autodesk/webdeploy/production"
+    [[ -d "$production_dir" ]] || return 0
+
+    local current_alias="$production_dir/Autodesk Fusion.app"
+    if [[ ! -e "$current_alias" ]]; then
+        debug_log "Autodesk Fusion old bundles: no current alias at $current_alias"
+        return 0
+    fi
+
+    local current_target=""
+    if [[ -L "$current_alias" ]]; then
+        current_target=$(readlink "$current_alias" 2> /dev/null || true)
+    else
+        # Alias / Finder alias: resolve via AppleScript; fall back to pwd -P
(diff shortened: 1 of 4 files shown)


tw93:
Thanks for finding and implementing the cleanup path for #1438. I kept the contribution and pushed a maintainer safety pass to the same branch: candidates now require exact Fusion bundle/version evidence, current/equal/all newer staged versions are retained, and updater state, alias identity, and target identity are rechecked at the final deletion sink.

Verification is clean: the full Bats suite passed (1,647 passed, 6 environment-dependent skips), the latest-`main` merge tree passed 338/338 relevant safety tests, and all Go/integration/install/static checks passed.

Issue #1447, opened 2026-08-20
[BUG] mo clean unloads protected Chrome Remote Desktop broker and leaves host offline


nathankim0:
## Describe the bug

`mo clean` can leave Chrome Remote Desktop offline even after the registration-file fix from #808.

The host registration files and launchd plists remain on disk, but the system broker job is unloaded. The user LaunchAgent then starts `remoting_me2me_host`, which exits with `INITIALIZATION_FAILED` because the Mach service `chromoting.agent_process_broker_mojo_ipc` no longer exists.

This is not a missing-whitelist-file problem: Mole already contains `org.chromium.chromoting*` in `DATA_PROTECTED_BUNDLES`. The failure is caused by the order and protection-mode mismatch in the orphaned-system-service cleanup.

## Steps to reproduce

1. Set up Chrome Remote Desktop on macOS and verify the machine is online.
2. Verify both jobs are loaded:
   - `system/org.chromium.chromoting.broker`
   - `gui/$UID/org.chromium.chromoting`
3. Run `mo clean` with administrator access.
4. Verify these files still exist:
   - `/Library/LaunchDaemons/org.chromium.chromoting.broker.plist`
   - `/Library/LaunchAgents/org.chromium.chromoting.plist`
   - `/Library/PrivilegedHelperTools/org.chromium.chromoting.json`
   - `/Library/PrivilegedHelperTools/org.chromium.chromoting.me2me_enabled`
5. Run `launchctl print system/org.chromium.chromoting.broker` and observe that the job is no longer loaded.
6. Open `https://remotedesktop.google.com/access`; the machine is shown as offline.

## Expected behavior

A protected, installed Chrome Remote Desktop service must not be classified as orphaned or unloaded. If Mole decides not to remove a service plist, it must leave the corresponding launchd job in its original state.

## Debug logs

The Chrome Remote Desktop unified log after restarting the user LaunchAgent shows:

```text
All permissions granted!
Parsing new host configuration.
Applying new host configuration.
Suspending process to wait for broker outcome
WARNING: Cannot connect to IPC through server name chromoting.agent_process_broker_mojo_ipc. Endpoint is invalid.
ERROR: Failed to connect to agent process broker.
SendHostOfflineReason: sending INITIALIZATION_FAILED.
Host returned non-permanent exit code INITIALIZATION_FAILED (2)
```

After manually bootstrapping the existing broker plist and restarting the user LaunchAgent, all three processes remain running and the web UI immediately returns to `Online`:

[...]

PR #1448, opened 2026-08-20
fix(clean): keep standalone helper app services loaded

nathankim0:
## Summary

- Treat an existing executable nested inside a standalone PrivilegedHelperTools .app as ownership evidence.
- Prevent the orphan-service cleanup from resolving it as a missing parent app and unloading its LaunchDaemon.
- Add a Chrome Remote Desktop broker regression test that verifies parent resolution, launchctl unload, and removal are never reached.

Addresses #1447

## Testing

- ./scripts/check.sh
- MOLE_TEST_NO_AUTH=1 bats tests/clean_apps.bats tests/clean_hints.bats (74 passed)
- Focused regression test verified red before the fix and green after it
- MOLE_TEST_NO_AUTH=1 ./scripts/test.sh (1648 tests; 3 failures in untouched local/environment-sensitive cloud, Spotlight, and browser tests; the browser failure passed when rerun alone)

### DIFF
--- a/lib/clean/apps.sh
+++ b/lib/clean/apps.sh
@@ -1174,6 +1174,16 @@ clean_orphaned_system_services() {
         # is gone, the binary itself is orphaned, so this plist is too. See #1082.
         if [[ "$binary_exists" == "true" ]]; then
             if [[ "$binary" == /Library/PrivilegedHelperTools/* ]]; then
+                # Some services are installed as standalone helper apps with no
+                # parent under /Applications. Their existing nested executable
+                # is exact ownership evidence, so running the parent-app resolver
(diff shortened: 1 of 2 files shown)


tw93:
@nathankim0 thank you for the precise reproduction and focused fix. The standalone helper-app guard is fail-safe and fixes the Chrome Remote Desktop path before it can reach launchctl or removal. I validated the exact merge result against current main: 1,636 Bats tests passed with 0 failures, plus Go, module-loading, integration, and installation checks. I am keeping #1447 open because the broader unload/remove transaction needs a separate launchd-domain state design; this PR does not pretend to solve that second part. Merging the focused fix now.

PR #1454, opened 2026-08-24
feat(clean): allow tuning system cleanup time budget

matheus-apolonio:
## Summary
- add a bounded `MOLE_TIMEOUT_SYSTEM_CLEANUP_SEC` override for the privileged system-cleanup section
- retain the 120-second default and restrict overrides to 30–600 seconds
- document the one-off slow-filesystem workaround and cover default, valid, and invalid values

The current fixed two-minute budget can be insufficient for a healthy Mac with a large diagnostics tree. A global bound remains necessary so a stalled privileged scan cannot appear hung indefinitely.

## Validation
- `bash -n lib/core/timeouts.sh lib/clean/system.sh`
- direct helper checks for default, 600-second override, and invalid values
- `./scripts/check.sh --format` (formatter unavailable in this environment)

`bats` is not installed in this environment, so the Bats suite could not be invoked here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ Prefer a walkthrough? Watch the [Mole tutorial video](https://www.youtube.com/wa
 Mole can remove files, so it validates paths, protects shared and system-owned locations, and asks for confirmation when an action needs it. When Mole cannot prove an item is safe to change, it skips or refuses it.
 
 - `clean`, `uninstall`, `purge`, `installer`, and `remove` can delete files. Review them with `--dry-run` first, and add `--debug` when needed.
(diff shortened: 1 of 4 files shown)


tw93:
@matheus-apolonio Thanks for the patch, and for matching the project conventions this closely. The clamp is correct and the heredoc assertions end in `|| exit 1`, which is the shape this repo needs and often gets wrong.

I am closing it on product grounds rather than code quality. Mole treats a new environment variable as the same weight as a new setting: it earns its place only when no single default can serve everyone, and when fixing the default has been considered and rejected first. Today the 120-second budget degrades rather than fails, the section prints `time limit reached, skipped remaining slow scans` and a second `mo clean` picks up from there, so a slow disk costs an extra run rather than losing anything. Without a report of that being insufficient on a real machine, the override adds a surface I would have to keep supporting.

If you do hit it, open an issue with the timings and I will raise the default or make the degraded message clearer. That is the direction I would take instead of an override.

PR #1460, opened 2026-08-25
fix(purge): exclude node_modules/vendor/Pods from project container detection

### LINKED ISSUE #1459: [BUG] mo purge deletes dist/ inside node_modules, leaving packages half-installed
## Summary

`mo purge` queues `node_modules/<pkg>/dist` (and `build`, `coverage`, …) for deletion when a `node_modules` directory sits where `discover_project_dirs` globs it as a *container*. Removing those directories leaves the package half-installed: `package.json` and the rest of the package stay in place, so a plain `npm install` does not necessarily put the deleted files back.

This contradicts the contract the repo documents for `purge` in `.claude/skills/mole/SKILL.md:82`:

> It deliberately does not touch anything that needs a network to restore (`node_modules/`, `Pods/`, `venv/`), so a purge is always recoverable with a local rebuild.

## Observed

On a real machine with a stray `~/node_modules`, `mo purge --dry-run` listed 19 package-internal `dist` directories:

```
✓ [DRY RUN] ~/node_modules/markdown-it/dist, 598KB
✓ [DRY RUN] ~/node_modules/khroma/dist, 422KB
✓ [DRY RUN] ~/node_modules/uuid/dist, 184KB
✓ [DRY RUN] ~/node_modules/@upsetjs/venn.js/build, 168KB
✓ [DRY RUN] ~/node_modules/d3-geo/dist, 127KB
...
```

Note that `~/node_modules` itself is **not** in the list. Only its children are.

## Root cause

`is_project_container` (`lib/clean/project.sh:37`) rejects `Library`, `Applications`, `Movies`, `Music`, `Pictures`, `Public`, and dot directories, but has no `node_modules` exclusion. Every npm package ships a `package.json`, which is the first entry in `MOLE_PURGE_PROJECT_INDICATORS`, so the `-maxdepth 2` probe matches immediately.

The chain:

1. `discover_project_dirs` globs `$HOME/*/` and reaches `$HOME/node_modules`.
2. `is_project_container "$HOME/node_modules" 2` finds `node_modules/markdown-it/package.json` → returns 0.
3. Each `node_modules/<pkg>/` is then treated as a project root.
4. `dist` and `build` are in `MOLE_PURGE_TARGETS`, so `node_modules/<pkg>/dist` is emitted at depth 1 of that "project".
5. `filter_nested_artifacts` cannot collapse these into the parent, because `$HOME/node_modules` was never emitted as a target — the scan started *below* it.

Inside an ordinary project (`~/dev/app/node_modules/pkg/dist`) the behavior is correct: `node_modules` is emitted as an artifact of the project root and the nested filter drops the inner `dist`. The defect is specific to `node_modules` being classified as a container rather than as an artifact.

## Reproduction

```bash
TMPHOME=$(mktemp -d)
mkdir -p "$TMPHOME/node_modules/markdown-it/dist"
printf '{"name":"markdown-it","version":"14.1.0"}' > "$TMPHOME/node_modules/markdown-it/package.json"

# is_project_container returns 0 for the node_modules directory
```

With the function extracted verbatim from `lib/clean/project.sh:37` and `PROJECT_INDICATORS=("${MOLE_PURGE_PROJECT_INDICATORS[@]}")`, `is_project_container "$TMPHOME/node_modules"` returns 0.

## Impact

`npm` treats a package as installed based on the presence of its directory and the version in `package.json`. With `dist/` gone but `package.json` intact, `npm install` can report the tree as up to date and leave the package broken; `require()` / `import` then fails at runtime. Recovery needs `npm ci` or deleting `node_modules` outright — a network restore, which is exactly what `purge` promises never to require.

## Suggested fix

Reject package-manager directories as containers in `is_project_container`, alongside the existing `Library` / `Applications` entries:

```bash
[[ "$basename" == "node_modules" ]] && return 1
```

`vendor` (PHP Composer) and `Pods` (CocoaPods) have the same shape — `vendor/<vendor>/<pkg>/composer.json`, `Pods/<Pod>/` — and are worth checking as part of the same fix. `is_protected_purge_artifact` already special-cases `vendor` when it is a *target*, but not when it is a *container*.

A regression test would assert that a `node_modules` tree containing packages with `dist/` produces no purge candidates below `node_modules`.

## Environment

- Mole 1.51.0 (Homebrew)
- macOS 26.0.1, arm64

buiducnhat:
## Summary

`is_project_container` in `lib/clean/project.sh` returned `0` for `node_modules` directories because they contain `package.json` (the first entry in `MOLE_PURGE_PROJECT_INDICATORS`). This caused `mo purge` to delete `dist/` and `build/` directories inside individual packages, leaving them half-installed.

## Root Cause

The container exclusion list covers `Library`, `Applications`, `Movies`, `Music`, `Pictures`, `Public`, and dot-directories, but had no entry for `node_modules`, `vendor`, or `Pods`. When `discover_project_dirs` globs `$HOME/*/`, any `node_modules` directory at `$HOME` level is treated as a project container, causing each child package to be scanned as a project root.

## Fix

Add `node_modules`, `vendor`, and `Pods` to the exclusion list in `is_project_container`. This ensures package-manager directories are not treated as project containers, so they fall through to the existing artifact/nested-filtering logic.

## Verification

- `node_modules` at any level is now rejected as a project container
- `vendor` (PHP Composer) and `Pods` (CocoaPods) are excluded for the same reason
- Backward compatible: the existing `is_protected_purge_artifact` logic already handles these as targets

Closes #1459

### DIFF
--- a/lib/clean/project.sh
+++ b/lib/clean/project.sh
@@ -51,6 +51,12 @@ is_project_container() {
     [[ "$basename" == "Music" ]] && return 1
     [[ "$basename" == "Pictures" ]] && return 1
     [[ "$basename" == "Public" ]] && return 1
+    # Package-manager directories: they contain package.json/composer.json
+    # but are not project containers; deleting their children leaves
+    # packages half-installed and requires a network restore.
+    [[ "$basename" == "node_modules" ]] && return 1
+    [[ "$basename" == "vendor" ]] && return 1
+    [[ "$basename" == "Pods" ]] && return 1
 
     # Single find expression for indicators.
     local -a find_args=("$dir" "-maxdepth" "$max_depth" "(")


tw93:
@buiducnhat thanks for catching this container shape. The underlying bug is already fixed on `main` through #1459 by deriving the exclusion from the complete `PURGE_TARGETS` list, with regressions for both container classification and emitted candidates, so this hard-coded three-name variant is superseded and no longer needs to merge.

PR #1464, opened 2026-08-26
Speed up clean and purge scans

M-Hassan-Raza:
Clean and purge were doing a lot of repeated filesystem work on larger project trees. This pr prunes nested purge hits earlier, reuses scan, size and process evidence, and runs independent project cache scans concurrently while keeping partial timeout output out of deletion paths.

On my machine clean --dry-run went from a median 87.15s to 76.43s. In a controlled fixture with identical candidates it went from 25.87s to 21.35s. The largest purge root scan went from 30.10s to 9.81s, and the full purge dry-run now completes in about 286-297s where main did not finish within 608s.

Purge still revalidates the configured root, physical root, candidate identity and recent activity at the final removal boundary.

### DIFF
--- a/lib/clean/caches.sh
+++ b/lib/clean/caches.sh
@@ -576,15 +576,19 @@ clean_project_caches() {
 
     local scan_index
     for ((scan_index = 0; scan_index < ${#root_matches_files[@]}; scan_index++)); do
-        root_matches_file="${root_matches_files[$scan_index]}"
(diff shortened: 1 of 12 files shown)


tw93:
Thanks—this is merged. I kept the performance work and added maintainer follow-ups for shared deadlines, cancellation cleanup, and final live-cache rechecks. All checks are green.

PR #1477, opened 2026-08-29
docs(status): document zombie metrics and attribution

r266-tech:
Follow-up to #1476.

## Summary

- show the new zombie row in the `mo status` example
- document the `zombie_count`, bounded `zombie_parents`, and
  `zombie_parents_complete` JSON fields
- clarify the read-only behavior and the distinction between unmeasured, measured zero,
  and incomplete parent attribution

## Verification

- parsed the updated `mo status --json` example with Python's JSON parser
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,10 +313,11 @@ $ mo status --json
 }
 ```
 
-Zombie diagnostics are read-only and do not affect the health score or terminate processes. If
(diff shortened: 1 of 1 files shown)


tw93:
@r266-tech, thank you for following up with the documentation. I clarified the cached watch-sample and parent-attribution semantics and merged it, so the README now matches the Nightly behavior from #1476 and will be included in the next stable release.

PR #1486, opened 2026-08-31
feat: add triage command to diagnose a bogged-down Mac

tszaks:
## What

`mo triage` answers the question users actually arrive with — "why is my Mac so slow?" — when `status` shows the metrics but not the cause. It detects five specific failure classes that make a Mac crawl while nothing obvious is wrong, and explains each finding in plain language:

- **Swap exhaustion** — swap used vs. total with warn/critical thresholds. This is the headline number when "everything is slow": high load with few busy processes is memory wait, not compute.
- **Runaway system daemons** — a known-safe respawnable set (ControlCenter, fileproviderd, bird, CursorUIViewService, NotificationCenter) plus WindowServer as report-only. Because `ps %cpu` on Darwin is a lifetime average (cputime ÷ elapsed), a long-lived process showing a high value has been stuck the whole time; the detector cross-checks against a live `top -l 2` second sample so it only flags daemons that are both historically and currently burning.
- **Leaked automation browsers** — headless Chrome trees orphaned by dead Playwright/agent/MCP sessions (daemons reparented to launchd, Chrome processes on ephemeral `playwright_chromiumdev_profile-*` temp profiles older than a day), plus stale profile dirs in the user temp dir. On the machine this was built against, one accumulation reached 248 processes holding 5.9 GB.
- **iCloud sync spiral** — build artifacts (`node_modules`, `dist`, `.next`, `.vercel`, `.turbo`) under iCloud-synced Desktop/Documents make `fileproviderd` churn indefinitely: rewrites mid-upload spawn `name 2` conflict duplicates, which create more files to sync, which create more conflicts. Detected by combining artifact count, conflict-dup count, and fileproviderd heat; only fires when Desktop & Documents sync is actually enabled.
- **Stale CloudStorage domains** — leftover duplicate provider domains (e.g. `GoogleDrive-x (5-29-26 6:50 PM)`) that keep a second sync engine watching the filesystem.

## Behavior

- `mo triage` is **always read-only**. A test pins that a default run never prints an action verb.
- `mo triage --fix` applies only the safe remedies, each category confirmed separately (`--yes` to skip): restart respawnable daemons via SIGTERM (launchd brings them back clean), kill orphaned browser trees one PID at a time with per-PID failure counting (a bulk kill of a long list can silently no-op), and remove stale temp profiles through `safe_remove` after a `pgrep` live-use check.
- The **iCloud spiral is deliberately guidance-only**: the fix path prints an ordered manual playbook instead of automating. Renaming a direct child of the CloudDocs sync root blocks in `rename(2)` indefinitely (full-domain reconciliation), and a corrupted local index (`brctl` showing endless `itemNotFound` retries) needs a reboot before any move completes. Automating around those failure modes is not safe; explaining them is the product value.

## Product decision filter

1. **Which command family?** A new one, closest kin to `status` — but `status` is pinned as a read-only dashboard that "should not become an alerting daemon", and diagnosis-plus-gated-remediation is a different contract. It follows the same shape as `optimize`: explicit, bounded, explainable actions.
2. **Safe by default, previewable, testable without auth?** Detection is read-only and is the default; every fix is per-category confirmed; nothing uses sudo, osascript, or launchctl; the suite runs under `MOLE_TEST_NO_AUTH=1` with no real auth paths.
3. **Can the user verify before Mole changes anything?** Yes — the default run is the preview, findings name exact PIDs/paths/counts, and `--fix` re-states each action at its confirmation.
4. **Is the target rebuildable/disposable?** Process kills target orphans and launchd-respawnable daemons; the only deletion is ephemeral automation profiles in the user temp dir, routed through `safe_remove` (rest of user data is never touched — the one risky domain gets prose, not code).
5. **Better as docs or a warning?** The detection genuinely requires cross-referencing live process state, so it can't be a doc; the one part that is better as guidance (iCloud) ships as guidance.

On the config-knob rule: the three `MOLE_TRIAGE_*` threshold overrides follow the existing `MOLE_OPTIMIZE_DIAG_CPU_THRESHOLD` precedent in `lib/optimize/diagnostics.sh` (same validation shape). Happy to strip them to bare constants if you'd rather not grow that surface.

## Testing

- `./scripts/check.sh` clean (shfmt, ShellCheck, go vet, module loading, function-duplication gate).
- `MOLE_TEST_NO_AUTH=1 ./scripts/test.sh` — full suite green.
- New `tests/triage.bats` (7 cases): help, unknown-option rejection, read-only default pinned by absence of action verbs, detectors running standalone under `set -euo pipefail` on `/bin/bash` 3.2, non-numeric threshold override rejection, fix helpers no-op on empty findings, dispatch wiring.
- Verified live on a machine mid-incident: correctly identified a runaway `fileproviderd`, 10 artifact dirs under synced folders with 202 conflict duplicates, and a stale Google Drive domain, while reporting healthy swap as healthy.

### DIFF
--- a/bin/triage.sh
+++ b/bin/triage.sh
@@ -1,0 +1,122 @@
+#!/bin/bash
+# Mole - Triage command.
+# Diagnoses why the Mac feels bogged down and applies safe fixes.
+# Detection is read-only. --fix authenticates once (password or Touch ID),
+# then applies every safe fix and reports what it did.
+
+set -euo pipefail
+
+export LC_ALL=C
+export LANG=C
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+source "$SCRIPT_DIR/lib/core/common.sh"
+source "$SCRIPT_DIR/lib/core/sudo.sh"
+source "$SCRIPT_DIR/lib/triage/detect.sh"
+source "$SCRIPT_DIR/lib/triage/fix.sh"
+
+trap 'stop_sudo_session 2> /dev/null || true' EXIT INT TERM
+
+TRIAGE_APPLY_FIXES=false
+TRIAGE_ASSUME_YES=false
+
+show_triage_help() {
+    echo "Usage: mo triage [OPTIONS]"
+    echo
+    echo "Diagnose why this Mac is slow: swap exhaustion, runaway system daemons,"
+    echo "leaked automation browsers, iCloud sync spirals, stale cloud-sync domains."
+    echo
+    echo "Options:"
+    echo "  --fix             Authenticate once, apply all safe fixes, report results"
+    echo "  --yes             With --fix: skip the authentication gate (scripting)"
+    echo "  -h, --help        Show this help"
+    echo
+    echo "Detection is always read-only and only reports problems; a healthy check"
+    echo "prints nothing. Fixes never touch user data: they restart respawnable"
+    echo "daemons, kill orphaned automation browsers, and remove stale browser"
+    echo "profiles from the user temp dir. The iCloud spiral gets manual migration"
+    echo "steps under --fix instead of an autofix; automating it is not safe."
+}
+
+for arg in "$@"; do
+    case "$arg" in
+        --fix) TRIAGE_APPLY_FIXES=true ;;
+        --yes | -y) TRIAGE_ASSUME_YES=true ;;
+        -h | --help)
+            show_triage_help
+            exit 0
+            ;;
+        *)
+            echo "Unknown option: $arg"
+            show_triage_help
+            exit 1
+            ;;
+    esac
+done
+
+main() {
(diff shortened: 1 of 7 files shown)


tszaks:
Reworked the output to match the house grammar after using it on a real machine: findings-only DIAGNOSIS block (healthy checks stay silent, like optimize's PERFORMANCE DIAGNOSIS), one dense ⚙ System stat line, gray ⊙ sub-hints instead of prose, ➤ action sections under --fix, and the iCloud migration steps now render only under --fix when the spiral was actually detected. Tests updated to pin the findings-only grammar (9 cases).

tszaks:
Withdrawing this — after living with it, a separate command is the wrong shape. The genuinely cleanable part (leaked automation browser processes and stale profiles) belongs inside mo clean's existing surface, per the project's own decision filter. May follow up with that much smaller change.

PR #1494, opened 2026-09-01
fix(windows): pass install directory by name during update

transiteration:
I ran into this while using `mo update` on Windows. Instead of refreshing the existing installation, the installer created a directory literally named `-InstallDir`.

The updater was passing the installer arguments through an array, so PowerShell treated them as positional values. This changes it to hashtable splatting so `InstallDir` and `AddToPath` are passed by name.

I also added a regression test for this case!

Thanks

I tested with PowerShell 7 and Pester 5.9.1: 101 tests passed.

### DIFF
--- a/bin/update.ps1
+++ b/bin/update.ps1
@@ -218,13 +218,13 @@ else {
     Write-Host "Updated source: $before -> $after" -ForegroundColor Green
 }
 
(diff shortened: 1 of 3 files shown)


transiteration:
thx! @youxi798 @tw93

PR #1501, opened 2026-09-02
feat(clean): clean Feishu/Lark embedded webview Service Worker cache (~28GB reclaimed)

YeJe-cpu:
### What

`mo clean` now clears the Service Worker `CacheStorage` of Lark / Feishu desktop's embedded Chromium webview, at:

```
~/Library/Application Support/LarkShell/aha/users/*/profile_explorer/Service Worker/CacheStorage
```

This routes through the existing `clean_service_worker_cache` helper — the same one already used for Chrome and Arc — so no new deletion logic is introduced.

### Why

Lark / Feishu desktop embeds a Chromium webview (the `aha` "explorer" profile) to render its in-app docs, sheets, slides, and AI surfaces. Every workspace origin the user opens precaches the full editor bundle into that profile's Service Worker `CacheStorage`, and stale workspaces plus superseded precache versions are never evicted — so it grows without bound at a non-standard path the browser cleaners never reach. This is the same class of problem as #843 (Final Cut Pro) and #1277 (JianyingPro): heavy generated cache outside the standard cache dirs.

### Measured value (real machine, macOS 15.7)

| Fact | Value |
|---|---|
| Total `profile_explorer/Service Worker/CacheStorage` | **27 GB** (main account) + 1.2 GB (second account) |
| Origin buckets | 9 (one per workspace/origin ever opened) |
| Oldest cached files | ~11 months (2025-10), never evicted |
| Growth | accumulating monthly, 10,916 files added in the last month alone |
| Cached origins (extracted from index) | `ask.feishu.cn`, `bytedance.larkoffice.com`, several `*.feishu.cn` tenants |
| Precache buckets | `docx`, `bitable`, `slides`, `base_app`, `knowledge_ai` — the doc-editor bundles |

Removing it reclaimed ~28 GB; afterwards Feishu docs, sheets, chat, and login all worked normally (the editor bundle is simply re-precached on next open).

### Safety / non-targets

- Only `CacheStorage` is targeted — never the sibling `ScriptCache` (MV3 worker bytecode), same as the Chrome/Arc treatment.
- Documents live on Lark's servers; auth lives in `Cookies` / `Local Storage`, neither of which this path touches.
- The shared helper already honors `PROTECTED_SW_DOMAINS` and the user whitelist, and deletes via `safe_remove` (Trash-routed, logged, path-protected). `feishu.cn` / `larkoffice.com` are not in `PROTECTED_SW_DOMAINS`, consistent with how the existing cleaner treats hash-keyed origin buckets.
- No-op when the `aha/users` root is absent; a `declare -f` guard skips cleanly if the shared helper is not loaded.

### Tests

`tests/clean_app_caches.bats`:
- `clean_communication_apps` calls the shared cleaner once per `profile_explorer` profile with the correct `CacheStorage` path, and ignores non-profile siblings under `users/`.
- `clean_feishu_service_worker_caches` is a no-op when the profile root is absent.

`./scripts/check.sh` passes (shellcheck, syntax, function-duplication audit). The two tests were mutation-checked (a wrong path turns the assertion red).

Happy to adjust naming or gate on a running check if you'd prefer — I mirrored the Chrome treatment, which cleans `CacheStorage` unconditionally.

### DIFF
--- a/lib/clean/app_caches.sh
+++ b/lib/clean/app_caches.sh
@@ -434,14 +434,72 @@ clean_code_editors() {
 # the sibling ScriptCache (MV3 worker bytecode). Documents live on Lark's
 # servers and auth lives in Cookies / Local Storage, neither of which this path
 # touches; a cleared editor bundle is re-precached on next open.
+feishu_or_lark_running() {
+    mole_pgrep_any \
+        -x "Feishu" \
+        -x "Lark" \
+        -f '/(Feishu|Lark)[.]app/'
+}
+
+_feishu_service_worker_delete_guard_allows() {
+    mole_clean_process_guard feishu_or_lark_running "Feishu or Lark started"
+}
+
 clean_feishu_service_worker_caches() {
-    declare -f clean_service_worker_cache > /dev/null 2>&1 || return 0
-    local sw_root="$HOME/Library/Application Support/LarkShell/aha/users"
-    [[ -d "$sw_root" ]] || return 0
-    local _profile
-    for _profile in "$sw_root"/*/profile_explorer; do
-        [[ -d "$_profile" ]] || continue
-        clean_service_worker_cache "Feishu" "${_profile%/}/Service Worker/CacheStorage"
+    local -a sw_roots=(
+        "$HOME/Library/Application Support/LarkShell/aha/users"
+        "$HOME/Library/Application Support/LarkInternational/aha/users"
+    )
+    local -a root_labels=("Feishu" "Lark")
+    local -a cache_paths=()
+    local -a cache_labels=()
+    local root_index sw_root app_label physical_root _profile cache_path physical_cache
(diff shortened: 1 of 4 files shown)


YeJe-cpu:
Thanks a lot @youxi798 for the hardening pass here 🙏

I'd mirrored the older Chrome Service Worker path, which quietly carried its looser habits into a new deletion sink. Your commit is exactly the safety layer it was missing: the tri-state process guard that defers instead of deleting a live app's files, the physical-path containment plus TOCTOU identity snapshot at the sink, and threading the shared timeout deadline through find/du/remove so a truncated scan can never feed a partial delete plan. Extending it to LarkInternational was a nice catch too.

I read the whole diff and the five new guard tests carefully — learned a lot about how Mole holds destructive paths to a higher bar than "it's just a cache". Appreciated getting to contribute this one.

PR #1504, opened 2026-09-02
feat(optimize): diagnose memory pressure, idle VMs, and stuck processes

tszaks:
## What

`PERFORMANCE DIAGNOSIS` currently scans five hardcoded CPU families (`cloudshell`, `syspolicyd`, `windowserver`, `spotlight`, `coresim_disk_images`). That structure means it cannot see the two failure shapes that most often make a Mac feel slow. This adds both, plus a third, all inside the existing section — no new command, no new menu entry.

**1. Memory pressure.** A high load average with nothing visibly busy is swap thrash, not compute. The diagnosis now reports swap percentage and, critically, *names the processes over 1GB actually holding the memory*, so the user gets a cause instead of a symptom.

**2. Idle virtual machines.** Docker Desktop's Linux VM claims its full memory reservation even with zero containers running — 8.2GB on the machine this was developed against. Completely invisible to disk-oriented checks, and the single largest reclaimable resource on a developer Mac. It only calls that memory reclaimable when the owner confirms no containers are running; if `docker ps` cannot answer, it reports the VM's size and makes no recommendation.

**3. Stuck processes.** A process wedged in a run loop shows up as high CPU-time over total elapsed time, which is exactly what `ps TIME`/`ELAPSED` measure. The motivating incident was `ControlCenter` at 139h of CPU across 392h of uptime while a healthy `ControlCenter` sits near 0%.

All three are read-only and print nothing when healthy, so a clean machine still gets the single reassuring `No sustained high-CPU bottleneck detected` line.

## Threshold rationale (the interesting part)

I first set the runaway floor at 50% and it **missed the very case that motivated the feature**. 139h over 392h is 35%. The floor is therefore 25%, which catches both real incidents observed:

| Process | CPU time | Uptime | Sustained |
|---|---|---|---|
| `ControlCenter` | 139h | 392h | 35% |
| `fileproviderd` | 41.6h | 62h | 67% |

Guards against false positives: requires 12h+ of runtime so a momentary spike cannot trip it, skips `kernel_task` (thermal management legitimately accumulates CPU), and skips the families the CPU scan above already headlines so one problem is never reported twice under two names.

## Product decision filter

1. **Which command family?** `optimize` — it already owns `PERFORMANCE DIAGNOSIS`. This deepens that surface rather than adding one.
2. **Safe by default, previewable, testable without auth?** Read-only; no `sudo`/`osascript`/`launchctl`; the whole suite runs under `MOLE_TEST_NO_AUTH=1`.
3. **Can the user verify before Mole changes anything?** Nothing is changed. Findings name exact processes, sizes, and PIDs.
4. **Rebuildable/disposable target?** No deletion involved.
5. **Better as docs or a warning?** It requires live process correlation, so it cannot be a doc — and the one part that is advice (quit Docker Desktop) ships as advice, not an action.

On the config-knob rule: the `MOLE_OPTIMIZE_*` variables added here are **test-injection seams**, matching the existing `MOLE_OPTIMIZE_PS_SAMPLE_1/2` precedent in the same file, not user-facing tuning. They exist because reading live system state made three *existing* `optimize.bats` assertions environment-dependent — they assert a quiet run, and my dev machine genuinely sat at 91% swap, so they failed for a reason unrelated to the code under test. Those three setups now pin memory state the same way they already pin CPU samples. Happy to collapse the threshold constants to bare values if you'd rather have fewer.

## Testing

- `./scripts/check.sh` clean (shfmt, ShellCheck, go vet, module loading, function-duplication gate).
- `MOLE_TEST_NO_AUTH=1 bats tests/optimize.bats tests/optimize_db.bats tests/optimize_perf_diagnosis.bats` — 115 passing.
- New `tests/optimize_perf_diagnosis.bats`, 12 cases: time parsing across `mm:ss` / `hh:mm:ss` / `dd-hh:mm:ss`, garbage input returning 0 so it cannot fabricate a runaway, memory silent when healthy, memory naming holders when exhausted, idle-VM reporting, idle-VM *not* claiming reclaimable while containers run, runaway firing on the real `ControlCenter` shape, and four separate false-positive guards.
- Verified live: correctly reported `swap 23GB of 25GB used (95%)` with the actual top holders named, and stayed silent on the idle-VM check once Docker was quit.

One bug worth flagging since it is easy to repeat: I initially stripped the `ps` header with `awk 'NR>1'` *after* `sort -rn`. Because the non-numeric header sorts to the bottom, that filter silently discarded the **largest** process — the single row that matters most. Fixed by stripping the header before sorting; pinned by the `[[ "$output" != *"COMM"* ]]` assertion.

### DIFF
--- a/lib/clean/project.sh
+++ b/lib/clean/project.sh
@@ -144,9 +144,10 @@ write_purge_config() {
     local tmp_file
     tmp_file=$(mktemp_file "mole-purge-paths") || return 1
 
-    if ! cat > "$tmp_file" << EOF; then
+    if ! cat > "$tmp_file" << EOF
 $header
 EOF
+    then
         rm -f "$tmp_file" 2> /dev/null || true
         return 1
     fi
--- a/lib/optimize/diagnostics.sh
+++ b/lib/optimize/diagnostics.sh
@@ -392,8 +392,20 @@ run_optimize_diagnostics() {
         fi
     done
 
-    if [[ -z "$primary_family" ]]; then
+    # Memory and stuck-process checks run alongside the family CPU scan. They
+    # cover the two shapes the family list cannot see, and each stays silent
+    # when healthy, so a clean machine still prints one reassuring line below.
+    local extra_findings=0
+    local mem_out runaway_out vm_out
+    mem_out=$(opt_diag_memory_pressure) || true
+    vm_out=$(opt_diag_idle_vm) || true
+    runaway_out=$(opt_diag_runaway_process) || true
+    [[ -n "$mem_out" || -n "$vm_out" || -n "$runaway_out" ]] && extra_findings=1
+
+    if [[ -z "$primary_family" && "$extra_findings" -eq 0 ]]; then
         echo -e "  ${GREEN}${ICON_SUCCESS}${NC} No sustained high-CPU bottleneck detected"
+    elif [[ -z "$primary_family" ]]; then
+        : # extra findings print below; no CPU-family bottleneck to headline
     else
         label=$(opt_diag_family_label "$primary_family")
         echo -e "  ${YELLOW}${ICON_WARNING}${NC} Likely bottleneck: ${label} (~${primary_avg}% CPU sustained)"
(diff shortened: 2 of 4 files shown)


youxi798:
@tszaks thank you for the contribution.

Merged on `main` via #1516. Memory pressure, idle VM, and runaway-process diagnosis are live under optimize PERFORMANCE DIAGNOSIS. I softened the idle-VM line so it does not assume every Virtualization.framework VM is Docker Desktop.

Pick it up with `mo update --nightly`.

PR #1505, opened 2026-09-02
feat(clean): reclaim leaked automation browsers and surface unused simulator runtimes

tszaks:
## What

Two additions to the Developer tools family in `mo clean`, both targeting leftovers that nothing on macOS ever reaps.

**1. Leaked automation browsers** (`clean_dev_automation_browsers`). When a Playwright/agent harness dies without cleanup, its browser daemon reparents to launchd and the headless Chrome tree keeps running, holding RSS indefinitely. One real machine accumulated **248 processes holding 5.9GB** plus 36 stale profile directories. It stops orphaned Playwright daemons (any age) and browser processes on `playwright_chromiumdev_profile-*` automation profiles older than a day, then removes stale profile temp dirs (2h+) through `safe_clean`.

A user's real browser can never match: only processes tied to ephemeral automation profiles are considered, and a profile a live process still references is skipped via `pgrep`. Browser processes need to be a full day old because a younger one may belong to an active automation session.

**2. Unused simulator runtimes** (`check_orphaned_simulator_runtimes`). An installed simulator runtime whose last device was deleted becomes unreferenced, and nothing reports it — not `simctl`, not Xcode, and not Mole's existing passes. `clean_xcode_simulator_runtime_volumes` handles stale mount points and the unavailable-simulator pass handles devices orphaned by a removed runtime; this is the inverse case. A real machine carried an unreferenced iOS 18.4 runtime at **8.2GB** indefinitely.

This one is **review-only**, matching how downloaded toolchain payloads are treated elsewhere in the classification rules: a runtime is only re-servable by Apple, so it stays off the blanket delete path. The value is naming the exact orphan, its size, and the owner command, which `simctl` cannot report directly.

## Safety

- Dry-run previews the process count and lets `safe_clean` preview the directories; it never signals a process.
- Deletion goes through `safe_remove`, never raw `rm -rf`.
- Both `simctl` probes are timeout-bounded via `MOLE_TIMEOUT_PKG_LIST_SEC` and **propagate cancellation** (124/128+) rather than reporting a false clean.
- Kills are sent one PID at a time with per-PID accounting, because a bulk `kill` over a long list can silently no-op.

## Testing

- `./scripts/check.sh` clean.
- `MOLE_TEST_NO_AUTH=1 bats tests/clean_dev_caches.bats tests/dev_extended.bats tests/clean_automation_browsers.bats tests/clean_orphaned_runtimes.bats` — 201 passing, including the required `lib/clean/dev.sh` hotspot suites.
- `tests/clean_automation_browsers.bats` (4 cases): kills only orphaned daemons and day-old automation browsers while a *fresh* automation session and an unrelated browser survive; dry-run signals nothing; stale profiles cleaned while in-use and fresh profiles survive; silent no-op when nothing leaked.
- `tests/clean_orphaned_runtimes.bats` (6 cases): reports a zero-device runtime with its size and owner command, stays silent when every runtime has a device, reports multiple orphans, handles a runtime with no reported size, propagates probe timeout instead of a false clean, and no-ops when `simctl` was never resolved.

One bug caught by these tests before it shipped, worth mentioning because it would have broken the common path: expanding the empty `orphan_lines` array under `set -u` is an unbound-variable error on the bash 3.2 macOS ships. Since "no orphans" is the normal case, that would have failed `mo clean` for everyone with a healthy machine. Guarded and pinned.

### DIFF
--- a/lib/clean/dev.sh
+++ b/lib/clean/dev.sh
@@ -2735,6 +2735,95 @@ _debug_simctl_probe_stderr() {
     debug_log "simctl probe $attempt stderr: $excerpt"
 }
 
+# Installed simulator runtimes that no device uses. Distinct from
+# clean_xcode_simulator_runtime_volumes (stale mount points) and from
+# unavailable-simulator cleanup (devices orphaned by a removed runtime):
+# this is the inverse, a runtime left behind after its last device went
+# away. Nothing in macOS or Xcode reports it, so an 8GB download can sit
+# unreferenced indefinitely.
+#
+# Review-only by design. A runtime is a toolchain payload that only Apple
+# can serve again, so it stays off the blanket delete path the same way
+# other downloaded toolchain roots do; the value here is naming the exact
+# orphan and the owner command, which is information the user cannot get
+# from simctl directly.
+check_orphaned_simulator_runtimes() {
+    command -v xcrun > /dev/null 2>&1 || return 0
+    [[ "${_MOLE_SIMCTL_RESOLUTION_STATUS:-}" == "ready" ]] || return 0
+
+    local runtime_list="" device_list="" probe_status=0
+    runtime_list=$(run_with_timeout "$MOLE_TIMEOUT_PKG_LIST_SEC" \
+        env DEVELOPER_DIR="$_MOLE_SIMCTL_DEVELOPER_DIR" xcrun simctl runtime list -v 2> /dev/null) || probe_status=$?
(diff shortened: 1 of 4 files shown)


youxi798:
@tszaks thank you for the contribution.

Merged on `main` via #1516: Playwright leaked automation browser cleanup and orphaned simulator runtime review both shipped in Developer tools.

Pick it up with `mo update --nightly`.

PR #1506, opened 2026-09-02
fix(uninstall): suppress dead-terminal countdown read errors

### LINKED ISSUE #1503: mo uninstall countdown spams "read: read error: 0: Input/output error" when the hosting terminal is gone
## Environment

- Mole 1.53.0 (Homebrew, `/opt/homebrew/Cellar/mole/1.53.0`)
- macOS 27.0, arm64
- Interpreter: `/bin/bash` 3.2.57 (stock macOS bash, per the `#!/bin/bash` shebang)
- Terminal: tmux pane

## What happened

After `mo uninstall` finishes removing the selected apps, it shows the 5-second countdown prompt:

```
Press Enter to return to the app list, press q to exit (5)
```

If the hosting shell/terminal goes away while the script is still in this loop (for example: the uninstaller is a suspended zsh job and the user force-exits the shell with a second `exit`, or the pane/tab is closed), each poll of `read` fails immediately with EIO and bash prints one error per countdown tick:

```
Press Enter to return to the app list, press q to exit (4) /opt/homebrew/Cellar/mole/1.53.0/libexec/bin/uninstall.sh: line 1832: read: read error: 0: Input/output error
Press Enter to return to the app list, press q to exit (3) ... read error: 0: Input/output error
... (repeats for every remaining tick)
```

The uninstall itself has already completed at this point — the impact is purely the error spam, but it looks alarming.

## Root cause

`bin/uninstall.sh`, the countdown loop (~line 1830):

```bash
while [[ $_countdown -gt 0 ]]; do
    printf "\r${GRAY}Press Enter to return to the app list, press q to exit (%d)${NC} " "$_countdown"
    if IFS= read -r -s -n1 -t 1 _key; then   # <-- no stderr suppression
        _pressed=true
        break
    fi
    ((_countdown--))
done
```

Two observations:

1. The existing `mole_tty_is_foreground` guard before the loop protects against SIGTTIN (reading while no longer the foreground process group). It does not cover the case where the session leader / controlling terminal is already gone — then `read(2)` returns EIO immediately instead of raising SIGTTIN, and bash's `read` builtin prints `read error: <fd>: <strerror>` to stderr on every iteration.

2. `drain_pending_input` in `lib/core/ui.sh` (line 293) already handles this exact situation with `2> /dev/null`:

```bash
while IFS= read -r -s -n 1 -t "$idle_timeout" _ 2> /dev/null; do
```

The countdown read is the only interactive read missing that treatment.

Note: on macOS `/bin/bash` 3.2, the return code of `read` is `1` for timeout, EOF, and hard I/O errors alike, so the failure mode cannot be distinguished by exit status alone — stderr suppression is the consistent fix, matching the existing codebase convention.

## Suggested fix (one line)

```bash
if IFS= read -r -s -n1 -t 1 _key 2> /dev/null; then
```

Behavior in the normal case is unchanged (key press / timeout); in the dead-terminal case the loop finishes its ticks silently instead of printing five bash errors.

## Reproduce

1. Run `mo uninstall` and remove an app.
2. When the "Press Enter ... (5)" countdown appears, press Ctrl+Z to suspend, then `exit` twice (or just close the pane).
3. Observe one `read: read error: 0: Input/output error` per remaining countdown tick.

r266-tech:
## Summary

Closes #1503.

When the hosting terminal disappears after an uninstall, macOS Bash 3.2
returns EIO from each timed countdown read and prints a misleading error for
every remaining tick. Suppress only that builtin diagnostic; timeout, keypress,
countdown, return-code, and uninstall behavior remain unchanged.

## Tests

- `bats tests/uninstall.bats tests/uninstall_tty_foreground.bats tests/uninstall_scan_bash32.bats` (123 passing)
- `bash -n bin/uninstall.sh lib/uninstall/batch.sh tests/uninstall.bats`
- `git diff --check`

The new regression drives the real interactive `main` path with a failing
timed `read`, verifies the synthetic EIO diagnostic is suppressed, and checks
that all five countdown ticks still render.

### DIFF
--- a/bin/uninstall.sh
+++ b/bin/uninstall.sh
@@ -1835,7 +1835,7 @@ main() {
         local _pressed=false
         while [[ $_countdown -gt 0 ]]; do
             printf "\r${GRAY}Press Enter to return to the app list, press q to exit (%d)${NC} " "$_countdown"
-            if IFS= read -r -s -n1 -t 1 _key; then
+            if IFS= read -r -s -n1 -t 1 _key 2> /dev/null; then
                 _pressed=true
                 break
             fi
--- a/tests/uninstall.bats
+++ b/tests/uninstall.bats
@@ -3273,6 +3273,62 @@ INNER
     [ "$status" -eq 0 ]
 }
 
+@test "completed uninstall suppresses dead-terminal countdown read errors (#1503)" {
+    local apps_cache
+    apps_cache="$(mktemp "${BATS_TEST_TMPDIR:-$BATS_RUN_TMPDIR:-$HOME}/tmp-1503-countdown.XXXXXX")"
+
+    run env HOME="$HOME" PROJECT_ROOT="$PROJECT_ROOT" APPS_CACHE_FILE="$apps_cache" \
+        /bin/bash --noprofile --norc << 'INNER'
+set -euo pipefail
+source "$PROJECT_ROOT/lib/core/common.sh"
+
+log_operation_session_start() { :; }
+hide_cursor() { :; }
+show_cursor() { :; }
+clear_screen() { :; }
+start_uninstall_interactive_screen() { :; }
+stop_uninstall_interactive_screen() { :; }
+scan_applications() { printf '%s\n' "$APPS_CACHE_FILE"; }
+load_applications() { :; }
+select_apps_for_uninstall() {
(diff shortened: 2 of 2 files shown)


youxi798:
@r266-tech thank you for the fix.

Merged on `main` in #1506. The uninstall countdown no longer spams EIO diagnostics when the hosting terminal disappears.

Closes #1503. Pick it up with `mo update --nightly`.

PR #1515, opened 2026-09-03
fix(clean): apply per-entry removal budget to XCTestDevices

HAOGRE:
## Problem

`clean_xcode_xctest_devices` removed the whole `~/Library/Developer/XCTestDevices` tree as **one** deletion item. Xcode creates one UUID-clone directory there per test run and never reaps them, so the root grows without bound. When the tree got large enough, the single removal exceeded the 30s per-item removal budget (`MOLE_TIMEOUT_DISK_VERIFY_SEC`), the `rm` was killed mid-tree, and the run reported one opaque failed item — leaving dozens of clones behind on disk.

Observed on a real machine (macOS 26, Apple Silicon):

```
$ du -sh ~/Library/Developer/*
 17M    XCPGDevices
2.1G    DVTDownloads
3.1G    Xcode
 31G    CoreSimulator
 84G    XCTestDevices
```

And the operation log after each `mo clean`:

```
[clean] FAILED /Users/haogre/Library/Developer/XCTestDevices (removal timed out)
```

84 GB of test clones with one 30-second budget meant roughly 6 clones per run got removed (the tail note said `1 item(s) exceeded the 30s removal budget`, without naming the item). Raising the budget globally is not a fix: the budget is shared with size checks and privileged probes, and the tree only grows, so any fixed value eventually loses.

## Change

1. **Per-entry deletion** (`lib/clean/dev.sh`): `clean_xcode_xctest_devices` now enumerates the first-level UUID entries and passes them to the existing guarded batch, matching the pattern `clean_xcode_system_coresimulator_caches` already uses. Sizing, the sink re-probe, and the removal budget now apply per clone (2–3 GB each, comfortably inside the budget). The root itself is never a deletion target — Xcode recreates clones inside it.

2. **Name the timed-out paths** (`lib/core/file_ops.sh`, `bin/clean.sh`): the closing `removal budget` note now lists the paths that timed out (capped at 3 + a remainder count), so the user learns which target was partly removed without opening the operation log:

```
◎ 1 item(s) exceeded the 30s removal budget and may be only partly removed: ~/Library/Developer/XCTestDevices/0635B5E2-…. Run clean again, or raise MOLE_TIMEOUT_DISK_VERIFY_SEC for slower disks.
```

No new flag or config key — the fix is structural, not a knob.

## Verification

- New Bats regressions, observed red against pre-fix code and green after:
  - `clean_xcode_xctest_devices targets each clone entry, not the root`
  - `clean_xcode_xctest_devices removal budget applies per clone entry`
  - `run with removal timeouts names the timed-out paths (#1384)`
- Existing XCTestDevices tests updated in place (guard/skip/defer/whitelist/dry-run behavior unchanged)
- `MOLE_TEST_NO_AUTH=1 bats tests/clean_dev_caches.bats tests/dev_extended.bats` — all green
- `MOLE_TEST_NO_AUTH=1 bats tests/file_ops_mole_delete.bats tests/file_ops_size.bats tests/file_ops_safe_remove_symlink.bats tests/user_file_ops.bats tests/core_safe_functions.bats` — all green
- `MOLE_TEST_NO_AUTH=1 bats tests/clean_core.bats tests/clean_apps.bats tests/cli.bats tests/clean_summary_cancel.bats` — all green
- `./scripts/check.sh --format` and `./scripts/check.sh` (incl. function-duplication audit) — pass
- Real-machine verification on macOS 26 (installed locally): after the fix, one `mo clean` run removed the remaining backlog in one pass:

```
✓ Xcode XCTestDevices test data · 9 items, 23.85GB
```

(down from 84 GB observed initially; earlier runs against the old code had only managed ~6 clones per 30s window)

### DIFF
--- a/bin/clean.sh
+++ b/bin/clean.sh
@@ -1442,6 +1442,8 @@ start_cleanup() {
     export MOLE_CLEAN_SIZING_TIMEOUTS
     MOLE_CLEAN_REMOVAL_TIMEOUTS=0
     export MOLE_CLEAN_REMOVAL_TIMEOUTS
+    MOLE_CLEAN_REMOVAL_TIMEOUT_PATHS=""
+    export MOLE_CLEAN_REMOVAL_TIMEOUT_PATHS
     log_operation_session_start "clean"
     DRY_RUN_SEEN_IDENTITIES=()
     DRY_RUN_TOTAL_PARTIAL=false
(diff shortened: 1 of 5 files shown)


youxi798:
@HAOGRE thank you for the report and the fix.

Merged on `main` via #1516. XCTestDevices now deletes per clone entry, and removal-timeout summaries name the paths that ran out of budget.

You can pick it up with `mo update --nightly`. Reopen if anything still misbehaves on your machine.

PR #1521, opened 2026-09-04
fix(cli): give mo analyze and mo status real help output

madsondeluna:
## Problem

`mo analyze` and `mo status` are the only two subcommands whose help comes from the Go `flag` package default. Every bash subcommand prints hand-written help; these two print the parser's fallback.

Before this change, on `main`:

```
$ mo analyze --help
Usage of /usr/local/bin/analyze-go:      <- stderr, names a binary the user never typed
  -json
    	output analysis as JSON instead of TUI
exit 0

$ mo analyze --bogus
flag provided but not defined: -bogus    <- stderr
Usage of /usr/local/bin/analyze-go:
  -json
    	output analysis as JSON instead of TUI
exit 2
```

Three problems in that output:

1. `Usage of /usr/local/bin/analyze-go:` names the bundled binary. The user typed `mo analyze`; `analyze-go` is an implementation detail of `bin/analyze.sh`.
2. Help goes to **stderr**. `mo purge --help`, `mo history --help` and the rest write to stdout, so `mo analyze --help | less` shows nothing.
3. A bad flag exits **2**. Every bash subcommand exits **1** and points at `--help`.

`mo status` is worse on point 1, because it has six flags and no other place documents them together.

## Change

Both `main.go` files now own their usage text next to the flag declarations, and route output by stream:

```
$ mo analyze --help
Usage: mo analyze [OPTIONS] [PATH]

Explore disk usage. Without PATH, scans a machine-wide overview.

Options:
  --json          Output the analysis as JSON instead of the interactive TUI
  -h, --help      Show this help message

Examples:
  mo analyze                 Machine-wide overview
  mo analyze ~/Library       Scan one directory
  mo analyze --json /Volumes Machine-readable output
exit 0, stdout

$ mo analyze --bogus
flag provided but not defined: -bogus
Use 'mo analyze --help' for usage information
exit 1, stderr
```

Keeping the text in `main.go` rather than in `bin/analyze.sh` is deliberate: `mo status` has six flags, and a copy in the wrapper goes stale the first time one is added. A test walks `flag.CommandLine` and fails when a declared flag is missing from the usage text.

## Behaviour change worth calling out

`mo status` argument errors move from exit 2 to exit 1. That covers unknown flags, `--proc-cpu-threshold` below zero, `--proc-cpu-window` at or below zero, and an unparseable `--watch --interval`. The reason is consistency with the ten bash subcommands, which all exit 1 on a bad option. Nothing in `tests/` asserted the old code.

`mo analyze --json` and `mo status --json` / `--watch` output are untouched.

## Scope

`cmd/analyze/main.go` is described in AGENTS.md as bootstrap only, flag parsing, `main()`, helpers, which is exactly what this touches. No TUI, scanner, cache or metrics code is involved.

## Tests

New `cmd/analyze/usage_test.go` and `cmd/status/usage_test.go` cover, for both binaries: help on stdout with exit 0 for `-h`, `--help` and `-help`; no `Usage of ` or bundled binary name in the output; an unknown flag on stderr with exit 1 naming `--help`; and a valid invocation printing nothing. `cmd/status` additionally pins that every declared flag is documented, and that the three validation failures exit 1.

```
$ go test ./cmd/...
ok  	github.com/tw93/mole/cmd/analyze	5.093s
ok  	github.com/tw93/mole/cmd/status	2.340s

$ bats tests/cli.bats
41 tests, 0 failures

$ ./scripts/check.sh
=== Checks Completed ===
```

### DIFF
--- a/cmd/analyze/main.go
+++ b/cmd/analyze/main.go
@@ -53,11 +53,11 @@ func parseArgs(args []string, stdout, stderr io.Writer) (int, bool) {
 	err := flag.CommandLine.Parse(args)
 	switch {
 	case errors.Is(err, flag.ErrHelp):
-		fmt.Fprint(stdout, usageText)
+		_, _ = fmt.Fprint(stdout, usageText)
 		return 0, false
 	case err != nil:
-		fmt.Fprintln(stderr, err)
-		fmt.Fprintln(stderr, "Use 'mo analyze --help' for usage information")
+		_, _ = fmt.Fprintln(stderr, err)
+		_, _ = fmt.Fprintln(stderr, "Use 'mo analyze --help' for usage information")
 		return 1, false
 	}
 	return 0, true
--- a/cmd/analyze/main.go
+++ b/cmd/analyze/main.go
@@ -4,8 +4,10 @@ package main
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


madsondeluna:
One thing a reviewer would reasonably want checked, so I checked it: nothing probes these two binaries' help or exit codes.

`install.sh:1506` runs `run_install_probe_with_timeout 5 "$INSTALL_DIR/mole" --help > /dev/null 2>&1`, which invokes the bash router, not `analyze-go` or `status-go`, and reads only the exit status. `mo --help` still exits 0, and `show_help` in `lib/manage/update.sh` is untouched, so the verification probe behaves identically. Neither `install.sh` nor `lib/manage/update.sh` mentions `analyze-go` or `status-go` at all:

```
$ grep -n -- '--help\|analyze-go\|status-go' install.sh lib/manage/update.sh
install.sh:1506:        if run_install_probe_with_timeout 5 "$INSTALL_DIR/mole" --help > /dev/null 2>&1; then
lib/manage/update.sh:972:        [[ "$name" == "help" ]] && display="mo --help"
```

So the exit-2-to-1 change on bad flags cannot reach the fail-closed install verification path.

PR #1523, opened 2026-09-04
feat(cli): suggest the closest command on a typo, and send the error to stderr

madsondeluna:
## Problem

A mistyped command gives the user nothing to act on, even though the list needed to do better is already in the repo:

```
$ mo clena
Unknown command: clena
Use 'mole --help' for usage information.
exit 1
```

`MOLE_COMMANDS` in `lib/core/commands.sh` already exists as the shared command list behind help text and shell completion. It just was not consulted here.

The three lines also go to **stdout**, so `mo clena > out.txt` writes the diagnostic into the file and leaves the terminal blank.

## Change

```
$ mo clena
Unknown command: clena
Did you mean 'mo clean'?
Use 'mole --help' for usage information.
exit 1, stderr
```

`mole_suggest_command` in `lib/core/commands.sh` reads the same `MOLE_COMMANDS` array, so a command added there is matched with no second edit. It tries an unambiguous prefix first (`mo unins` reaches `uninstall`, which edit distance alone would not), then falls back to Levenshtein distance.

The distance function keeps two indexed rows rather than a matrix because bash 3.2 has no associative arrays. Inputs are one command word against fourteen names, so the cost is not observable.

## The threshold, and why it is not a single number

An input of three characters or fewer tolerates one edit; anything longer tolerates two. At distance 2, `mo rm` is as close to `remove` as it is to nothing in particular, and on a tool that deletes files a confident wrong suggestion is worse than silence. `mo rm`, `mo up` and `mo xyzzy` therefore print no suggestion at all, which the tests pin.

The suggestion is never executed, only printed, so this does not introduce prefix abbreviation as a calling convention.

## Streams

`Unknown command`, the suggestion, and the `--help` pointer move from stdout to stderr. `run` in Bats merges both, so the three existing assertions in `tests/cli.bats` (`unknown-command`, `check`, `doctor`) are unaffected; a test in the new file pins the split explicitly.

## Scope

`mole` stays a router: it gains three lines in the existing `*)` branch and the helper lives in `lib/core/commands.sh` next to the list it reads. No new flag, environment variable or config key.

## Tests

New `tests/command_suggestion.bats`, 10 cases: single- and double-character typos across five commands, unambiguous prefixes, case folding, silence when nothing is close, silence on a two-character input, a check that the suggestion is always a real member of `MOLE_COMMANDS`, plus end-to-end runs of `mole clena` and `mole xyzzy` and the stdout/stderr split.

```
$ bats tests/command_suggestion.bats
10 tests, 0 failures

$ ./scripts/check.sh
✓ shfmt format clean
✓ ShellCheck passed
=== Checks Completed ===
```

## Product Decision Filter (AGENTS.md)

1. **Does it belong to an existing command?** It is command routing itself, the `*)` branch of `mole`'s dispatch, reading the `MOLE_COMMANDS` list that already backs help text and completion.
2. **Safe by default, testable without real auth, explainable in one screen?** It prints one extra line and executes nothing. The Bats file needs no auth and no fixtures.
3. **Can the user verify what will change?** The suggestion is text; the user still has to type the real command.
4. **Rebuildable or evidence-backed target?** No files are touched.
5. **Better as documentation or a warning?** The help text already lists every command, and the user who typed `clena` did not read it. Naming the near match at the point of failure is the shorter path.

No new flag, environment variable or config key.

## Taking half of this

The stdout-to-stderr move stands on its own and is the plainer of the two changes. If the suggestion itself is unwanted, dropping `mole_suggest_command`, the two lines in `mole` that call it, and the seven unit tests leaves a working stream fix behind.

## bash 3.2

`mole_suggest_command` builds its distance matrix from two indexed arrays because macOS bash has no associative arrays. The tests invoke `/bin/bash --noprofile --norc` explicitly rather than PATH's `bash`, following `tests/no_color.bats`, so the 3.2 path is what actually runs:

```
$ /bin/bash --version | head -1
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)
```


tw93:
@madsondeluna Thanks for catching this. The stream bug is real and I have kept your fix: the diagnostic now goes to stderr, with a regression test in `tests/cli.bats`.

I dropped the closest-command suggestion. The router already handles the one confusion that is predictable, `optimize|optimise` and `analyze|analyse`, by exact enumeration, and `mo completion` gives tab completion that removes the guess entirely. On a tool that deletes files a wrong suggestion is worse than none, and the edit-distance thresholds would need rethinking every time a command is added. Nobody has asked for it in the issue history either, so it is surface Mole does not need.

Your report also led somewhere useful: seven more sites print the same kind of diagnostic to stdout, in `mole`, `bin/optimize.sh`, `bin/purge.sh`, `bin/installer.sh` and `bin/uninstall.sh`. I am fixing those in a follow-up so the whole surface matches `bin/history.sh`.

Merging now. It will be in the next release, or immediately via `mo update --nightly`.

PR #1530, opened 2026-09-05
fix(uninstall): refresh names after language changes

r266-tech:
## Summary

- fingerprint the ordered `AppleLanguages` snapshot alongside each cached display name
- re-resolve only the localized name when that fingerprint changes, while retaining cached size and last-used metadata
- carry the current fingerprint through both the immediate cache snapshot and deferred metadata refresh; legacy rows without one heal on their next scan

## Why

The localized resolver added for #1520 depends on the user's language order, but the persistent v3 cache currently accepts an unchanged app as warm without checking which language produced its name. Changing macOS language preferences can therefore keep showing the old-language name indefinitely, including after the seven-day metadata refresh.

This stays outside app matching and deletion: it changes only whether a cached display name is reusable.

## Verification

- New Bash 3.2 integration regression failed on unmodified `main` by returning the cached Chinese name, then passed with the patch and retained the cached size/last-used fields.
- `MOLE_TEST_NO_AUTH=1 bats tests/uninstall.bats tests/uninstall_scan_bash32.bats` — 124/124 passed
- `./scripts/check.sh` — passed (format, `go vet`, ShellCheck, syntax, duplication and diagnostics checks)
- `go test ./...` — passed
- Full `TERM=xterm-256color MOLE_TEST_NO_AUTH=1 ./scripts/test.sh`: 1,834 passed, 6 skipped, with one unrelated pre-existing `clean_dev_caches.bats` failure reproduced unchanged on the exact base commit (`codex resolution treats a failed mdfind as unanswered`)

### DIFF
--- a/bin/uninstall.sh
+++ b/bin/uninstall.sh
@@ -1064,7 +1064,15 @@ _scan_finalize_index() {
                 display_name = $2
                 bundle_id = $3
                 app_mtime = $4
-                if (NF >= 11) {
+                # A merged row is the 5-field scan row plus the 7 cache fields
+                # appended above, so 12 is the only width the current writers
+                # produce. The threshold has to track that sum: while the cache
+                # block was 6 fields wide, 11 named the same shape, and leaving
+                # it at 11 after the language signature landed would have let a
+                # hypothetical 11-field row take this branch and read every
+                # cached_* value shifted by one, so the signature itself would
+                # render as the display name.
+                if (NF >= 12) {
                     inline_size_kb = $5
                     cached_mtime = $6
                     cached_size_kb = $7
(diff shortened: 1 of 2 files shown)


tw93:
@r266-tech Thank you, and the diagnosis is the part I appreciate most: the cache called a name warm without ever asking which language produced it, and the deferred refresh passed the old name straight through, so nothing in the seven-day path would have healed it either. I ran your regression against main before merging and got the stale Chinese name back, then watched the same file pass on your branch.

I pushed one commit to your branch before merging. With the signature in it a merged row is 12 fields wide, but the branch that reads the cached block still asked for 11, which was that shape's width while the cache block was 6 fields. It picks correctly either way today, since both writers of the scan file emit exactly 5 fields, so the change is only to name the real sum and leave the arithmetic in a comment for whoever adds the next field.

Keeping the v3 schema instead of bumping it was the right call for this: a version bump would have discarded everyone's cached size and last-used rows over a problem that only affected the name. It ships in the next CLI release.

PR #1544, opened 2026-09-08
fix(clean): reach WeChat and WeCom sandboxed container caches

dare2live:
## Summary

WeChat (`com.tencent.xinWeChat`) and WeCom (`com.tencent.WeWorkMac`) ship sandboxed, so the existing `~/Library/Caches/<bundle id>` entries in `clean_communication_apps` match nothing on a current install. On macOS 26.5.2 neither `~/Library/Caches/com.tencent.xinWeChat` nor `~/Library/Caches/com.tencent.WeWorkMac` exists, while about 1.4GB of rebuildable cache sits in the app containers where cleanup never looks.

This is the same shape as JianyingPro (#1277): the app keeps heavy caches outside `~/Library/Caches`, so standard cleanup never sees them.

`clean_wechat_container_caches` adds the container paths, measured on macOS 26.5.2 with WeChat 4.x and WeCom 4.x:

| Target | Reclaimed |
|---|---|
| WeCom `cefcache/<profile>/Service Worker` | 482MB |
| WeCom `cefcache/<profile>/Cache` | 424MB |
| WeCom `cefcache/<profile>/Code Cache` | 48MB |
| WeCom `Application Support/WXWork/Log` | 81MB |
| WeChat `Documents/app_data/log` | 187MB |
| WeChat `.wxapplet/WMPF` | 161MB |
| **Total** | **~1.4GB** |

Explicit non-targets, excluded as user data or account state:

- WeChat `Documents/xwechat_files/**` — chat databases and received files
- WeChat `Documents/app_data/radium/users` — per-account webview state
- WeCom `Documents/Profiles/*/Messages1` — chat messages
- WeCom `Documents/cefcache/<profile>/` root — cookies, Account Web Data, login state
- WeCom `Documents/Profiles/*/Publishsys` — distribution payloads, reclaimable value unmeasured, so out of scope

## Safety Review

Yes, this changes cleanup behavior. Three boundaries are worth review:

**Owner probe.** WeCom's main executable is a localized string, so an `-x` match on a romanized name would never fire and the guard would silently pass while the app was running. The probe matches the ASCII helper bundles (`WeComAgent`, `wecom-agent`) and executable paths instead. Verified on a machine with both apps installed: with WeChat and WeCom quit, every probe reports no match, so none of them is a resident menu-bar agent that would permanently skip cleanup the way a tray helper would. The guard is rebound through `safe_clean_guarded` at every sink, and `mole_pgrep_any` returning 2 (unknown) fails closed.

**Profile selection.** `cefcache` is a Chromium user-data-dir, so most of its children are component-updater payloads (`AutofillStates`, `CertificateRevocation`, `CookieReadinessList`, ...) holding version directories rather than per-account profiles — 34 directories on the test machine, of which only the real profiles own Chromium cache directories. A profile is therefore selected structurally, by the cache directories it owns, not by directory name.

**Symlink handling.** Each profile root is resolved with `cd -P` and skipped when the physical path differs, so the sink stays inside the container that was inspected.

No change to path validation, protected directories, sudo boundaries, or release integrity.

## Tests

Automated:

```
MOLE_TEST_NO_AUTH=1 bats tests/clean_app_caches.bats     96 passed, 0 failed
./scripts/check.sh                                       shfmt, ShellCheck, syntax,
                                                         function duplication all pass
```

Five new cases in `tests/clean_app_caches.bats`, each on an isolated fixture root:

1. cleans every container target and never emits a chat-data path
2. skips every target while an owner process is running
3. fails closed when process state is unknown
4. refuses a symlinked WeCom profile
5. ignores WeCom component payloads under `cefcache`

Red/green verified: cases 1-4 fail against the pre-change source (the function does not exist), and case 5 fails against a first draft that globbed `cefcache/*/` without the structural predicate.

Manual, on macOS 26.5.2 with WeChat running and WeCom quit:

- `./mole clean --dry-run` completes and lists no WeChat/WeCom target, because the owner guard defers the family while WeChat is running.
- With the probe stubbed to "not running", the function resolves exactly the intended targets and no chat-data path.

`go test ./...` was not run; this change touches no Go. Two unrelated failures appear in a full `./scripts/test.sh` sweep on this machine: `validate_path_for_deletion allows protected input-method names in Trash (#1517)` also fails on a pristine checkout, and `clean_browsers never enters Firefox cleanup while Firefox is running` passes in isolation and across its whole file (73 passed, 0 failed) with this change applied.

## Safety-related changes

- New cleanup targets under `~/Library/Containers/com.tencent.xinWeChat` and `~/Library/Containers/com.tencent.WeWorkMac`, gated by a new `wechat_or_wecom_running` owner probe that fails closed on unknown process state.


tw93:
@dare2live thank you for the careful report and the measured cache boundaries. I tightened the final path and process checks, narrowed Service Worker cleanup to CacheStorage, and merged it for the next release.

PR #1545, opened 2026-09-08
Fix purge cancellation and speed up project navigation

M-Hassan-Raza:
Purge could continue after interrupted work, accept closed input as confirmation, and lose the focused artifact when the terminal resized. This fixes those paths and reports incomplete scans and unknown sizes explicitly.

Adds paging, project jumps and Unicode search while preserving selections. Preparation avoids repeated work, scan slots refill as workers finish, and menu redraws reuse formatted text until resize. Final deletion checks and scan scope stay intact.

A synthetic 600-artifact preparation run went from 49.2s to 27.4s with identical candidates. Navigation with 3,000 artifacts went from about 90ms to 21ms per key.

Verified with 308 focused acceptance tests, Go tests and repository checks. The full shell suite still has ten upstream-baseline failures; two transient Mail failures passed reruns on both versions.


tw93:
@M-Hassan-Raza thank you for the thorough purge improvements. The safer cancellation and incomplete-scan handling, along with the faster project navigation, are now merged and will ship in the next release.

PR #1562, opened 2026-09-14
fix(uninstall): handle third-party cask scan failures

### LINKED ISSUE #1558: [BUG] Mo 卸载 rebased 扫描文件过程中意外退出
## Describe the bug

卸载通过 brew 安装的 [Rebased](https://github.com/detachhead/rebased) 时扫描文件过程中会直接中断退出， 增加了 --debug 也看不到有效的日志。

## Steps to reproduce

1. mo --debug uninstall
2. select Rebased
3. mo 扫描过程中中断

## Debug logs

> ----------------------------------------------------------------------
> Mole Debug Session, 2026-09-13 23:24:26
> ----------------------------------------------------------------------
> User: -
> Hostname: Mac-mini.local
> Architecture: arm64
> Kernel: 25.5.0
> macOS: 26.5.1, 25F80
> Shell: /bin/zsh, xterm-256color
> Sudo Access: Required
> ----------------------------------------------------------------------

## Environment

```
 mo --version

Mole version 1.54.0
macOS: 26.5.1
Architecture: arm64
Kernel: 25.5.0
SIP: Enabled
Disk Free: 74.73GB
Install: Homebrew
Shell: /bin/zsh
```

## Additional context

```
 mo --debug uninstall
[DEBUG] Debug logging enabled. Session log: /Users/name/Library/Logs/mole/mole_debug_session.log
◎ Selected 1 apps:
1. Rebased           1.13GB  |  Last: 3m ago
⠙ Scanning files...%
```

lazzyms:
Fixes #1558

## What changed

- Keep exact Caskroom detection usable for third-party tap casks when `brew info --cask <token>` cannot resolve the short token.
- Preserve timeout and signal failures, and keep final Homebrew uninstall verification unchanged.
- Surface ordinary uninstall scan failures instead of returning silently before the preview.
- Add regressions for third-party cask detection and visible scan failures under Bash 3.2-compatible execution.

## Why

Rebased is installed from a third-party Homebrew tap. Its unique app bundle path and installed cask token are already strong ownership evidence, but the optional `brew info` lookup can fail for a short token and currently aborts the batch scan. The interactive caller runs with `set -e`, so that refusal drops the user back to the shell with no explanation.

The fallback remains fail-closed: the exact Caskroom evidence is required, timeouts/signals still abort, and `brew_uninstall_cask` still verifies both the cask record and app path after removal.

## Validation

- `TERM=xterm-256color MOLE_TEST_NO_AUTH=1 ./scripts/test.sh`
- Direct Bash 3.2-compatible execution of both new regression scenarios (Bats is not installed in the local environment)


tw93:
@lazzyms Thanks for digging into this. Both halves are already fixed on main: the batch scan now names the phase that failed and gives a next command instead of returning silently, and a failed `brew info` lookup no longer aborts the scan when the cask is clearly installed.

The reason I am closing this rather than merging it is one difference in the fallback. On main the installed token is accepted only when the Caskroom entry is a symlink that resolves to the exact app the user selected, rechecked after the probe. A unique token plus a matching bundle path is not enough by itself, because a copied bundle or a shared basename satisfies it, and the uninstall would then remove a cask the user did not choose. Timeout and signal failures still abort in both versions.

Closing as already resolved. Thanks for the fix and for writing the regression tests.

PR #1574, opened 2026-09-16
fix(uninstall): keep selector rows on one line (#1573)

### LINKED ISSUE #1573: [BUG] mo uninstall leaves orphan 'ay' character below list in standard 80-column terminals
## Describe the bug                                                                                                                                                                                                 

   In standard 80-column terminals (such as default macOS Terminal or Ghostty), navigating through the application list in `mo uninstall` causes text lines to exceed the terminal width and wrap onto the next line.

   This breaks the single-line absolute cursor positioning (`\033[row;1H`), causing trailing `"ay"` characters to leak into the padding row and line prefixes during scrolling (e.g. `ay○ 豆包`, `ay○ Orca`, and an
 orphan `ay` on the bottom blank line).

## Steps to reproduce

   1. In an 80×24 terminal (or any narrower terminal), run `mo uninstall`.
   2. Ensure at least one installed app was used yesterday (whose last used date is formatted as `Yesterday`, such as WPS Office or 豆包).
   3. Scroll through the list using Up/Down arrow keys.
   4. An orphan `ay` stably appears on the blank line below the list, and `ay` leaks in front of subsequent items (e.g. `ay○ 豆包`, `ay○ Orca`).

## Expected behavior

   Rendered item lines should never exceed `terminal_width`. Each application should strictly occupy a single physical line, without any wrapped characters or persistent visual artifacts.

## Root Cause (Minimal Surface)

   In `lib/ui/app_selector.sh`:
   ```bash
   # Fixed elements: "  ○ " (4) + " " (1) + size (9) + " | " (3) + max_last (7) = 24
   local fixed_width=24
   local available=$((terminal_width - fixed_width))
 ```

 1. max_last is underestimated: It assumes date strings are at most 7 characters (e.g. 12d ago, Unknown). However, format_last_used_summary in lib/core/ui.sh returns "Yesterday" (9 chars) and "This year" (9 chars),
    which is 2 characters wider than budgeted.
 2. CJK Ambiguous checkbox width: In UTF-8/CJK environments, the empty checkbox ○ (U+25CB) or arrow ➤ occupies 2 columns instead of 1, adding another column to the prefix.
 3. Dynamic calculation mismatch: Because available = terminal_width - 24 dynamically adjusts name padding based on terminal width, any line with Yesterday will always exceed the physical terminal boundary by
    exactly 2 columns, regardless of whether the window is 80 columns or narrowed further.

 When available_width takes the full budget:
 4~5 (prefix) + available (name) + 1 + 9 (size) + 3 + 9 (Yesterday) = terminal_width + 2
 The trailing 2 characters ("ay") wrap onto the next line, polluting the subsequent line's prefix and the bottom padding row.

 Screenshot

 Image

 Environment

 - Mole version: 1.54.0
 - macOS: 27.0 (Build 26A428)
 - Architecture: arm64
 - Kernel: 27.0.0
 - SIP: Enabled
 - Shell: /bin/zsh
 - Terminal: 80x24 (Default Terminal.app / Ghostty)

 Suggested Minimal Fix

 Adjust fixed_width in lib/ui/app_selector.sh from 24 to 27 to safely accommodate 9-character date strings (Yesterday / This year) and CJK checkboxes, keeping the diff minimal and ensuring lines never wrap across
 terminal boundaries.

md786-dotcom:
Fixes #1573

## Summary

- Keep each uninstall selector row on one terminal line.
- Increase the reserved width from 24 columns to 27 columns.
- On a narrow terminal, clamp the fallback name width to the remaining space.

## Safety Review

- This change affects uninstall list display only.
- This change does not affect path validation, protected directories, symlink handling, sudo, or release integrity.

## Tests

- `MOLE_TEST_NO_AUTH=1 bats tests/uninstall.bats -f '1573'`
  - Before the fix: failed (`overflow cols=40 total=45`)
  - After the fix: passed
- `MOLE_TEST_NO_AUTH=1 bats tests/menu_trap_restore.bats`

## Safety-related changes

- None.

## Not in this change

Steam games show `N/A (Steam-managed)` in the size column.
That text has 19 characters.
The size column has 9 characters.
Those rows stay too wide by approximately 10 columns.
The cause is the same. The trigger is different.
This change does not correct that problem.

NOTE: Validation / Unit Integration Tests also failed on 3d8ac234852bdc28538e11cb5734a5623f1c5109. So, seems like an unrelated issue.


tw93:
@md786-dotcom sorry we did not merge this. The wrap was already covered by the issue author's #1575, so we used that branch and took your Steam label and narrow-terminal clamp there instead. No further action needed on this PR.

Issue #1579, opened 2026-09-17
[BUG] mo uninstall still aborts when an installed cask that ships only a binary cannot be read by brew info


huangy7:
## Describe the bug

`mo uninstall` aborts the whole batch and removes nothing when an installed third-party cask cannot be parsed by `brew info --cask`.

#1558 fixed this symptom, but only for casks whose app is symlinked into Caskroom — that fallback lives in `_detect_cask_via_caskroom_search`. A cask that ships **only a `binary`** puts nothing there for the search to find, so detection falls through to `_detect_cask_via_brew_list`, where `lib/uninstall/brew.sh:253` still reads:

```bash
[[ $info_rc -eq 0 ]] || return 2
```

with no ownership fallback. `_batch_scan_app_details` then treats rc 2 as fatal, and `batch_uninstall_applications` aborts the entire scan, so **every** app selected in that run is left untouched.

## Steps to reproduce

1. Install a cask that ships a CLI binary only, from a GoReleaser-generated tap:

   ```bash
   brew tap esengine/reasonix
   brew install --cask reasonix
   ```

   Its cask uses `#{staged_path}` inside a `postflight_steps` block, which Homebrew 6+ no longer exposes on `Homebrew::InstallSteps::DSL`, so the cask is unreadable:

   ```console
   $ brew info --cask reasonix
   Error: Cask 'reasonix' is unreadable: undefined local variable or method 'staged_path' for an instance of Homebrew::InstallSteps::DSL
   ```

2. Note the shape that matters: the cask declares only `binary "reasonix"`. There is no `app` stanza and no `Reasonix.app` anywhere in Caskroom. A `/Applications/Reasonix.app` exists independently (bundle id `com.wails.reasonix-desktop`), so `brew list --cask` name-matches on `reasonix` while `brew info --cask reasonix` cannot be read.

3. `mo uninstall Reasonix` (or select it in `mo uninstall`)

4. The file scan runs for ~40s, then the batch aborts. Nothing is removed.

## Evidence

The ownership probe returns "could not determine" rather than "not Homebrew":

```console
$ get_brew_cask_name "/Applications/Reasonix.app"
rc=2      # 0 = brew-managed, 1 = not brew, 2 = cannot determine
```

`brew list --cask` still lists it cleanly, so it looks healthy:

```console
$ brew list --cask | grep reasonix
reasonix
```

Three attempts, all with zero effect (`~/Library/Logs/mole/operations.log`):

```text
[...]

Issue #1344, opened 2026-08-03
[BUG] Crashes on User essentials


dmt0:
## Describe the bug

Silently crashes

## Steps to reproduce

1. Run command: `mo clean`
2. Gets to User Essentials
3. Crashes

## Expected behavior

Completes the scan

## Debug logs

```bash
 ✘  ~  mo clean --debug
[DEBUG] Debug logging enabled. Session log: /Users/dima/Library/Logs/mole/mole_debug_session.log

Clean Your Mac

◎ Use --dry-run to preview, --whitelist to manage protected paths
✓ Admin access already available

⚙ Intel | Free space: 47.41GB
✓ Whitelist: 21 core patterns active

➤ System
  / Cleaning system caches...[DEBUG] Finding, sudo, in /Library/Caches: *.cache, age: 7d, type: f
  | Cleaning system caches...[DEBUG] Privileged cleanup scan failed for /Library/Caches (status 1)
[DEBUG] System caches cleanup incomplete (status 1)
  | Cleaning system crash reports...[DEBUG] Finding, sudo, in /Library/Logs/DiagnosticReports: *, age: 7d, type: f
  / Cleaning system logs...[DEBUG] Finding, sudo, in /private/var/log: *.log, age: 7d, type: f
  \ Cleaning third-party system logs...[DEBUG] Directory does not exist, skipping: /Library/Logs/Adobe
  - Cleaning third-party system logs...[DEBUG] Directory does not exist, skipping: /Library/Logs/CreativeCloud
  | Cleaning third-party system logs...[DEBUG] Finding, sudo, in /Library/Logs: adobegc.log, age: 7d, type: f
[DEBUG] Keeping /Library/Updates: managed by Software Update
  - Cleaning rebuildable system service caches...Path validation: protected path skipped: /Library/Caches/com.apple.iconservices.store
  \ Cleaning rebuildable system service caches...[DEBUG] Skipped sudo remove for protected path: /Library/Caches/com.apple.iconservices.store
[DEBUG] Accessible rebuildable GPU caches cleanup incomplete (status 1)
[DEBUG] Stale wallpaper downloads cleanup incomplete (status 1)
  / Cleaning system diagnostic logs...[DEBUG] Finding, sudo, in /private/var/db/diagnostics: *, age: 7d, type: f
  / Cleaning system diagnostic logs...[DEBUG] Finding, sudo, in /private/var/db/DiagnosticPipeline: *, age: 7d, type: f
[...]

Issue #43, opened 2025-10-29
Add support for Dia, Comet, Orion, and Zen browser cache and data cleaning


mkhalila:
Hi! I have Dia, Comet, Orion, and Zen browsers installed on my system, but Mole doesn't currently detect or clean their cache and data. 

Could you please add support for these browsers by including their default cache/data paths to the cleanup script? This would help keep these browsers optimized along with the others that Mole already supports.

I am happy to help provide paths etc. if that helps

Thank you for maintaining this useful utility!

tw93:
Thank you very much for your feedback. 4 new browser supports have been added,I will release a version today.
```
- Dia (~/Library/Caches/company.thebrowser.dia/*)
- Comet (~/Library/Caches/Comet/*)
- Orion (~/Library/Caches/com.kagi.kagimacOS/*)
- Zen (~/Library/Caches/zen/*)
```

PR #1580, opened 2026-09-17
fix(uninstall): name the app and the cask check when ownership is undetermined

huangy7:
## Summary

Refs #1579.

- When a selected app's cask ownership cannot be determined, `mo uninstall` aborts the batch. The Homebrew branch of the abort hint told the user to run `brew list --cask`, but that command never parses a cask definition: a cask brew cannot read still lists as installed and healthy, so the listing diagnoses nothing. `brew info --cask <cask>` is what surfaces the failure. (`brew info --cask` with no argument does not help either — it does not load installed cask definitions.)
- With several apps selected, the message also gave no way to tell which selection failed. `_batch_scan_app_details` now carries the app name out the same way `_batch_scan_stage` is already carried, so the abort can name it.
- This follows the AGENTS.md rule that a gate which refuses must name the cause and the next action; a rule line is added to lock it.

## Safety Review

- **Uninstall behavior: message only.** rc 2 still aborts the batch and nothing is removed. This does not change what gets deleted, or the ownership decision behind it — only the text printed after the abort.
- **Path validation, protected directories, symlink handling, sudo boundaries, release/install integrity:** unchanged. No new command is executed; `brew info --cask` is printed as guidance, never run.

## Tests

Automated:

- `bats tests/uninstall.bats` (`--filter "Homebrew"`) — includes `batch uninstall reports an inconclusive Homebrew scan before any removal`, which owns this exact path (`get_brew_cask_name` stubbed to return 2) and is the test this PR updates. Verified **red** against unmodified `lib/uninstall/batch.sh` (fails on the app-name assertion at `tests/uninstall.bats:4229`) and **green** after the change.
- `bats tests/brew_uninstall.bats` — 22/22 pass.
- `shellcheck mole install.sh bin/*.sh lib/*/*.sh scripts/*.sh` — clean.
- `shfmt -i 4 -ci -sr -d lib/uninstall/batch.sh` — no diff.
- `./scripts/test.sh` — 1913 tests, 4 failures, 10 skipped. Three of the four reproduce on unmodified `main` in a clean `git worktree`, unrelated to this change:
  - `tests/clean_core.bats` #43 `_clean_mail_downloads removes old attachments`
  - `tests/clean_mail_downloads.bats` #5 `mo clean completes when the Mail Downloads du stalls (#1344)`
  - `tests/scripts.bats` `contributor requirements match the release and Go toolchains` — `CONTRIBUTING.md`'s `Go x.y+` no longer matches `go.mod`
  
  The fourth was the stale assertion this PR updates, which is now green.

Manual checks: none required — no destructive path is touched, and the stub-based test above exercises the abort end to end without deleting anything.

## Safety-related changes

- None.


tw93:
@huangy7 thank you for the fix. Merged as written. The abort now names the app and points at `brew info --cask`. Run `mo update --nightly` to pick it up.

PR #1582, opened 2026-09-17
fix: resolve real SCRIPT_DIR when mole is invoked via a symlink

idhyt:
## Summary

On macOS, bash 3.2 does not always populate BASH_SOURCE[0] with the resolved physical path, and the original construct

    SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"

resolves to the symlink's directory rather than the actual install location. Subsequent `source "$SCRIPT_DIR/lib/..."` imports then fail.

Replace it with a portable resolver that:
  - uses `readlink -f` when it is available (Linux coreutils)
  - otherwise iteratively resolves symlinks with `readlink` + `cd -P`, handling both relative and multi-level links, a common macOS setup
  - finally derives SCRIPT_DIR from the canonicalized SCRIPT_PATH


tw93 (approved):
LGTM. Real bug: symlink invoke (e.g. Homebrew bin → Cellar) left SCRIPT_DIR on the link dir, so `source "$SCRIPT_DIR/lib/..."` failed. GNU `readlink -f` + BSD walk is the right split; verified abs symlink resolves to the physical install and still sources lib. Left `MOLE_ENTRY_SCRIPT="${BASH_SOURCE[0]}"` alone, which matches the update-entrypoint contract in shell-and-test-pitfalls. Ready to squash when you want.

tw93:
@idhyt thank you for the report.

This is on main now. A cross-directory `mole` symlink sources `lib/` from the real install, and a Homebrew-style prefix still uses `lib/` next to the invoked `bin/`.

`mo update --nightly` picks it up; the next stable release will include it too.

idhyt:
Thanks, nice work~

Issue #1512, opened 2026-09-03
[BUG] VS Code obsolete extension cleanup silently skips on macOS 15 (plutil -p rejects JSON)


md786-dotcom:
## Describe the bug

`clean_editor_obsolete_extensions` in `lib/clean/app_caches.sh` reads each editor's `.obsolete` file (JSON) with:

```bash
plutil -p "$obsolete_file" 2> /dev/null | sed ...
```

On macOS 15.x, `/usr/bin/plutil -p` cannot parse JSON. It exits with `Unexpected character { at line 1`. The stderr redirect hides the error, the pipeline returns no keys, and the function silently cleans nothing.

This is fail-safe (nothing wrong is deleted), but the feature is dead on affected macOS versions.

## Steps to reproduce

1. Use macOS 15.x (reproduced on 15.6.1, build 24G90).
2. Create a fake VS Code extensions root with an `.obsolete` JSON file and one stale extension directory:

```bash
HOME=$(mktemp -d)
ext_root="$HOME/.vscode/extensions"
mkdir -p "$ext_root/pub.ext-old-1.0.0"
printf '{\n  "pub.ext-old-1.0.0": true\n}\n' > "$ext_root/.obsolete"
```

3. Confirm `plutil -p` fails on that file:

```bash
plutil -p "$ext_root/.obsolete"
# Unexpected character { at line 1
```

4. Run the Bats regression (no real VS Code install needed):

```bash
MOLE_TEST_NO_AUTH=1 bats -f "clean_editor_obsolete_extensions removes only dirs" tests/clean_app_caches.bats
```

5. The test expects `CLEAN:.../pub.ext-old-1.0.0` but gets empty output.

## Expected behavior

Mole reads the `.obsolete` JSON keys and offers only those extension directories for cleanup. The four tests under `clean_editor_obsolete_extensions` in `tests/clean_app_caches.bats` (#910, #1461) should pass on macOS 15.

## Debug logs

Not applicable — the function returns success with no output and no debug line when parsing fails.

Minimal parse check:

```text
$ plutil -p /path/to/.obsolete
Unexpected character { at line 1

$ plutil -convert xml1 -o - /path/to/.obsolete | plutil -p -
"pub.ext-old-1.0.0" => 1
```

`plutil -convert xml1` accepts the JSON; only `-p`/`-lint` on the raw file rejects it.

## Environment

- macOS 15.6.1 (24G90), arm64
[...]

Issue #1461, opened 2026-08-26
[FEATURE] Reconcile unregistered VS Code extension directories beyond `.obsolete`


TyceHerrman:
## Feature description

Add a conservative fallback for VS Code-compatible extension directories that are no longer registered to any editor profile but are absent from the editor's `.obsolete` marker.

This is a follow-up to #910. That change safely removes direct-child directories explicitly named by `.obsolete`, but the marker is a transient removal journal rather than a complete inventory of extensions on disk. If it is missing, empty, truncated, or incomplete, `mo clean` has no candidates even when old extension versions are no longer registered.

## Use case

On VS Code Insiders 1.135.0-insider, `~/.vscode-insiders/extensions/.obsolete` was a zero-byte file while `extensions.json` registered only the current versions:

```text
openai.chatgpt-26.5818.61809-darwin-arm64
anthropic.claude-code-2.1.245-darwin-arm64
ms-mssql.mssql-1.45.0
```

Older valid extension directories remained beside them, including:

```text
openai.chatgpt-26.5818.31338-darwin-arm64
openai.chatgpt-26.5818.41705-darwin-arm64
openai.chatgpt-26.5818.21641-darwin-arm64
openai.chatgpt-26.5814.41407-darwin-arm64
anthropic.claude-code-2.1.241-darwin-arm64
anthropic.claude-code-2.1.239-darwin-arm64
anthropic.claude-code-2.1.238-darwin-arm64
anthropic.claude-code-2.1.237-darwin-arm64
anthropic.claude-code-2.1.235-darwin-arm64
```

These nine unregistered OpenAI and Anthropic directories used about 3.5 GiB. Because the current Mole implementation only enumerates keys from `.obsolete`, none are offered for cleanup in this state.

## Proposed solution

Keep `.obsolete` as the authoritative fast path, then optionally reconcile the extension root against the editor's profile registries:

1. Skip reconciliation while the corresponding editor is running or if its process state cannot be determined safely.
2. Discover and parse the default extension registry plus every profile-specific extension registry for that editor.
3. Build an exact keep-set from the registered extension locations across all profiles.
4. Inspect only non-hidden, non-symlink, immediate child directories of the extension root.
5. Treat a directory as a candidate only when it contains valid extension metadata, its on-disk identity matches the directory being inspected, and its exact path is absent from every profile keep-set.
6. Ignore staging or temporary directories such as hidden UUID directories and `.vsctmp` entries.
[...]

Issue #968, opened 2026-05-24
[BUG] Chrome MV3 extension popups break after mo clean


Lemonawa:
## Describe the bug

Running `mo clean` breaks Chrome MV3 extension popups after Chrome is reopened. The clean output shows Chrome Service Worker ScriptCache being deleted, and after that extension UIs stop opening until each extension is toggled off/on in `chrome://extensions`.

This matches issue #785: the root cause is cleaning Chrome's Service Worker ScriptCache while Chrome is not running.

## Steps to reproduce

1. Quit Chrome completely
2. Run command: `mo clean`
3. Reopen Chrome
4. Click any MV3 extension icon
5. See error: popup/UI does not open until the extension is disabled and re-enabled in `chrome://extensions`

## Expected behavior

Chrome extensions should keep working after `mo clean`, or `mo clean` should avoid deleting the Chrome Service Worker ScriptCache when that would break MV3 extension service workers.

## Debug logs

Please run the command with `--debug` flag and paste the output here:

```bash
mo clean --debug
# Example: mo clean --debug
```

<details>
<summary>Debug output</summary>

```text
From `mo clean --debug --dry-run` on my machine:

  → Chrome Service Worker ScriptCache 25 items, 18.6MB dry
[DEBUG] PERF [Chrome Service Worker ScriptCache: path scan] 0.269s
[DEBUG] Cleaning: Chrome Service Worker ScriptCache, 25 items
[DEBUG] === Chrome Service Worker ScriptCache ===
[DEBUG] PERF [Chrome Service Worker ScriptCache: size calc] 0.014s
[DEBUG] PERF [Chrome Service Worker ScriptCache: deletion] 0.005s
[DEBUG] TIMEOUT on CacheStorage scan for:
/Users/lemonawa/Library/Application Support/Google/Chrome/Default//Service Worker/CacheStorage
```

</details>

## Environment

Please run `mo update` to ensure you are on the latest version, then paste the output of `mo --version` below:

```text
Mole version 1.39.1
macOS: 26.5
Architecture: arm64
Kernel: 25.5.0
SIP: Enabled
Disk Free: 93Gi
Install: Manual
Shell: /bin/zsh
```

## Additional context

I checked issue #785. It reports the same Chrome MV3 extension breakage and says it was fixed in a later version / nightly. This report is the same class of bug from the user perspective.

tw93:
[...]

Issue #964, opened 2026-05-24
[BUG] mo clean --dry-run aborts with "Malicious Script Blocked" notification.


TsubasaNoNai:
## Describe the bug

Run mo clean --dry-run
"Malicious Script Blocked. A script was blocked because it contains malware. This script did not harm your Mac." Error pops up. 

## Steps to reproduce

1. mo clean --dry-run

## Expected behavior

The command finishing and the system not claiming malware. 

## Debug logs

I hope you understand I'm a bit reluctant to run any more commands. 

## Environment
Mole version 1.39.1
macOS: 26.5
Architecture: arm64
Kernel: 25.5.0
SIP: Enabled
Disk Free: 36Gi
Install: Homebrew
Shell: /bin/zsh

## Additional context

<img width="328" height="302" alt="Image" src="https://github.com/user-attachments/assets/52e5966d-228c-4108-8f65-1db1f1b2873b" />

tw93:
@TsubasaNoNai thank you for the report. This is fixed on `main` in d4623fbe by stopping `mo clean` and `mo clean --dry-run` from touching Chromium Service Worker ScriptCache paths, which were the risky paths behind the system warning.

You can get the fix now with `mo update --nightly`. The next stable release will include it too. I also verified locally that `mo clean --dry-run` no longer reports `Service Worker` / `ScriptCache` cleanup.

arjenfvellinga:
I still get this error running `mo clean --dry-run`

I assume the fix from may is in the latest released version.

Mole version 1.49.2
macOS: 26.6
Architecture: arm64
Kernel: 25.6.0
SIP: Enabled
Disk Free: 20.34GB
Install: Homebrew
Shell: /bin/zsh

PR #1588, opened 2026-09-17
feat: clean superseded Chrome extension versions

jiji262:
## Summary

Chrome installs every extension update as a new `Extensions/<id>/<version>_<n>` directory. The superseded copy is removed only by Chrome's own `ExtensionGarbageCollector`, which runs once per profile, 30 s after that profile loads. A long Chrome session therefore keeps a full copy of every extension that updated. Measured on a daily-driver Mac (macOS 15.6.1, Chrome 152, 24 profiles, Chrome running since Sep 10): **25 stale version dirs, 747,028 KB (~730 MB)** in the two profiles that were in use; the 22 profiles that were not loaded had none.

This adds `clean_chrome_extension_old_versions` in `lib/clean/user.sh`, called from `clean_browsers` next to `clean_chrome_old_versions`.

### Keep-set

Per profile and per extension id, keep:

1. `extensions.settings.<id>.path`: the version Chrome loads.
2. `extensions.settings.<id>.idle_install_info.path`: an update waiting for the extension to go idle.
3. Every directory of the highest version (numeric per component, `_<n>` ignored).

1 and 2 are read from both `Preferences` and `Secure Preferences` (branded Chrome uses the latter). That is exactly the keep-set of Chromium's collector (`GetInstalledExtensionsInfo()` + `GetAllDelayedInstallInfo()` in [`extension_garbage_collector.cc`](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/extensions/extension_garbage_collector.cc)) plus the highest version, so Mole only removes a subset of what Chrome removes itself on the profile's next load.

"Keep the highest version" alone is not safe. On the Mac above, an extension had prefs `path = …/1.0.91_0` and `idle_install_info.path = …/1.0.93_0`: the running version was the lower one. If Chrome finds the referenced directory gone, it registers the extension from its cached manifest anyway. Web Store extensions are then disabled as corrupted and re-downloaded; other extensions stay broken (`installed_loader.cc`, `chrome_content_verifier_delegate.cc`).

### Fail-closed rules

- **Chrome must be closed.** `is_google_chrome_running` must report not running and `SingletonLock` must be absent from the user-data dir. The lock is held by any Chromium build using that directory, including ones pgrep would not match. Running → deferred like the other Chrome caches. Unknown → `skipped (process state unknown)`. A local lock whose pid is gone (left by a crash) is also reported as unknown rather than deferred silently. Both are re-checked for every directory at the sink.
- **Eligibility first.** Before probing Chrome, a filesystem-only pass lists dirs below each extension's highest version and runs `mole_cleanup_targets_exist`. A whitelisted or protected-only profile never records a deferral.
- **Profile-level refusals.** Each prefs file's `extensions.settings` is read once into memory, so a read failure later in planning cannot pass for a missing reference. If a file exists but does not read and parse, the whole profile is left alone. The profile's targets are dropped if either prefs file's `dev:inode:size:mtime:ctime` changes during planning. If it changes between planning and removal, the run stops with `stopped (preferences changed)`. A version dir replaced by a symlink before the final check stops with `stopped (target changed)`, and the removal is bound to the inode checked there.
- **Extension-level refusals.** An extension is left alone when: it has no reference; a reference is not exactly `<id>/<version dir>` (absolute unpacked paths included); a referenced directory is missing; or any child is a symlink or a directory name Chrome does not create (`^[0-9]{1,10}(\.[0-9]{1,10}){0,3}_[0-9]{1,2}$`).
- **Bounded reads.** Every `plutil` read runs under `run_with_timeout`, inside one planning budget. A timeout discards the whole plan (`skipped (preferences read timed out)`), and a signal cancels the run. `plutil` prints its "no value" error on stdout, so presence is decided by exit status and value shape, never by output.
- **Standard sink.** Removal goes through `safe_clean_guarded`, so whitelist, protected-path, dry-run preview, and parent/target identity rebinding apply as usual. No new sudo, no new deletion primitive, no `declare -f` fallback.

### Explicit non-targets

`Extensions/Temp`, whole `<id>` directories (including ids missing from prefs), `UnpackedExtensions/`, `Local Extension Settings`, `Sync Extension Settings`, `Extension State`/`Rules`/`Scripts`, `IndexedDB`, `Service Worker/ScriptCache` (#785 #964 #968), Chrome Beta/Canary/Dev user-data dirs, and other Chromium browsers.

### Owner-metadata questions (bugs §14)

- **Inventory or journal?** `extensions.settings` is Chrome's installed-extension registry, and Chrome's own collector deletes every version directory it does not reference. The data is used as deletion authority by its owner, not reconstructed by Mole.
- **Every profile and writer enumerable?** The keep-set is computed per profile from that profile's own prefs, so there is no cross-profile union like the reverted VS Code reconcile (55ea2644 → 08407721). Writers are Chromium processes on that user-data dir, which `SingletonLock` covers.
- **Shared lock or GC command?** There is no GC command. Mole cannot hold `SingletonLock`, so it requires the lock to be absent and re-checks the lock, the process, and the prefs stamp immediately before each removal.
- **New reference after planning?** Not while Chrome is closed. On launch, Chrome promotes `idle_install_info` (kept) and installs new versions into a fresh `<version>_<n>` name, never reusing an existing directory (`file_util.cc`, first free suffix).
- **Interruption equivalent to a cache miss?** Yes. A partially removed directory is unreferenced, so Chrome never loads it and its collector removes the remainder.

### Value caveat

Chrome reclaims these directories itself about 30 s after the next launch of each profile. Mole's gain is reclaiming them while Chrome is closed, plus profiles that were used and then not reopened. If that is too transient for default `clean`, I'm happy to narrow or drop it.

## Also fixed: `clean_user_core.bats` deleted a real Chrome version

While verifying this change, `MOLE_TEST_NO_AUTH=1 bats tests/clean_user_core.bats` (the AGENTS.md hotspot command) removed `/Applications/Google Chrome.app/Contents/Frameworks/Google Chrome Framework.framework/Versions/152.0.7977.76`. That was the version the running Chrome was still using, next to a staged `152.0.7977.83`. Several `clean_browsers` cases stub `pgrep` as "not running" without setting `MOLE_CHROME_APP_PATHS`, so `clean_chrome_old_versions` fell back to `/Applications`.

The first commit pins `MOLE_CHROME_APP_PATHS` / `MOLE_EDGE_APP_PATHS` / `MOLE_BRAVE_APP_PATHS` to the fixture `HOME` in `setup_file`. It adds `browser old-version cleaners stay inside the fixture HOME`, which fails when the exports are removed.

## Tests

- `./scripts/check.sh` passes (ShellCheck, function duplication audit, destructive sink audit).
- `MOLE_TEST_NO_AUTH=1 bats tests/clean_browser_versions.bats tests/clean_user_core.bats`: 115/115.
- `MOLE_TEST_NO_AUTH=1 bats tests/clean_core.bats tests/clean_app_caches.bats tests/clean_cached_device_firmware.bats`: all pass except 6 `clean_editor_obsolete_extensions` / #1461 / #1512 cases in `clean_app_caches.bats`, which fail the same way on a clean `main` checkout on this machine. Two TTY section-header tests in `clean_core.bats` failed once in a background run, then passed on rerun and in an earlier full run.
- Mutation check: 20 single-line mutations of the keep-set, prefs merge, parse check, owner re-check, stamp, identity binding, and 124/≥128 classification each make at least one new test fail.
- Real Mac, read-only: the planner on real prefs produced exactly the same 25 directories as an independent Python implementation of the rule, in about 1 s. `MOLE_TEST_NO_AUTH=1 MOLE_DRY_RUN=1 ./mole clean --dry-run`: exit 0. The extension row was deferred because Chrome was running.
- Not run locally: the full `./scripts/test.sh`. After `clean_user_core.bats` deleted a real app version on this machine, I only ran files that a static audit confirmed stay inside the fixture `HOME`, and I'm relying on CI for the rest.

## Safety-related changes

- New deletion target: stale Chrome extension version directories, through `safe_clean_guarded` with a per-path guard (Chrome closed, `SingletonLock` absent, prefs unchanged, identity bound).
- Test isolation fix so `clean_user_core.bats` can no longer reach `/Applications` browser bundles.


tw93:
@jiji262 The test isolation fix is on `main` with your authorship. Without those three exports the guard test really does fail on `[[ "$output" != *"APP:/Applications/"* ]]`, so the suite could reach a live browser bundle on any machine that has Chrome installed. That was worth catching on its own. I moved the new `@test` below `teardown_file`, it was sitting between the two file-level hooks.

I am not taking the extension cleaner, and your own measurement is the reason. All 25 stale directories were in the two profiles that were in use, and those are exactly the profiles Chrome's collector clears about 30 s into their next load, while the 22 profiles that had not been loaded held none to begin with. So the window Mole would cover is "Chrome is closed and I want the space back before I open it again". That is real, but it costs 338 lines of new deletion logic aimed at a target that reclaims itself, and the deletion surface in `clean` is the one thing I keep small on purpose.

That is not a comment on the implementation. The keep-set derivation, the `SingletonLock` handling and the prefs-stamp rebinding are more careful than plenty of what is already in there, and the `idle_install_info` case where the loaded version is the lower one is a trap I would not have predicted. Thank you for the work, and for stating the caveat yourself instead of leaving me to find it.

